Issue #1, opened 2016-01-09
Parser for signatures


jonsterling:
We need to implement a parser for developments. The parser will still be a little tricky, because as it goes, it needs to accumulate parsing rules for declared operators (and eventually, custom notations—but we'll leave that for after the basics are done). The idea is that when a declaration like the following is parsed,

````
Def something[u:σ, v:τ](M : ({σ}[τ].τ)τ, ...) : τ = [
  N
].
````

... we need to remember how to parse applications of the `something` operator. So, in the future we must parse `something[u,v]` as `CUST ("something", [(u, σ), (v, τ)], arity)`, where `arity` is the representation of `({σ}[τ].τ, ...)τ`.

The parser will be slightly less tricky than the old one, since we're abandoning the old "open-ended datatype" approach to custom operators, so there's no need for side effects in the parser. Additionally, we always parse first to ASTs, and only then do we convert them to ABTs, which means that all our parsers will be significantly easier (since we do not need to assign identity to variables and symbols immediately).

Other forms of declaration (like `Thm` and `Tac`) will do exactly the same thing as `Def`; in an evaluator for developments, we will separately execute `Thm` declarations.

jonsterling:
(By the way, the design of developments is now such that there is no separate `Operator` declaration. In practice, it was possible to use this to do something suspicious things like mutually recursive definitions, etc. And it was kind of noisy and surprising to many users!)

jonsterling:
And previously, (though I'm not sure this is the right ticket in which to make this remark), we had the following hierarchy of forms of equivalence:

- `α`-equivalence: in the logical framework
- `δ`-equivalence: based on unfolding of definitions
- `~`-equivalence (computational): based on the operational semantics
- `=`-equivalence (behavioral): based on the types

In the new order, we remove the notion of `δ`-equivalence and subsume it under `~`-equivalence. It might be interesting to consider extending the ABT framework such that you can define real algebraic theories (as opposed to the "free theory" for a signature, with no equations). Then, we would have another kind of equivalence that everything must respect; however, I want to put off something like that for a while, since I don't have any use-case for it at the moment.

jonsterling:
Thanks to @ak3n for getting the ball rolling on this! We can close this ticket now.

PR #3, opened 2016-01-09
Set up machinery for signature phases, morphisms

jonsterling:
The idea is that we can parse and interpret JonPRL signatures in different phases; then, we can transport signatures along partial "signature-morphisms". So, we'd start with parsing a `SIGNATURE[string]`, which has the structure of a signature, but all the relevant datatypes are just strings.

Then we create a signature morphism `SIGNATURE[string] >-> SIGNATURE[abt]`, which actually parses all the strings into abts, sorts, etc. We could also have another phase `SIGNATURE[valid]` w/ morphism `SIGNATURE[abt] >-> SIGNATURE[valid]`, which actually executes the included tactic scripts, etc.

The benefit of doing this incrementally rather than all at once is that we can take advantage of the existing structure whilst doing things like parsing or executing; this is important, because the nature of our parser changes as we work our way down a signature (because of notation declarations, etc.).


jonsterling:
This PR introduces general-purpose machinery that will make it a lot easier to define the parser for developments/signatures (ticket #1)! Here's all that's needed:

1. Define a parser for `StringSignature` that just captures strings in place of all the interesting stuff; so, it basically just parses the *structure* of a signature, without any of its content.

2. Define a signature morphism `ParseSignatureMorphism : SIGNATURE_MORPHISM where S1 = StringSignature and S2 = AstSignature`. In here, you will be obligated to provide a parser for sorts, valences, and expressions; you may even provide a different parser depending on the sort that is being parsed into, which will be useful to provide nice notation for tactic scripts, as opposed to terms, etc. All this falls out of the basic structure of `SIGNATURE_MORPHISM`.

Then, the `TransportSignature` functor can be used to take a stringy signature into a fully parsed one.

PR #5, opened 2016-01-10
Use symbols rather than strings for opids

jonsterling:


jozefg:
Nice! Looks pretty slick to me

jonsterling:
@jozefg Cool, thanks!

I recall that originally I thought we might do this in the future, but as soon as I started trying to decide what the static semantics for signatures should look like, it became clear that it would be better to do this sooner rather than later.

PR #7, opened 2016-01-11
Validate custom operators - WIP

### LINKED ISSUE #6: enforce arities of CUST operators
As I was working on the elaboration semantics for signatures, I realized that we aren't actually enforcing that `CUST` operators carry the correct arities. We'll need to introduce a well-formedness check for all terms that are introduced into a signature, that walks the abt to find instances of `CUST[ϑ](a)`, looks up the arity `a'` assigned to `ϑ` in the signature, and checks that `a == a'`.

- [ ] update elaboration spec
- [x] update code

jozefg:
This fixes #6. It currently includes all the code but requires a small amount of changes in the spec before it's ready to be merged in.

### DIFF
--- a/src/signature.mlb
+++ b/src/signature.mlb
@@ -27,6 +27,13 @@ local
     signature/abt_signature.sml
     signature/bind_signature_morphism.sml
   end
+
+  local
+    syntax.mlb
+  in
+    signature/validation_morphism.sml
+  end
+
 in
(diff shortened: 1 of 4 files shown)


jonsterling:
I'm going to merge this in now, since we can update the spec at whichever time we actually formally define the elaboration semantics for abt syntax.

Issue #8, opened 2016-01-11
smash SIGNATURE_MORPHISM bureaucracy


jonsterling:
[Twenty Manifestations of Bureaucracy](https://www.marxists.org/reference/archive/mao/selected-works/volume-9/mswv9_85.htm)

jozefg:
Haha :+1: I'll tackle this tomorrow

PR #9, opened 2016-01-12
Removed the pattern imposed by SIGNATURE_MORPHISM (#8)

jozefg:
Most of the transformation was quite mechanical; all of the interesting code remains but the glue around it is different. This version is much more efficient though because we don't rebuild the intermediate information we need for every declaration.


jonsterling:
Excellent, this looks great! Nicely done, thanks.

jozefg:
Thanks!

PR #12, opened 2016-01-18
Add signature parser.

ak3n:


jozefg:
Nice :D

jonsterling:
@ak3n Thanks very much, nicely done! I'll review this in detail later on today. Welcome to the Red JonPRL Group!

ak3n:
Thanks!

I hope that it's not so bad and ready to fix and write tests.

src/signature/parser.sml
```diff
@@ -0,0 +1,99 @@
...
+
+  fun inBrackets p = char #"[" >> p << char #"]"
+  fun inParentheses p = char #"(" >> p << char #")"
+  fun inBraces p = char #"{" >> p << char #"}"
+
+  val skipChars = repeatSkip (space <|> char #"\t" <|> char #"\n")
+
+  val opid = repeat1 letter wth String.implode
+  val sortid = repeat1 letter wth String.implode
+  val term = repeat1 letter wth String.implode
```
jonsterling: Probably we can't use `letter` here, since we want to parse arbitrary strings of any characters in (later on, to be parsed into actual terms in a separate phase of parsing/elaboration).

Can we use a parser that always succeeds here?

src/signature/parser.sml
```diff
@@ -0,0 +1,99 @@
...
+  val metaid = repeat1 letter wth String.implode
+  val symid = repeat1 letter wth String.implode
+
+  val sortlist = separate sortid (char #"," >> space)
+
+  val valence = opt (opt (inBraces sortlist) &&
+    opt (inBrackets sortlist) << char #".") && sortid when
+      (fn (SOME (SOME s1, SOME s2), s3) => SOME (s1, s2, s3)
```
jonsterling: Out of curiosity, if every case here returns `SOME`, can we just change this to return the tuple directly instead?

jonsterling:
For tests, we can follow the very simple approach for testing that Danny proposed in #13; the basic idea is:

1. Add a file like `signature_parser_test.jonprl` to the repository
2. Create a structure like `TestSignatureParser`, along the lines of the following pseudo-code:

~~~~sml
structure TestSignatureParser = 
struct
  val input = readFile "signature_parser_test.jonprl"
  val parsed = parseSignature input
  val () =  
    if someUsefulCheck parsed then
      ()
    else
      raise Fail "Oh no!"
end
~~~~

Then, this test structure would be included in some `test.mlb` file or something which imports whatever other `mlb` files it needs, and then to run the tests, we just run `mlton test.mlb`. Sound OK?

ak3n:
@jonsterling yes, looks fine. I can try to add tests tomorrow.

If there are no any problems with it and it can be merged, I can create a new request.

jonsterling:
OK, great! Let's merge this in now, and the tests can be done separately (along with any possible fixes that the tests suggest).

Thanks for your hard work & contribution! :smile:

ak3n:
I'm happy to help!

PR #20, opened 2016-01-24
[Definition] Nominal LCF

jonsterling:


jonsterling:
I've made some modifications here.

1. Uniform continuity of tactics is not needed to implement the dynamics. Only the usual continuity is needed.

2. Atomic tactics are not necessarily uniformly continuous; in fact, uniform continuity of "atomic tactics" is *not* a consequence of the Fan Theorem, since we have not justified the notion that the sequences of atoms form a fan, only that they form a spread.

Atoms are weird—there is some sense in which there are not an infinite number of atoms in existence at any stage, so it seems plausible that we could use the Fan Theorem—but neither can we observe that there are only a finite number of atoms in existence, so I do not think that the Fan Theorem is quite applicable here.

In any case, it is of no consequence because we did not need a uniform modulus of continuity.

/cc @jozefg @freebroccolo

jonsterling:
Another change I'd like to try making is to have "concrete" tactics elaborate to actual ML tactics—currently, we just give an operational semantics to the concrete tactics applied to a judgment, but in order to interpret things like recursion / a `fix` tactical, this won't suffice. Initially, I didn't think it was possible to carry this through, but now I am more hopeful.

jonsterling:
(One benefit of doing it the that way would be that I can literally elaborate these to plain LCF scripts, using the pre-defined `THEN`, `THENL`, etc. tacticals—rather than having to do all this nasty list-munging here.)

jonsterling:
At first, I had been hoping that `xs <- ([] <- t1; t2); t3` and `(xs <- t1; t2); t3` would be equivalent, but I have shown that they are not; however, I think this is a feature rather than a bug. The only difference is that in the first, `t3` knows about the names in `xs` and can consume any of them that are left over from `t1` and `t2`, whereas in the second script, `t3` is totally ignorant of the names in `xs`.

In fact, had I thought about the syntax a bit, this should have been obvious—the fact that these have a different meaning is just a manifestation of the fact that we have imposed a *scoped* dynamics rather than a *free* dynamics for tactic scripts. So, the definitions that I've put in seem to be pretty well-behaved in this sense!

jonsterling:
OK, this appears to be good to go! One thing I'd like to explore later on is an efficient compilation scheme for tactic scripts. Seems like this would be amenable to some sort of abstract machine treatment.

PR #21, opened 2016-01-31
Dependent LCF

jonsterling:
This switches our tactic elaborator to target Dependent LCF; the change was nearly trivial, because Dependent LCF (mostly) shares a signature with Classic LCF. I also removed the need for these "lazy" versions of the `THEN-` tacticals (I don't know why I didn't do what I have now done in the first place.)


jonsterling:
:bomb:s away!

Issue #11, opened 2016-01-16
operational semantics


jonsterling:
We can implement the operational semantics in a way that resembles the following:

```sml
datatype 'a step = 
    STEP of 'a
  | FINAL

(* we can clean up a lot of the gnarly aspects of our old code
   by providing the following: *)
structure StepMonad : MONAD where type 'a t = 'a step
structure StepFunctor : FUNCTOR where type 'a t = 'a step

exception Stuck

(* Represents the judgment [Sig !- M :-> N], subject to the following
   postsupposition: [freeSymbols N] shall be a subset of [freeSymbols M].
   i.e. we are propounding a *scoped dynamics* in Bob's terminology. *)
val step  : sign -> abt -> abt step
```

The `step` operation takes a signature, so that it can look up definitional extensions that are declared in the signature.

@jozefg sound all right? Under the new order, there's no need to worry about universe levels when unfolding definitions, so this should be pretty straightforward.

jozefg:
Yeah this all looks pretty good to me! I'm happy to try to implement this once we're at that point :) Always fun to write very slow interpreters

PR #22, opened 2016-01-31
Start stubbing out the structure of the dynamics

jonsterling:
This is to begin addressing #11.

@jozefg Does the `DYNAMICS` signature look reasonable to you?


jonsterling:
I'll merge this now, but I expect we'll have bugfixes to do as soon as we have any way to try this code out.

PR #32, opened 2016-02-06
Add Universe Hierarchy

### LINKED ISSUE #31: Add universe hierarchy

jonsterling:
To resolve #31


jonsterling:
With Dependent LCF, we'll be able to write a cumulativity rule that allows you to use the refiner to construct the level parameter; however, in order to do this, we will need to add the `Base{τ}` type, and we will also need to add some new rules for reasoning about level ordering internally.

I'll leave out the `cum` rule until I can get around to this.

PR #33, opened 2016-02-06
Add frontend based on old version.

ak3n:
I am not sure about frontend.cm. I tried to test it via `CM.make "jonprl.cm"` in sml and looks like there are no problems.


jonsterling:
@ak3n Wow, thanks so much! I'll try this out and get it merged in. I really appreciate it! :smile:

PR #41, opened 2016-02-08
Fix issue 40 by rearranging the syntax of Nominal LCF

### LINKED ISSUE #40: [Nominal LCF] associativity is wrong
Here's a dilemma:

1. We want tactic sequencing to associate to the left: so, `t1; t2; t3` should be `THEN(THEN(t1;t2); t3)`.
2. We want name scoping to associate to the right; so, in `a <- t1; b <- t2; c <- t3`, the name `a` should be in scope for the rest of the script.

jonsterling:
To resolve #40 

Still need to update the Definition.


jonsterling:
Was a lot less painful than I expected... The basic idea of the change is that where previously we had `u* <- tac; multitac`, it is not `u* <- multitac; tac`. The former "seems" more intuitive (when you think about what's actually going on when you sequence tactics), but the latter is actually what we really want.

Whereas under plain LCF, it is possible to view tactic scripts in a sort of "big-step/natural" style, the Nominal LCF perspective forces a procedural point of view.

PR #45, opened 2016-02-09
Richer judgmental apparatus

### LINKED ISSUE #43: Add goals other than truth
If we added a goal like `H >> IsType{τ}(T)` whose evidence is a universe level, it would be possible to write some cool level inference heuristics as a rule or tactic.

This idea also lets us relax the old "PRL" semantics such that the main postsuppositions are only required for the truth sequents.

jonsterling:
To resolve #43, I've introduced a new form of sequent that is meant to prove that something is a type, at *some* (unspecified) level. The synthesis/evidence of the judgment is a closed level expression such that the type is a member of that universe.

This allows us to do level inference as a tactic outside the refiner, which is awesome. It is possible to imagine a similar treatment being used to enable other forms of implicit parameter (/cc @freebroccolo).

### DIFF
--- a/src/dynamics/dynamics.sml
+++ b/src/dynamics/dynamics.sml
@@ -105,7 +105,7 @@ struct
       case theta $ args of
            CUST (opid, params, arity) $ args =>
              stepCust sign (opid, arity) cl
-         | LVL_OP (LBASE _) $ _ => FINAL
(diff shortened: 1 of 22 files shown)


freebroccolo:
Very nice :)

PR #60, opened 2016-02-14
Pi types; big refiner chagnes

jonsterling:
- [x] type eq
- [x] member eq
- [x] ap eq
- [x] intro
- [x] elim
- [x] ext

### DIFF
--- a/emacs/red-jonprl-mode.el
+++ b/emacs/red-jonprl-mode.el
@@ -31,7 +31,7 @@
 
 (defvar red-jonprl-tactic-name-regexp
   (regexp-opt
-   '("auto" "eval-goal" "ceval" "csym" "cstep" "witness" "trace" "intro" "elim" "eq")
+   '("auto" "eval-goal" "ceval" "csym" "cstep" "witness" "trace" "intro" "elim" "eq" "id" "normalize" "unfold")
    'words))
 
 (defun red-jonprl-font-lock-defaults ()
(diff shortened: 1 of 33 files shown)


jonsterling:
So, there was a problem in the way I'd designed the dependent refiner before—I had rather lazily just made the valence of the synthesis of a sequent judgment bind no variables, and in most cases this works out fine, but you can end up in a very bad situation when you begin to extend the context, which could lead to possible unsoundness.

For instance, let `#m` be a metavariable of valence `.exp`; let's say that `#m` is represents the synthesis of a subgoal which added a new variable `z` to the context. Then, in the partial evidence for the current goal, I mention both `[a/z]#m[]`  and `[b/z]#m[]`; however, since `#m[]` doesn't have `a,b` free, these are the same. So, in intermediate proof states, I can have an inaccurate list of goals.

The correct solution is to close over the variables that we are extending the context with (producing evidence of non-trivial valence), so that we can then instantiate the evidence of the subgoals in the correct way. At first, I thought it would be necessary to have a very ugly (and badly performing) situation where the valence of a sequent's evidence is based on the number of hypotheses in the sequent, but I realized that it is possible to introduce generality in a compositional way. The sequent expresses which variables are meant to be bound—and this only matters in the synthesis of the main goal, whereas as far as the subgoals are concerned, these variables are just free.

For example, the sequent `G |> H >> P` produces as evidence an abstraction that closes over the variables in `G`; `dom(G)` of course should be a subset of `dom(H)`.

I'm sorry if this is difficult to understand—it may be best to simply look at the code, at least until I formalize the PRL.

Issue #67, opened 2016-04-11
Refactor Nominal LCF elaborator


jonsterling:
I've got a cleaner method of elaboration in the paper draft, I think, which is a bit more compositional. We should try it out...

PR #68, opened 2016-04-12
Nominal lcf elaborator rewrite

jonsterling:
This is a rewrite of the Nominal LCF elaborator, along the lines I proposed in the draft paper. However, I've made some improvements here that make the semantics a bit less weird.

It may be worthwhile to refactor the grammar of Nominal LCF a bit (in particular, the sorts and sort assignments).

I'm thinking for the paper, we should simply extract a "publishable" core functor from this code and do a Literate ML thing rather than a mathematical semantics. Might be a good fit for TFP or something along those lines, not sure.

#67


src/nominal-lcf/elaborator.fun
```diff
@@ -151,50 +164,60 @@ struct
...
+               LCF (SEQ taus) $ [([],[]) \ m, (us, _) \ t] => (us, elaborateMultitactic sign rho m) :: go t
+             | _ => [([], extendTactic (elaborateTactic sign rho t))]
+        end
+    in
+      contractTactic
+        o List.foldr
+            (fn ((us, front : Refiner.nmultitactic), rest : Refiner.nmultitactic) => fn alpha => fn st =>
+              let
+                val beta = prepend us alpha
+                val (beta', modulus) = probe (prepend us beta)
```
jonsterling: nice thing about this change is that the core conceit (the probing for the modulus of continuity) is now consigned to a *single point* in the code, right here.

PR #80, opened 2016-04-30
Experimental bidirectional rules

jonsterling:


src/syntax/ctt/operator.sml
```diff
@@ -12,6 +12,7 @@ struct
     | SQUASH of Sort.t
     | ENSEMBLE of Sort.t * Sort.t
     | DFUN | FUN | LAM | AP
+    | DFUN_DOM | DFUN_COD
```
jonsterling: These operators take a term like `dfun(A; [x].B[x])` and extract the (co)domain! It's extensional because `dfun` is canonical.

jonsterling:
I think based on this experiment, it may be more desirable to have a judgment like `M synth` rather than `M = N synth` or whatever; then, the chk-inf rule would synthesize both sides and check that the results are equal.

jonsterling:
Furthermore, this machinery could subsume the `P TYPE` judgment machinery, which we are only using for level synthesis. We would just need to add an noncanonical operator to project the level from a universe (which we can absolutely do in our computation system!)

A nice observation here is that when you inspect the result of a type synthesis judgment, it is always on the basis of canonical forms. As a result, it is possible for us to do that inspection directly in the computation system.

jonsterling:
What is the meaning explanation for type synthesis?

To begin with, I am thinking of the bidirectional rules as the identification of the essentially dialectical character of the *proof theory* of CTT, which was obscured in RedPRL's reactionary predecessors.

I believe that in fact this will work much the same as in intensional bidirectional type theory, except that in the checker will always be a bit stronger than the synthesizer—in such cases, the synthesizer delegates to the user to supply a type, which is then passed to the checker.

I think we may say that the checker must take the answers given by the synthesizer at face-value, because type synthesis comes from either the way that a neutral term is formed (self-evident) or from the checker itself (in the case that the synthesizer delegates to the checker in a change-of-mode).

PR #84, opened 2016-05-09
Upgrade abt lib

jonsterling:


jonsterling:
Lots of improvements here:

1. You no longer need to pass around a metavariable context when constructing terms; this was achieved by adding a few little annotations to the view/patterns for metavariable applications.

2. Free variables/metavariables/symbols are computed incrementally & lazily. I expect this will be a big performance improvement.

3. The ABT framework is now compatible with purely functional notion of variable/symbol; however, the Dependent LCF library needs to be updated to support this still.

PR #92, opened 2016-07-03
start adding json stuff

jonsterling:
- [x] atomic sorts
- [x] atom operators
- [x] level operators
- [x] ctt operators
- [x] tactic operators
- [x] record operators
- [x] top-level operators
- [x] LCS sorts
- [x] LCS operators
- [x] rig up JSON encoder for redprl abts
- [x] encode signatures
- [x] decode signatures

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -16,3 +16,6 @@
 [submodule "lib/sml-unparse"]
 	path = lib/sml-unparse
 	url = https://github.com/RedPRL/sml-unparse.git
+[submodule "lib/sml-json"]
+	path = lib/sml-json
+	url = https://github.com/RedPRL/sml-json.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -19,3 +19,6 @@
 [submodule "lib/sml-json"]
 	path = lib/sml-json
 	url = https://github.com/RedPRL/sml-json.git
+[submodule "lib/sml-parcom"]
+	path = lib/sml-parcom
+	url = https://github.com/RedPRL/sml-parcom.git
--- a/mlb-path-map
+++ b/mlb-path-map
@@ -1,1 +1,0 @@
-LIBS lib
--- a/script/mlton.sh
+++ b/script/mlton.sh
@@ -1,4 +1,6 @@
-#!/bin/sh
+#!/bin/bash
+
+LIBS=$(pwd)/lib
 
 mkdir -p bin
-mlton -mlb-path-map mlb-path-map -output bin/redprl src/frontend.mlb
+mlton -mlb-path-var "LIBS $LIBS" -output bin/redprl src/frontend.mlb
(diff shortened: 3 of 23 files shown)


jonsterling:
Hmm, I can't figure out what's going wrong with the mlb file...

jonsterling:
Ah, I think I see what is happening. I seem to have misunderstood how MLton resolves paths. Here's what I expected to happen:

> With path map `LIBS lib`, any MLB file resolves `$(LIB)` to the `lib` folder at the root directory of the project, or, the directory from which `mlton` was run.

The above is consistent with how SML/NJ's CM appears to work. Instead, what appears to happen is the following:

> With path map `LIBS lib`, any MLB file resolves `$(LIB)` to the `lib` folder at the level of the current `mlb` file.

I can see how this behavior makes sense, but it seems to make much more difficult the case where an application can specify the location of its libraries dependencies. Under this behavior, an absolute path would seem to be required, but this absolute path must be determined once for every working copy of the repository. Do I understand this right, @MatthewFluet?

We can solve this locally by using `pwd` in the build script, though.

MatthewFluet:
@jonsterling Yes, MLton's MLB path variables are somewhat different from SML/NJ's CM path anchors.  Within an MLB file, when processing a file reference, any MLB path variables are (recursively) expanded and the resulting path is interpreted as relative to the immediate containing MLB file (or interpreted as absolute, if the path has a leading `/`).  In fact, MLB path variables need not correspond to full paths at all and, therefore, aren't changed when processing nested MLB files.

From what I see with RedPRL, with all(?) of the dependencies pulled in as submodules under the (root) `lib` directory, then using an absolute path would work, but so would simply using `../lib` or `../../lib` as necessary.  I acknowledge that the later does inhibit some kinds of moving sources without also updating various MLB files.

jonsterling:
@MatthewFluet thanks for clarifying! I've ended up using the absolute path thing, and determining the proper one in the build script.

The `../lib` thing works well enough internally to RedPRL, but the problem is that many of our dependencies themselves have dependencies which are shared, so in that case, the absolute path thing is needed. One thing that might be ideal would be to sufficiently functorize our libraries such that there is only one level of dependencies; this might make things a bit frustrating to use, though, so we haven't done it.

jonsterling:
I think I'll merge this now, and then I can begin to use this machinery in another pull request.

The point of this work, btw, was to provide a machine-readable serialization format for RedPRL signatures which is not sensitive to changes in concrete notation; I want to use this to cache proofs, kind of like Agda does with the `.agdai` files.

MatthewFluet:
@jonsterling Ah, yes, now I see that you want `./lib/json-lib` to use the shared `./lib/cmlib` rather than its "local" sub-module `./lib/json-lib/lib/cmlib`.  The absolute path thing is probably the best approach.

PR #98, opened 2016-08-06
Revamp records; don't use dep-isect as primitive representation

### LINKED ISSUE #91: consider primitive dependent records
I'm finding that the subgoals being generated for dependent intersection are still fairly difficult to deal with. It may be better to have primitive dependent record types (which you can still compose by dependent intersection), rather than only having singleton records primitively.

jonsterling:
I know I've gone back and forth on this a lot, but I've found that
dep-isect is really detrimental to automation. The new record type is
*extensionally* equivalent to the dep-isect representation, and we could
add a rule to that effect if we wanted.

This commit moves us back to rules like those in
http://homepages.inf.ed.ac.uk/rpollack/export/TLCA03extended.pdf

(resolves #91)

### DIFF
--- a/src/syntax/syntax.fun
+++ b/src/syntax/syntax.fun
@@ -21,6 +21,8 @@ sig
   val $$ : symbol operator * term bview spine -> term
   val out : term -> term view
 
+  val variable : string -> variable
+
   val debugToString : term -> string
 end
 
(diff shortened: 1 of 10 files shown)


src/syntax/syntax.fun
```diff
@@ -21,6 +21,8 @@ sig
   val $$ : symbol operator * term bview spine -> term
   val out : term -> term view
 
+  val variable : string -> variable
```
jonsterling: oops, don't need this.

PR #100, opened 2016-08-07
Cubical syntax & dynamics, take 1

jonsterling:
This is very much WIP, but I'm starting to add the syntax (and dynamics) of cubical type theory. Because RedPRL uses a machine dynamics, stuff will be a bit more intricate than in the paper, but lets us implement things more concisely & makes things less brittle in the long run.

- [x] implement dimension substitution


jonsterling:
note: when we start adding canonical hcoms, we will need to ensure that they are replaced with non-canonical ones when they are affected by dimension substitutions. hcom is one example of where the standard LCS approach doesn't work out so well, since the canonicity of an hcom is determined by its subterms. One option might be to let the canonical hcoms take their cap-and-composition dimensions as static parameters, to avoid confusion.

jonsterling:
I'm going to experiment with a different way of representing the dimension syntax which may make things a bit more clear and less ad hoc.

jonsterling:
So (as discussed), the way I've rigged it up, internally the canonical hcoms are different operators from the non-canonical ones; this is a consequence of the Howe/LCS apparatus. However, outside the core syntax module, they are not distinguishable and hcoms which are canonical will be represented as such. Because of the abstraction provided here, the tricky case that we discussed with dimension substitutions is handled automatically.

jonsterling:
OK! Got the dynamics implemented for the weak booleans. No doubt there are bugs in the code, since it's never been run, but I will merge after Travis validates the build and then proceed to begin adding some judgment forms & rules for cubical type theory.

PR #101, opened 2016-08-10
[WIP] Keep executed signatures.

ak3n:
Right now it fails:
screen shot 2016-08-10 at 02 53 37

and creates duplicates:
screen shot 2016-08-10 at 02 55 22

So, there is need to replace Telescope.append and find a correct way to use the result of old elaborations.

### DIFF
--- a/src/interactive/commands.sml
+++ b/src/interactive/commands.sml
@@ -56,7 +56,7 @@ struct
       printPart num list
     end
 
-  fun handleCommand command =
+  fun handleCommand oldSign command =
     case command of
       Execute code =>
         let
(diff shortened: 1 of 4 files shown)


jonsterling:
Neat!

Is the problem that it is somehow repeatedly appending the same objects to the telescope or something? Or could you go into a bit more detail about what's happening?

src/interactive/commands.sml
```diff
@@ -66,14 +66,15 @@ struct
             INL s => raise Fail s
           | INR sign =>
             let
-              val elab = RefineElab.transport o ValidationElab.transport o BindSignatureElab.transport
-              val sign' = elab sign
+              val elab = ValidationElab.transport o BindSignatureElab.transport
+              val sign' = RefineElab.transport (AbtSignature.Telescope.append (oldSign, elab sign))
```
ak3n: There is need to replace `append` with some kind of merge which should replace entities with the same name from sign in oldSign.

ak3n: Also, we pass `oldSign` to `RefineElab`, but not to `BingSignatureElab`. `BindSignatureElab` takes StringTelescope and returns SymbolTelescope. How to transform SymbolTelescope to StringTelescope to pass it to `BindSignatureElab`?

jonsterling: @ak3n If it seems like we need to convert symbol telescopes to string telescopes, I think we've taken a wrong turn... Can you think of any other way to do what you need?

jonsterling: Also, keep in mind that the signature elaboration architecture is not set in stone, and we can change it in whatever way would best enable what you're trying to do 😄

jonsterling:
I'm going to close this because the signature apparatus has been rewritten; I have a feeling that it will be easier to do what you want under the new architecture... Happy to chat about it any time.

PR #102, opened 2016-08-10
begin implementing rules of cubical type theory

jonsterling:


jonsterling:
@carloon @jozefg 

One wrinkle is that dimension substitution seems to need to be defined on *metavariables*, and it needs to "block" until the metavariable is substituted with a term; and I'm not sure that this is possible with the current ABT framework. One place this comes up is in the intro rule for the identification type:

1. The first subgoal synthesizes/extracts an abstraction like `{u}[].M[u]()`. In later subgoals, we may refer to this by the metavariable `M` which has valence `{dim}[].exp` (this means, that in the syntax, the metavariable must occur applied to a *symbol* of sort `dim`). [This should raise red flags, since we really want something like a metavariable that gets applied to a dimension expression, i.e. `x|0|1`.]

2. The second subgoal needs to be something like `H >> M{u}()<0/u> = P0 : A<0/u>`.

However, dimension substitution can hardly do anything at this point! One option would be to have a notion of "cubical metavariable", with extended valences. I'd like to develop a cubical syntactic framework at some point, but now doesn't seem to be the right time to do that.

At first, I thought we could get around this problem is to use an *internal* dimension abstraction and application instead of the primitive substitution. In this case, the second subgoal would look like this: `H >> (<v>M{v}())@0 = P0 : A<0/u>`. However, I'm not sure even this would suffice! If you apply a computation rule, then we're kind of screwed again, since this will just turn into `M{v}()` (which is just what we *don't* want!).

--------------------------------------------------

This is kind of an impasse at the moment, I think... One thing I'm considering is generalizing the ABT library so that it can take symbols & renamings in some user-provided monad. Then, if we instantiate the framework at the `– + {0,1}` monad, we'd get precisely what we need (at the expense of some annoying bureaucracy in all cases of symbolic parameters which are not intended to denote dimensions).

Anyone have any suggestions?

cangiuli:
We've talked about this out-of-band, but FYI I am @cangiuli on GitHub.

jonsterling:
@cangiuli thanks! I'll start making a plan for how to move forward once I get back from IU.

jonsterling:
@cangiuli I've been thinking about it a bit more with @david-christiansen, and came to the tentative conclusion that it may be not too hard to design a "cubical abstract machine" which gives us the best of both worlds—in which the hcom *operators* would be stable under dimension substitution in the way that we need (but their canonicity would of course vary in the appropriate ways). I'll keep you updated!

PR #104, opened 2016-08-16
prettify (wow!)

jonsterling:
@david-christiansen look at this trick I learned! Before merging, I'll want to make this a non-default option in the mode, though.

### DIFF
--- a/emacs/redprl.el
+++ b/emacs/redprl.el
@@ -55,8 +55,6 @@
   "Syntax table for RedPRL.")
 
 
-;; 𝕯𝖊𝖋
-
 (defconst redprl-keywords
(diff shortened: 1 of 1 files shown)


david-christiansen:
What a kludge! Are you sure you want this in the mode, as opposed to the mode's documentation?

david-christiansen:
Also, `(foo . (bar bz))` means the same thing as `(foo bar baz)`.

jonsterling:
> What a kludge! Are you sure you want this in the mode, as opposed to the mode's documentation?

Ha, fair! I think this qualifies as the sort of weird hack that the total two people who would want this can just add to their own configurations...

david-christiansen:
:-)

PR #106, opened 2016-08-16
cl-lib was bundled since Emacs 24.3

syohex:


jonsterling:
@syohex So sorry I didn't see this! Thank you very much for the patch

Since I don't know emacs very well, I'm going to defer to @david-christiansen's judgment to see if this looks ok.

jonsterling:
I'm going to just merge this, since it sounds reasonable

PR #111, opened 2016-08-31
cubical redprl

jonsterling:
uphold-cubical-thought


jonsterling:
This is really coming along! Some tasks to do:

1. add validation to signature elaborator to ensure that terms don't have more free symbols & variables than they ought to
2. add parameter spine to custom operators
3. add dynamics for custom operators

jonsterling:
@RedPRL/politburo @ak3n @ayberkt I'd like to summarize what's happening with this project, and offer my sincere apologies if any of these changes are going to cause trouble for anyone.

In brief, RedPRL is becoming Cubical RedPRL, as part of the project being pursued by advisor Bob Harper and colleague Carlo Angiuli at Carnegie Mellon. This intention has led to a number of changes in RedPRL's implementation:

1. A new version of the abstract binding tree framework has been designed & developed, in order to support cubical dimensions (among other things).

2. We must be more cautious about judgments and types, because there are a number of judgments that we cannot internalize as a type anymore (e.g. equality), because they would not satisfy the Kan conditions for cubical types.

3. Most of the types that I have developed so far needed to be removed; as we discover Kan/cubical versions of the standard useful types, we will add them in.

4. Based on my experience working with @david-christiansen on implementing interactive modes for RedPRL (and also based on discussions with @ak3n), I have redesigned the signature elaboration mechanism to be much more suitable for these tasks, and also easier to change in the future. As a result, most of the existing interactive stuff has been removed—I'm very sorry for this @ak3n, but enough had changed that I don't think the existing stuff was recoverable. I'm happy to work with you to try and get it up and running again under the new order, if you want to do that.

5. I have replaced the hand-written parser with a proper generated one based on mlyacc; this is something that @ayberkt and I had discussed doing a few months ago, and I decided it was time to finally do it. I am sure there is a lot of room for improvement, so I welcome anyone to hack on it.

6. In the interest of pushing back some of the creeping bureaucracy that this codebase has accumulated, I have been trying to eliminate hierarchy and work with very simple file / folder structure.

**A few months ago, RedPRL was a toddler. But today, RedPRL is like a tiny little baby again.** I'm sorry for the churn, but I'm hopeful that by being willing to move forward quickly with a new idea, and go back to the drawing board when we learn a lesson the hard way, we can make RedPRL great with even the limited engineering resources that we have.

uphold-cubical-thought

PR #129, opened 2016-09-22
[WIP] Pretty printer for terms

ak3n:


src/redprl/pretty.sml
```diff
@@ -1 +1,28 @@
...
+  open Abt O Unparse
+
+  fun @@ (f, x) = f x
+  infix 0 @@ $ $$ \
+
+  fun notation m = 
+    case Abt.out m of
+      (* A -> B *)
+      MONO DFUN $ [_ \ a, (_, []) \ bx] =>
+        SOME (atom @@ toString a ^ " -> " ^ toString bx)
```
jonsterling: Great start! One correction:

The way that ABTs work here is that they have a fixed arity; this means that there is only one correct way to write a DFUN, and it is with a single variable bound; so, this case will never fire. The case below this will fire, though.

If you want to add separate printing for when the bound variable does not occur in b, you could do something like:

```sml
if Var.Ctx.member (varctx bx) x then (* ... *) else (* ... *)
```

ak3n: Hmm, are you sure?

I wrote this case becuase of this line from grammar:

```ml
| term RIGHT_ARROW term (Ast.$$ (O.MONO O.FUN, [\ (([],[]), term1), \ (([],[]), term2)]))
```

jonsterling:
There are two different operators: FUN and DFUN--the former computes to the latter.

ak3n:
Aaaaah, sorry for my inattention!

jonsterling:
No worries!! You're doing great :)

src/redprl/pretty.sml
```diff
@@ -13,14 +15,30 @@ struct
...
   infix 0 @@ $ $$ \
 
+  val paramToString = P.toString ShowSym.toString
+
   fun notation m = 
     case Abt.out m of
-      (* A -> B *)
-      MONO DFUN $ [_ \ a, (_, []) \ bx] =>
-        SOME (atom @@ toString a ^ " -> " ^ toString bx)
+      (* , x *)
```
ak3n: Not sure about this one.

ak3n:
Looks like I have covered `atomicRawTerm`. Should I continue with `rawJudgment`, `atomicRawTac`, etc?

src/redprl/pretty.sml
```diff
@@ -1 +1,46 @@
...
+        SOME (atom @@ "univ [ " ^ paramToString p ^ " ]")
+    | (* loop [ p ] *)
+      POLY (LOOP p) $ [] =>
+        SOME (atom @@ "loop [ " ^ paramToString p ^ " ]")
+    | (* A @ p @ *)
+      POLY (ID_AP p) $ [(([],[]) \ a)] =>
+        SOME (atom @@ toString a ^ " @ " ^ (paramToString p) ^ " @")
+    | _ => NONE
+
+  and toString m =
```
jonsterling: I think you will need to have a function here like

```
and unparse m = UP.unparse notation m
and toString m = parens (done (unparse m))
```

src/redprl/pretty.sml
```diff
@@ -1 +1,46 @@
...
+        SOME (atom @@ "< " ^ ShowVar.toString x  ^ " > " ^ toString a)
+    | (* univ [ p ] *)
+      POLY (UNIV p) $ [] =>
+        SOME (atom @@ "univ [ " ^ paramToString p ^ " ]")
+    | (* loop [ p ] *)
+      POLY (LOOP p) $ [] =>
+        SOME (atom @@ "loop [ " ^ paramToString p ^ " ]")
+    | (* A @ p @ *)
+      POLY (ID_AP p) $ [(([],[]) \ a)] =>
+        SOME (atom @@ toString a ^ " @ " ^ (paramToString p) ^ " @")
```
jonsterling: First, this will print `p @ a @`, whereas what we want is `a @ p`; secondly, can you use the infix mechanism from the unparsing library?

It would look something like this:

```sml
infix' (Left, 5, "@") (unparse a, atom (paramToString p))
```

I'm not sure whether `5` is the right number, but put anything you like there and we can tweak it if we need to.

src/redprl/pretty.sml
```diff
@@ -1 +1,46 @@
...
+
+  val paramToString = P.toString ShowSym.toString
+
+  fun notation m = 
+    case Abt.out m of
+      (* , x *)
+      POLY (HYP_REF x) $ [] => SOME (atom @@ ", " ^ ShowVar.toString x)
+    | (* A -> B *)
+      MONO FUN $ [_ \ a, (_, []) \ b] =>
+        SOME (atom @@ toString a ^ " -> " ^ toString b)
```
jonsterling: Here, we should use the infixing mechanism, so, it would be something like this:

```sml
infix' (Right, 3, "->") (unparse a, unparse b)
```

src/redprl/pretty.sml
```diff
@@ -1 +1,46 @@
...
+    | (* (x : A) -> B *)
+      MONO DFUN $ [_ \ a, (_, [x]) \ bx] =>
+        SOME (atom @@ "(" ^ ShowVar.toString x ^ " : " ^ toString a ^ ") -> " ^ toString bx)
```
jonsterling: Likewise, I think we can use the infixing thing here as well.

src/redprl/pretty.sml
```diff
@@ -1 +1,51 @@
...
+        end
+    | (* < x > A *)
+      MONO ID_ABS $ [(([x], []) \ a)] =>
+        SOME (atom @@ "< " ^ ShowVar.toString x  ^ " > " ^ toString a)
+    | (* univ [ p ] *)
+      POLY (UNIV p) $ [] =>
+        SOME (atom @@ "univ [ " ^ paramToString p ^ " ]")
+    | (* loop [ p ] *)
+      POLY (LOOP p) $ [] =>
+        SOME (atom @@ "loop [ " ^ paramToString p ^ " ]")
```
jonsterling: Can you remove the spaces that are in `ID_ABS,UNIV,LOOP`?

Issue #112, opened 2016-09-16
open head expansion rules


jonsterling:
- [x] open head expansion for equality
- [ ] head expansion for type equality
- [ ] can we do something analogous for truth? (hopefully?)

cangiuli:
What do you mean by, something analogous for truth?

jonsterling:
@cangiuli Something along the lines of "To show `H >> P true` when `H >> P => P'`, it suffices to show `H >> P' true`." (where `P => P'` is this restricted open computation judgment we've been kicking around—currently approximated only very conservatively)

cangiuli:
As we discussed today, the current head expansion tactic should re-check the head operator for dimension name indices after each step.

favonia:
I believe this thread is subsumed by #207 and also #230 about changing the machine.

PR #141, opened 2016-10-05
add a safe version of eval

jonsterling:
#112 

@Cangiuli This should hopefully address what we discussed this evening.

### DIFF
--- a/src/redprl/refiner.fun
+++ b/src/redprl/refiner.fun
@@ -1021,7 +1021,7 @@ struct
           case out m of
              th $ _ =>
                if List.exists (fn (_, sigma) => sigma = P.DIM) @@ Abt.O.support th then
-                 raise E.error [E.% "Cannot eval", E.! m, E.% "because it has free dimension variables"]
+                 NONE
                else
                  Machine.next sign st
            | _ => NONE
--- a/src/redprl/refiner.fun
+++ b/src/redprl/refiner.fun
@@ -1009,13 +1009,37 @@ struct
            Env.lookup rho (#1 goal))
       end
 
+    local
+      open Machine.S.Cl infix <: $
+    in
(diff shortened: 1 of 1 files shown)


jonsterling:
And nicely enough, this check doesn't break any of our examples.

cangiuli:
If the original term passes the check, but what it steps to fails the check, does the error cause the entire tactic to be rewound?

jonsterling:
@cangiuli Under the current behavior, what should happen is that the tactic steps until it cannot step any further (under the check). So, if `M1 :-> M2 :-> M3 :-> M4`, and `M1,M2,M3` are good but `M4` has a free dimension variable in the head operator, this rule will step `M1` to `M3`.

jonsterling:
@cangiuli I suppose it is a little trickier than that in fact—because this is a machine, this may result in computation occurring in the principle argument to some non-canonical operator—and this non-canonical operator may have free dimension variables. I wonder if that makes a difference?

Maybe it is necessary to have a more sophisticated version of this.

cangiuli:
Right, you want to check the principal argument's head constructor. Otherwise you conclude `S1-elim(loop[x]; 5, _.true)` is in bool even though `<0/x>` of it computes to `5`.

jonsterling:
@cangiuli OK, that much makes sense, and I think at least that behavior is implemented here...

The case that I was worried about in my last comment is, I think, probably not possible given that we always start from a "freshly loaded" machine state, and we check this condition prior to taking each step.

jonsterling:
By the way, this is one of the nice things about using a machine as opposed to SOS or (worse!) natural semantics—because there are no recursive calls in the implementation of the transition system (as would happen in SOS), we can interleave this check with evaluation easily & safely. Otherwise, we would have to implement the whole thing over again I think...

PR #142, opened 2016-10-22
begin switching to new dependent lcf version

jonsterling:


jonsterling:
One (understandable) consequence of correcting the monad multiplication operator in Dependent LCF is that things like "repeat" don't quite do what you might naïvely expect them to...

Recall that running `repeat { t }` will **repetitively sequence `t` with itself** *on each subgoal*. Now, this is not the same as **sequencing `t` with itself on each subgoal** *repeatedly*. The latter is the behavior that a user might expect, but it is not immediately clear how to get it under the new paradigm.

I think the right approach is to add recursion to the *multitactics*. I believe that this would allow the desired behavior. I will defer that to future work though...

PR #148, opened 2016-11-01
Fix source position handling in the lexer

wilcoxjay:
The lexer was previously inconsistent about maintaining the source
position. The position was only being updated on newlines and
whitespace, and not when any real tokens were processed.

This problem was partially masked by the fact that the ending position
of each token was correct relative to the start position, but the end
position was not being written back to the global variable that keeps
track of the current position.

This patch factors out the update logic into a few helper functions.
* `posTuple n` takes the length of the current token as an argument, and
  returns the (start,end) position information while writing the end
  position back to the global variable `pos`.
* `posTuple n x` additionally takes an arbitrary value `x`, and returns
  the tuple `(x, start, end)`. This is useful when the token takes
  additional arguments beyond just the positions.

### DIFF
--- a/src/redprl/redprl.lex
+++ b/src/redprl/redprl.lex
@@ -8,6 +8,24 @@ type arg = string
 val pos = ref Coord.init
 val eof = fn (fileName : string) => Tokens.EOF (!pos, !pos)
 
+fun incPos n = pos := (Coord.addchar n o (!pos))
+
+fun posTuple n =
+  let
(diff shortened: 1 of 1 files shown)


wilcoxjay:
As a side effect, this appears to improve MLton compile times on my machine by a factor of almost 4...

jonsterling:
oh no! nice job catching this 😄

PR #149, opened 2016-11-01
[parser] give function applications source pos

wilcoxjay:
The nonterminal atomicRawTerm does not do annotation, and expects the
parent to do so. However, rawTerm directly called atomicRawTerm,
breaking this expectation.

The fix is to introduce another intermediate nonterminal, called
atomicTerm, which sits between atomicRawTerm and rawTerm. The naming
convention here is that "raw" means "potentially unannotated".

### DIFF
--- a/src/redprl/redprl.grm
+++ b/src/redprl/redprl.grm
@@ -86,10 +86,12 @@ end
 %nonterm
    start of Signature.sign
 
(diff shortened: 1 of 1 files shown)


jonsterling:
@wilcoxjay Excellent catch; this may explain some strange behavior I noticed before but couldn't track down.

Issue #110, opened 2016-08-18
function to get list of *occurrences* of free variables


jonsterling:
(i.e. with their annotations)

jonsterling:
This will be useful for providing better error messages; when a variable occurs unbound, we could underline the exact spot in Emacs.

This ticket should probably be addressed in [sml-typed-abts](https://github.com/RedPRL/sml-typed-abts/), by adding operations to the `ABT` signature.

ak3n:
What is purpose of [`varctx`](https://github.com/RedPRL/sml-typed-abts/blob/master/src/core/abt.fun#L206)?

jonsterling:
@ak3n This gives you the collection of free variables in a term, along with their sorts.

wilcoxjay:
I'd like to take a crack at this one. Here's what I'm thinking, based on the proposal above. Add operations
```
metaOccurrences : abt -> annotation Metavar.ctx
varOccurrences : abt -> annotation Var.ctx
symOccurrences : abt -> annotation Sym.ctx
```
to the ABT signature in `sml-typed-abts`.

Then, use these operations to get better error messages in [checkScope](https://github.com/RedPRL/sml-redprl/blob/93684fa8ad3dc16b3e298386e857fa822ce7a36e/src/redprl/signature.sml#L243). In two cases, that code only cares about the key but not the value in the mapping (ie, the variable, but not its sort/valence). In those cases, we can just replace `*ctx` with `*Occurrences`. But in the case of symbols, that code *does* care about the value, and so we need to call both of them. 

An alternate design would be to change the existing `*ctx` operations to return something like `(sort * annotation) Var.ctx`. This has the dual downside of forcing every client to explicitly ignore position information when they don't need it. I'm also not sure how this alternate design would affect the internals of `sml-typed-abts`, which makes several uses of the `*ctx` operations.

jonsterling:
I like your first proposal... Though it would probably need to be something like `metaOccurrences : abt -> annotation list Metavar.ctx` or something along those lines, since a variable can occur multiple times (and the contexts only include a variable once).

wilcoxjay:
Oh, right, that makes sense.

wilcoxjay:
This is closed by #150.

I'm going to open a similar issue wrt metavariables

PR #150, opened 2016-11-01
Give nicer error messages (with location info) on unbound variables and symbols

wilcoxjay:
~This is a work in progress. Please do not merge yet.~ This is ready to merge after approval.

Addresses #110.

Depends on RedPRL/sml-typed-abts#72.

I have made some progress here. So far, I've just done it for "normal" variables. Any early feedback on that part is welcome.

When trying to take the next step by implementing this for meta variables, I ran into the failure illustrated by the test in the most recent commit. Basically, it looks like ABT conversion fails out if there's a free metavariable, so we have no chance in checkScope to give a nice error message. Any ideas on what to do here?

### DIFF
--- a/src/redprl/signature.sml
+++ b/src/redprl/signature.sml
@@ -243,15 +243,19 @@ struct
     fun scopeCheck (metactx, symctx) term : Tm.abt E.t =
       let
         val termPos = Tm.getAnnotation term
+        val symOccurrences = Susp.delay (fn _ => Tm.symOccurrences term)
         val checkSyms =
           Tm.Sym.Ctx.foldl
             (fn (u, tau, r) =>
               let
                 val tau' = Tm.Sym.Ctx.find symctx u
                 val ustr = Tm.Sym.toString u
+                val pos = case Tm.Sym.Ctx.find (Susp.force symOccurrences) u of
+                               SOME (pos :: _) => SOME pos
+                             | _ => (print ("couldn't find position for var " ^ ustr); termPos)
               in
-                E.when (tau' = NONE, E.fail (termPos, "Unbound symbol: " ^ ustr))
-                  *> E.when (Option.isSome tau' andalso not (tau' = SOME tau), E.fail (termPos, "Symbol sort mismatch: " ^ ustr))
+                E.when (tau' = NONE, E.fail (pos, "Unbound symbol: " ^ ustr))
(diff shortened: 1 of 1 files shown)


jonsterling:
@wilcoxjay yes, it is necessary to know what the valences of the metavariables are in order to even do the conversion IIRC, so there is not a whole lot that can be done there I think... But I'd have to look at it to be sure.

wilcoxjay:
I've extended this approach to work with symbols as well.

I've confirmed through tests (not pushed) that this code correctly localizes free symbols in both applied operators and in metavariable applications. As noted in the comments in the `symOccurences` commit on RedPRL/sml-typed-abts, the annotations are not precise enough to specify exactly the position of the free symbol, but only its enclosing application operator. This is still an improvement over the previously given position, which ranged over the entire enclosing definition.

By the way, it would be nice to have a slightly more proper test suite. Especially for tests that should fail, it feels weird to put them in `test/test.prl`, which also seems to serve as a place for examples for newcomers.

At this point, I'd like to incorporate any feedback you have and then merge this without support for getting the locations of free metavariables. As discussed above, free metavariables cause AST-to-ABT conversion to fail earlier than we can do anything about it. This makes it hard to test. However, if you'd like to include the `metaOccurrences` function in the RedPRL/sml-typed-abts interface for uniformity, I'm happy to give it a try.

jonsterling:
Thanks for all your work! 😄 

- I've merged in the changes in the ABT lib, so you can update the submodule here.

- Regarding tests, it would be good to have some real tests about what should fail, etc.—and run these in CI such that the build succeeds of the test fails and fails if the test succeeds. However, I do not want to have fine-grained stuff in there that looks for a specific error message (that increases the maintenance burden too much).

In the meanwhile, `test.prl` has served as a sort of sanity-testing thing for stuff I want to make sure works (and failing things shouldn't go in this file). If we add real tests, we might rename `test.prl` to something else.

wilcoxjay:
I updated the submodule (I think I did it right...), and the build now passes.

src/redprl/signature.sml
```diff
@@ -243,23 +243,35 @@ struct
     fun scopeCheck (metactx, symctx) term : Tm.abt E.t =
       let
         val termPos = Tm.getAnnotation term
+        val symOccurrences = Susp.delay (fn _ => Tm.symOccurrences term)
```
jonsterling: Good idea to use a suspension!

PR #156, opened 2016-11-22
Print the build log on failure

wilcoxjay:
The recent changes to `script/test.sh` completely silenced build output, making it hard to debug travis failures. 

This PR prints the build log if the build fails.


jonsterling:
thanks!

PR #157, opened 2016-11-22
Incorporate annotations in elaborator's wrap

wilcoxjay:
The RedPRL error reporting facility allows the code throwing the error to specify a position, but such positions would be ignored everywhere, as far as I can tell.

This PR makes the elaborator's `wrap` function check to see whether the thrown exception contains a position. If so, it uses it; otherwise, it falls back on the position passed in as an argument. 

This is useful, for example, when wrapping a function that traverses a term. If an error occurs during the traversal, we may have more precise position information than "somewhere in the tree".

### DIFF
--- a/src/redprl/elab_monad.sml
+++ b/src/redprl/elab_monad.sml
@@ -102,7 +102,10 @@ struct
 
   fun wrap (pos, f) =
     Debug.wrap (fn _ => ret (f ()))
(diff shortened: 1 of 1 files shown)


jonsterling:
This looks like the right thing—let's merge it in and see what happens.

Issue #146, opened 2016-10-31
Allow RedPRL to read from stdin instead of file


jonsterling:
See https://github.com/freebroccolo/vscode-redprl/issues/2

PR #159, opened 2016-11-23
Parse signatures element-by-element

### LINKED ISSUE #147: Incrementalize parsing to support error recovery
As discussed in freebroccolo/vscode-redprl#1, it would be nice if RedPRL continued to process the rest of the file when it encounters a syntax error.

Because of lack of support in mlyacc for this sort of thing, I think we need to use mlyacc to parse one declaration/command at a time. Then, to parse a signature, we can call this in a loop that just continues when it encounters a parse error. The loop will have to take some recovery procedures on error, for example by scanning the token stream for the next declaration/command token before calling the parser again.

wilcoxjay:
Fixes #147.

In order to support better recovery from syntax errors, this PR changes the parser to parse signature *elements*. (An element is defined to be either a command or a declaration.) A signature is parsed "by hand" by calling the element parser in a loop.

When encountering a syntax error, the loop scans the token stream until it sees a token that is the beginning of some element, and then tries to parse another element starting there. So a signature with syntax errors can have most of its parts checked.

If any parse errors are encountered while processing a signature, the signature is considered to have failed to check.

The test `test/failure/incremental-parsing.prl` shows a signature consisting of a syntax-correct theorem, then a theorem with a syntax error, and then another syntax-correct one. The changes in this PR allow RedPRL to check the two syntax-correct theorems while printing an error for the one with a problem.

I put most of the "by hand" parser in `src/frontend/frontend.sml`, but it might make more sense to move it somewhere in `src/redprl/`, where the rest of the parsing infrastructure lives. Suggestions welcome.

### DIFF
--- a/src/frontend/frontend.sml
+++ b/src/frontend/frontend.sml
@@ -25,17 +25,87 @@ struct
             (RedPrlError.error [E.% s])
   end
 
-  fun parseSig fileName =
-    let
-      val input = TextIO.inputAll (TextIO.openIn fileName)
-      val lexer = RedPrlParser.makeLexer (stringreader input) fileName
-      val (sign, _) = RedPrlParser.parse (1, lexer, error fileName, fileName)
-    in
-      sign
-    end
+  fun parseElt fileName lexer =
+    RedPrlParser.parse (0, lexer, error fileName, fileName)
+
+  local
+    open Signature
+  in
+    fun processElt sign (DECL (nm, d, pos)) = Signature.insert sign nm (d, SOME pos)
+      | processElt sign (CMD c) = Signature.command sign c
+  end
+
+  fun logExn exn =
+    RedPrlLog.print RedPrlLog.FAIL (RedPrlError.annotation exn, RedPrlError.format exn)
+
+  local
+    val EOF = RedPrlLrVals.Tokens.EOF (Coord.init, Coord.init)
+    val DEF = RedPrlLrVals.Tokens.DCL_DEF (Coord.init, Coord.init)
+    val THM = RedPrlLrVals.Tokens.DCL_THM (Coord.init, Coord.init)
+    val TAC = RedPrlLrVals.Tokens.DCL_TAC (Coord.init, Coord.init)
+    val PRINT = RedPrlLrVals.Tokens.CMD_PRINT (Coord.init, Coord.init)
+
+    fun isBeginElt tok =
+      RedPrlParser.Token.sameToken (tok, THM) orelse
+      RedPrlParser.Token.sameToken (tok, DEF) orelse
+      RedPrlParser.Token.sameToken (tok, TAC) orelse
+      RedPrlParser.Token.sameToken (tok, PRINT)
+
+    fun isEof tok =
+      RedPrlParser.Token.sameToken (tok, EOF)
+
+    fun skipToBeginElt lexer =
+      let
+        val (next_tok, lexer') = RedPrlParser.Stream.get lexer
+      in
+        if isEof next_tok orelse isBeginElt next_tok
+        then lexer
+        else skipToBeginElt lexer'
+      end
+
+    fun recover lexer =
+      let
+        val (_, lexer) = RedPrlParser.Stream.get lexer
+      in
+        skipToBeginElt lexer
+      end
+
+    fun doElt fileName lexer sign =
+      let
+        val (elt, lexer) = parseElt fileName lexer
+      in
+        (true, processElt sign elt, lexer)
+      end
+      handle exn => (logExn exn; (false, sign, recover lexer))
+  in
+    fun parseSig fileName =
+      let
+        fun loop acc lexer sign =
+          let
+            val (next_tok, _) = RedPrlParser.Stream.get lexer
+          in
+            if RedPrlParser.Token.sameToken (next_tok, EOF)
+            then (acc, sign)
+            else
+              let
+                val (err, sign, lexer) = doElt fileName lexer sign
+              in
+                loop (acc andalso err) lexer sign
+              end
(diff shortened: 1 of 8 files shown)


wilcoxjay:
@freebroccolo If it's easy to test this branch with the vscode stuff without merging to master, could you go ahead and give it a try?

freebroccolo:
@wilcoxjay Nice. Seems to work.

redprl

freebroccolo:
> I put most of the "by hand" parser in src/frontend/frontend.sml, but it might make more sense to move it somewhere in src/redprl/, where the rest of the parsing infrastructure lives. Suggestions welcome.

I don't have any concrete suggestions but it might make sense to start thinking about how this stuff could be generalized to make `redprl` work a little bit more like merlin.

For instance, if we were to make `redprl` into a process that listened to commands over stdio or over a socket, then the incrementalized parsing you've implemented would be subsumed by a driver that reads a file and updates the `redprl` state with elements it successfully parses.

So one of the commands would be like `addElement` or something. But we would also need to support document editing commands which trigger reparsing of blocks of text sent by the editor along with ranges in the original document.

These edits could result in elements being modified, added to, or removed from the state depending on the affected ranges. Removals would occur when an edit is processed consisting of an empty block of text and a non-empty range.

Each time one of these edits would be processed, all of the following elements would need to have their associated ranges updated in the state too.

You can see the protocol that merlin uses [here](https://github.com/the-lambda-church/merlin/blob/master/doc/dev/PROTOCOL.md). In particular, see the [tell](https://github.com/the-lambda-church/merlin/blob/master/doc/dev/PROTOCOL.md#tell) command. If we were to do this, we'd want to basically always use the contextual command variants (see toward the end) which specify which file the commands should be processed with respect to and we'd want to use JSON-RPC messages from the start.

Anyway, it would take quite a bit of work to do all of this but in the meantime maybe it can give you some useful ideas about how to proceed.

wilcoxjay:
Yes, that's a good thing to think about. 

As an intermediate step, this is also vaguely related to #146 and freebroccolo/vscode-redprl#2. I may have a crack at that next.

For now, this PR doesn't change much about the way RedPRL process developments. For example, the signature is not *checked* incrementally. We still build up the whole signature and then check it. To really implement a merlin-like thing, as you suggest, we would also need to incrementalize signature checking. (This should not be overly difficult.)

jonsterling:
wow!! This is cool!

wilcoxjay:
@freebroccolo In response to your edit above: It seems to me that the essential thing is to start producing feedback without saving. The bit about not needing to re-parse is probably not important given the current scale of RedPRL developments. Is that right, or am I missing something?

In that case, I think just enabling RedPRL to read from stdin gives us the important part.

freebroccolo:
@wilcoxjay it depends on how you look at it I suppose. Technically, it should be enough for `redprl` to read from stdio, especially since there are no multi-document redprl developments yet. If we ran `redprl` every idle 500ms it would still be a lot better than what we have now where feedback is tied to file saves. So that is definitely worth having.

But one thing that is less than ideal about the current approach is that it pushes all of the complexity for supporting interactive features into the editor plugin in a kind of ad hoc way. We need to record all of the data that `redprl` dumps and then be able to provide it to the various widgets on user interaction. The stdio solution doesn't help with that part.

In the long run, the more interactive features we want to support in the editor, the more we'll end up approximating something like this `redprl` server component anyway, it will just be this implicit thing built into the plugin rather than a part of RedPRL itself.

But maybe that's okay for now. We could even build an independent node-based language server that could even be used by other editors. On the surface it would provide a more interactive interface but behind the scenes it just be continually reprocessing the files with the `redprl` binary.

Eventually, some features would require support in the core but that could be dealt with later.

So having feedback during editing is the most important part of it for now but theres a bigger issue here about how to make interactive features easier to support in general. I would say that it's probably not worth making radical changes to the core yet just to make the interactive stuff a bit nicer. But it is something to keep in mind regarding future changes or additions to the core.

wilcoxjay:
Oh, that's a great point! I totally buy that most of that complexity should be handled in an editor independent way.

I'm not sure whether it makes sense to (eventually) implement it in RedPRL proper or as a separate server, as you suggest. I think either could work, so it's probably whatever's easiest for maintenance purposes. 

In any case, I've implemented the stdin thing in #160 if you want to have a look!

freebroccolo:
@wilcoxjay I think I may have found an issue or bug with this implementation. It seems that when a parse error is encountered, the diagnostics for the element prior to the error are no longer reported. 

I've included some screenshots that demonstrate this. In the first screenshot, two items have goals remaining. But when a parse error happens immediately following the second item, we no longer receive that information on refresh.

Is it possible to fix this?

redprl-0

redprl-1

freebroccolo:
Another little thing I just noticed is that it looks like the goal numbering is getting reset for each item. Not sure if that is related to this change or not though.

wilcoxjay:
@freebroccolo Hm, yeah. 

Without having tried it yet, I think this is actually a function of the specific *kind* of syntax error you used in that example. For technical reasons the tokens `Thm`, `Def`, `Tac`, and `Print` are treated (along with EOF) as terminators for an item. Since your example essentially removes the `Thm` tactic from the stream, the parser gets confused because it can't find the end of the previous item. So it actually thinks those two items together are one big parse error. 

To confirm that this is the problem, you could try various other kinds of syntax errors and see if you see any weird behavior.

In any case, one idea on how to fix this is to not treat those tokens specially, but to treat the `.` token specially instead. When I was originally implementing this, I went with the current design because it meant less parsing had to be done by hand. But maybe this is a reason to do it the other way, since there's no way to have only part of a `.` token, we won't run into the same problem.

PR #160, opened 2016-11-24
Support reading input from stdin

wilcoxjay:
Fixes #146. Makes progress on freebroccolo/vscode-redprl#2.

Depends on #159. ~I'll clean up this branch after that PR is merged.~ EDIT: Done.

This PR adds a command line option `--from-stdin` to RedPRL. From the help message:
```
--from-stdin[=filename]   Read signature from stdin with optional diagnostic filename
```
The `=` syntax made it easier to work with in the (lack of a 😛 ) command line argument parsing library in RedPRL. I hope it isn't too hard for users or tools to adapt to.

PR #161, opened 2016-11-25
Refactor parser to improve error reporting when start-of-item token is malformed

wilcoxjay:
As @freebroccolo pointed out in the comments of #159, there was an issue caused by the parser using the start-of-item tokens (`Thm`, `Def`, etc.) as terminators for the previous item. 

When these tokens were malformed (eg, `hm` instead of `Thm`), RedPRL would see the item before the malformed token and the item starting with the malformed token as one big messed up item, which caused it not to analyze the first item. This is confusing because it intuitively seems like the first item is well formed and should be processed.

This PR uses the '.' token as the separator, which removes this confusing behavior.

For convenience, this PR depends on #160. So I'll clean up the branch once that's merged.

I've reproduced the old issue on my machine in flycheck/emacs and checked that this fixes it.

@freebroccolo: Could you just double check that this also fixes what you were seeing in vscode, and that there are no further issues of this sort that immediately pop up for you?

### DIFF
--- a/src/frontend/frontend.sml
+++ b/src/frontend/frontend.sml
@@ -20,13 +20,10 @@ struct
   local
     structure E = RedPrlError
   in
-    fun error fileName (s, pos, pos') : unit =
+    fun error fileName (s, pos, pos') =
       raise E.annotate (SOME (Pos.pos (pos fileName) (pos' fileName)))
             (RedPrlError.error [E.% s])
   end
-
-  fun parseElt fileName lexer =
-    RedPrlParser.parse (0, lexer, error fileName, fileName)
 
   local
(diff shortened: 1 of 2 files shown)


freebroccolo:
@wilcoxjay thanks. That seems to fix it:

vscode

jonsterling:
@wilcoxjay thanks! I've merged the PR this one depended on now.

wilcoxjay:
Great, thanks! I'll clean this up and merge tonight when I get back to my computer.

jonsterling:
@wilcoxjay thanks again for taking this on, comrade! Great job so far 😄

PR #162, opened 2016-11-28
Improve handling of lexical errors

wilcoxjay:
This PR makes the lexer more idiomatic by matching all possible input. (See Section 7.3 of the [mllex user manual](http://rogerprice.org/ug/ug.pdf).)

The lexer now just skips characters it doesn't recognize, while printing an error. 

This change also means that lexical errors are reported using the normal RedPRL error mechanism, so they show up as expected in IDEs.

### DIFF
--- a/src/redprl/redprl.lex
+++ b/src/redprl/redprl.lex
@@ -26,8 +26,6 @@ fun posTupleWith n x =
     (x, l, r)
   end
 
-exception LexerError of pos
-
 %%
 %arg (fileName : string);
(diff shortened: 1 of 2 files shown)


jonsterling:
@wilcoxjay I think this looks reasonable, thanks!

@freebroccolo what do you think?

freebroccolo:
Looks pretty straightforward. Seems like a good idea to me. 👍

jonsterling:
@freebroccolo thanks!

PR #164, opened 2016-11-29
Prototype notations for unbound variables in Pi and Sigma

### LINKED ISSUE #163: Unused variables in various binders
It would be nice if RedPRL supported various common notations for unused bound variables. 

As a first cut, allowing `_` to be used in binding position to signify that the bound variable will not be used.

Even better, some syntactic forms have well known shorthand notations for the case where the bound variable is not used. For example, `A -> B` instead of `(_ : A) -> B`.

wilcoxjay:
Fixes #163.

### DIFF
--- a/src/redprl/redprl.grm
+++ b/src/redprl/redprl.grm
@@ -42,6 +42,7 @@ end
    EOF
  | VARNAME of string
  | OPNAME of string
+ | UNDER
  | NUMERAL of int
  | COLON
  | LANGLE | RANGLE
@@ -104,6 +105,8 @@ end
  | term of ast
  | terms of ast list
  | funapp of ast
+
+ | boundVar of string
 
    (* abt abstractions *)
  | binder of symbol list * symbol list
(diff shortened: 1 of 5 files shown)


jonsterling:
I think this looks reasonable; somehow I thought it would be harder to do this, but I don't know what I was thinking.

I also don't really understand the significance of shift-reduce warnings etc., and whether they are a thing that one should worry about. I'm not an expert in parsing, and so far it's been enough to even get the parser to work, much less be conflict free. @freebroccolo Do you know what these shift/reduce and reduce/reduce conflicts mean, and whether we should worry about them?

freebroccolo:
@jonsterling Both of these conflicts mean there are ambiguities in the grammar. Usually this ambiguity has to do with a problem being able to detect where a certain sequence of tokens should end or begin some syntactic entity. When these conflicts happen, they are resolved automatically with some tie-breaking heuristic (e.g., longest possible parse, order of applicable rule in the grammar file, etc.).

It is worth fixing these conflicts I think but it will almost certainly require making the parser more complicated in some places by adding more states or other refactoring.

wilcoxjay:
Okay, I'm going to open an issue about the conflicts. Let's put this PR on hold until we understand how it impacts those.

wilcoxjay:
I've temporarily rebased this branch on top of #167 and fixed it to introduce no new conflicts. So this is in good shape as soon as we decide what to do with the rest of the parser.

wilcoxjay:
I've updated this again now that the parser changes have landed.

Issue #166, opened 2016-11-29
Fix parser conflicts


wilcoxjay:
The RedPRL parser currently reports a large number of shift/reduce and reduce/reduce conflicts. 

Here is my limited understanding of the classes of conflicts:
* mlyacc resolves shift/reduce conflicts by reporting the conflict and then choosing the shift. These are usually relatively easy to fix by adding precedence or associativity annotations. 
* mlyacc resolves reduce/reduce conflicts by reporting the conflict and then choosing the rule that appears earlier in the grammar file. These usually indicate some kind of "real" ambiguity in the grammar and may require some actual decision to be made.

The only remaining conflict to fix is:
- [x] between custom operators and grouping parentheses, described [here](https://github.com/RedPRL/sml-redprl/blob/76cb4750f9d7be063ca4aad37be940bebd5cf782/test/success/conflict.prl#L1-L5).

wilcoxjay:
I fixed most of the conflicts in PR #167. There were a few worth noting here, and there's one main one left.

* The one remaining cause of error is an ambiguity between custom operators and using parentheses for grouping. See theorem `custom-op-versus-parens` in `test/success/conflict.prl`. (Note that `conflict.prl` is intentionally marked as "succeeding" despite failing currently. It will succeed once we resolve the conflict.)

* There are two ways to convert a multitactic to a tactic:
    * Via <| |>, which is not currently used anywhere in the test suite
    * Implicitly, via the `rawTactic` rule, which is used extensively

   Was the implicit conversion intended to be used so pervasively? If so, is there any reason to keep the <| |> construction?

* The empty string is parsed as the identity tactic, which causes ambiguity in various places. For example, in a multitactic context, `[]` could conceivabily be parsed as the "empty" `each` mtac, or the singleton `each` mtac, with `id` as its only element. The parser currently chooses the latter, but its not clear that it should. This results in a "rule not reduced warning", because the parser never needs to produce the empty list of tactics.

* Identifiers are valid tactics, but I can't figure out what for (no uses in test suite). This causes a shift/reduce conflict with hypBindings. For now, I've resolved it by deleting the rule that parses IDENTs as tactics.

wilcoxjay:
I should also note that I was perhaps a bit too hasty with some of these changes -- I didn't stop to carefully test the behavior before and after. So, especially for constructs that weren't used anywhere, I may have broken something.

jonsterling:
[...]

PR #167, opened 2016-12-01
Fix most parser conflicts

wilcoxjay:
See discussion on #166


src/redprl/redprl.grm
```diff
@@ -80,7 +80,8 @@ end
 
 %nonassoc SQUIGGLE
 %right LEFT_ARROW RIGHT_ARROW DOUBLE_PIPE SEMI
-%left FUN_APP AT_SIGN
+%left FUN_APP
+%left AT_SIGN
```
jonsterling: for my reference, what does this change do?

src/redprl/redprl.grm
```diff
@@ -358,14 +368,13 @@ hypBindings
...
-  : multitac (Tac.multitacToTac multitac)
+  : multitac %prec SEMI (Tac.multitacToTac multitac)
 
 tactic
   : rawTactic (annotate (Pos.pos (rawTactic1left fileName) (rawTactic1right fileName)) rawTactic)
 
 tactics
-  : ([])
-  | tactic ([tactic])
+  : tactic ([tactic])
```
jonsterling: good idea, it seems like it was not necessary to have an empty one here

test/success/conflict.prl
```diff
@@ -0,0 +1,12 @@
+// The term syntax `f(foo)` is ambiguous: is it a custom operator f, or is
+// it a variable f with foo surrounded by parens.
+Thm custom-op-versus-parens : [ (f : (x : bool) -> bool) -> bool true ] by [
+  'lam([f]. f (f tt))
```
jonsterling: Ahh, this is a very good question, and it's not quite clear what to do here. I think that in the end, we will want to have a stage of parsing that comes even before AST, at which we have some sort of ambiguous "identifier applied to a spine of terms" constructor, which the elaborator interprets in the appropriate way based on the information it has about the current signature and local context—choosing either a custom operator application, a metavariable application or a function application. I don't think it's crucial to fix this right now, but we should keep it on the radar.

src/redprl/redprl.grm
```diff
@@ -333,23 +338,28 @@ atomicRawTac
 
   | operator LPAREN bindings RPAREN (Ast.$$ (operator, bindings))
   | metavar metavarArgs (Ast.$$# (metavar, metavarArgs))
-  | IDENT (`` IDENT)
```
jonsterling: Actually, I don't think we should remove this.

src/redprl/redprl.grm
wilcoxjay: Short version: this makes the `@` symbol bind more tightly than function application, which fixes some conflict I ran into at some point. I can dig up more details if you want.

It's not actually clear to me which one should bind more tightly, but I tried this and it seems to work. Best justification for anything :)

jonsterling: Ah! Thanks; I somehow thought that putting them on the same line would give them precedence in the order you wrote them, but clearly that wasn't the case 😄  This actually explains some confusing things that happened to me early on. thanks!

src/redprl/redprl.grm
wilcoxjay: Sounds good. I think this might have just been a red herring during development: at one point it fixed a bunch of conflicts, but it looks like later changes have obviated the need to remove it. I'll put it back now.

src/redprl/redprl.grm
```diff
@@ -338,6 +338,7 @@ atomicRawTac
 
   | operator LPAREN bindings RPAREN (Ast.$$ (operator, bindings))
   | metavar metavarArgs (Ast.$$# (metavar, metavarArgs))
+  | IDENT (`` IDENT)
```
jonsterling: thanks! the reason for having this, by the way, was that there are place where we bind a variable that ranges over a tactic; this happens, for instance, in the fixed point tactical. I don't think this tactical is used in the examples, because in practice you use a fancier tactical which is derived from it, called `REPEAT`.

jonsterling:
One question—I notice that we have a failing test in the `success` folder. Two things:

1. Should this instead be in the `failure` folder?
2. Why is the unexpected result (test expected to succeed but fails) not making travis report an error?

wilcoxjay:
Good points. 

Second point first: oops! The test script does not correctly set the exit code. I submitted PR #168 for this. I will rebase this PR on top of that branch for now so that we can get useful feedback from travis.

Back to the first point: I added the failing test to the success folder initially because I was planning to fix it before merging.

I still think this is a pretty serious bug that we should consider fixing now rather than later. But if we decide it's too much work, then we can come back to it later. 

If we decide to go that route, I might make a "would be nice to have passing" test folder, for this kind of thing. The idea being that we could have travis report failures from this folder, but not actually fail the build because of it. I don't want to put this test in the failure folder, because I believe it should pass, we're just not there yet.

jonsterling:
@wilcoxjay Thanks for your reply! I've merged in the PR about fixing the error code.

Regarding the conflict between different kinds of application, I agree it's something we should fix sooner rather than later—but I can't think of a solution to it that doesn't either involve (1) a ton of work, or (2) changing the concrete notation of RedPRL in a confusing way. But I'm interested if you have any ideas...

Things like Agda and Coq have got away with avoiding this problem by not allowing user-defined operators of non-empty arity, and instead leaning on their intrinsic type systems via function types. In our setting, that's not an option, so we need to have notation for both—either the notation should coincide and the elaborator can choose which it should mean, or the notation should be different (but then, what should it look like?).

wilcoxjay:
Yeah, I agree this is a tough problem. I think you've mentioned already the two basic possibilities:
1. Change something in the syntax to make it more obvious to the parser when something is an operator. For example, we could:
    * use something other than parens to group operator arguments (no great options here since `[]`,`{}`, and `<>` already have other meanings)
    * use some sort of "sigil" to designate operator names (kind of like `#` is used for metavariables right now)
    * require operator names start with a capital letter and require variables to start with a lower case letter
    * require a terminating semicolon on operator arguments (I think this works, since currently the only "real" ambiguity is when you have an operator with a single argument with trivial valence; this change would ask you to write that as `foo(arg;)`, which disambiguates it from just grouping parens around `arg`. This might still be hard to parse, but at least it should be possible in principle.)
2. Use contextual information to disambiguate the current syntax.
    * As you suggested earlier, we could parse into some pre-AST format and then use information in the signature and the local context of bound variables to decide what the concrete syntax meant. 
    * We could also try to add some kind of state to the parser that keeps track of the operators and bound variables in scope, and use this to do the disambiguation at parse time. (This is probably a bad idea, since we repeat some of this work later anyways, and it makes the parser significantly more complicated.)

In any case, it would be nice to have a short-term work around. The best I've come up with so far is, whenever you want to write `f (g x)`, instead write `let y = g x in f y`.

freebroccolo:
> use some sort of "sigil" to designate operator names (kind of like # is used for metavariables right now)

> require operator names start with a capital letter and require variables to start with a lower case letter

One of these two would be my preference.

I used to dislike the caps convention for languages like Haskell and OCaml but after spending a lot of time working on syntax highlighting and other editor related features I've come to appreciate simple solutions like that. You can always rely on some analyzer or refiner or whatever to disambiguate, like Agda does with syntax highlighting, but that approach has a lot of drawbacks (e.g., related to complexity, efficiency, etc.).

Another reason to prefer one of these two out of the others is that they would not require any backtracking (i.e., looking ahead to some special disambiguating token). Minimizing backtracking is good for highlighting since most popular highlighting implementations only support backtracking within a single line. It's also good for interactive features since you can get feedback sooner and again the parsing is simpler.

wilcoxjay:
I thought of another simple-but-kind-of-devious way to disambiguate it: disallow a space between the operator name and the opening paren; require a space between a variable applied to a grouped argument.

jonsterling:
The capital letter convention might be the most reasonable choice... I guess we'd also have to decide what to do about built-in operators—presumably these (e.g. `auto`, `repeat`, etc.) should be renamed as well.

How does everyone feel about going that direction?

wilcoxjay:
Sounds good.

I could go either way wrt builtin operators. The already "feel" different from user-defined operators, because you don't have to write the `()` after them. So we could just leave them the way they are. 

If we decide to change them to feel like user-defined operators, I think we should go all the way and require the `()` as well.

freebroccolo:
The capital letter convention sounds good to me. I don't have much of an opinion about built-ins. If they are known ahead of time, these can always be treated specially in the syntax highlighting or other tooling so it's not as big of an issue there.

jonsterling:
OK, great! Let's go with the capital letters, and don't change built-in operators for now.

wilcoxjay:
Okay, I'll update this PR with those changes in a bit.

jonsterling:
this is looking great!

test/examples.prl
```diff
@@ -62,28 +62,51 @@ Thm pathTest2 : [ paths({x}. (b:bool) -> bool; lam([b].b); lam([b].boolEta(b)))
...
+// Def singleton(#A : exp; #a : exp) : exp = [
+//   (x : #A) * paths({_}. #A; x; #a)
+// ].
+
+// Thm boolSingletonsContractible : [
+//     (a : bool) ->
+//     (p : singleton(bool; a)) ->
+//     paths({_}. singleton(bool; a); p; a) true ] by [
+//   { lam a. lam p. <i> };
+// ].
```
wilcoxjay: oops... one second :)

wilcoxjay:
I made the changes we discussed. 

I split the terminal `IDENT` into `VARNAME` (lowercase) and `OPNAME` (uppercase). I added a nonterminal `ident` which parsers either one (useful in places where we can allow either without ambiguity). 

The test `success/conflict.prl` now passes as expected.

I had to go through and make a decision for all the uses of `IDENT` in the old parser, whether to replace it with `VARNAME`, `OPNAME`, or `ident`. Some decisions were obvious, but some others I'm not sure about.

For example, I don't think there's any reason we need to require `Thm` or `Tac` names to be uppercase, but it just seemed weird to me to not do it. I think users will expect those three commands to be "the same", so having different lexical conventions for each one seemed odd. 

The only one I ended up allowing to be either upper- or lowercase was metavariables.

I'm happy to discuss any of these further.

jonsterling:
@wilcoxjay great job!

By the way, the reason for making `Thm` and `Tac` names be upper case is that each of these induces a new operator as well (just of a different sort)—and both theorems and tactics can take arguments. In fact, these elaborate to the same sort of thing in the signature that `Def` does—just that the elaboration of theorems is effectful (it runs the refiner).

I'll do a quick look to see that the choices of `VARNAME`, `OPNAME` and `IDENT` are in order. Thanks again!!

src/redprl/redprl.grm
```diff
@@ -256,21 +269,21 @@ customOpParams
...
 
 metavar
-  : HASH IDENT (IDENT)
+  : HASH ident (ident)
 
 dir
   : param SQUIGGLE_ARROW param ((param1, param2))
 
-term : rawTerm (annotate (Pos.pos (rawTerm1left fileName) (rawTerm1right fileName)) rawTerm)
+term : rawTerm %prec FUN_APP (annotate (Pos.pos (rawTerm1left fileName) (rawTerm1right fileName)) rawTerm)
```
jonsterling: this is interesting! What does this do?

wilcoxjay: There was a reduce/reduce conflict between this rule and the function application rule. This annotation resolved it, essentially by choosing to reduce as many function applications as possible before reducing it to a term.

PR #171, opened 2016-12-01
Handle new BadConversion exception from abt lib

wilcoxjay:
RedPRL/sml-typed-abts#78 adds a generic conversion error. This PR handles that error in RedPRL. The two PRs should be merged at the same time; neither is backwards compatible.

I also added a test demonstrating the new error message when arities don't match.

```
../../test/failure/bad-op.prl:3.22-3.31 [Error]:
  Arity mismatch for operator foo{}
```

### DIFF
--- a/src/redprl/signature.sml
+++ b/src/redprl/signature.sml
@@ -303,7 +303,7 @@ struct
       fun convertToAbt (metactx, symctx, env) ast sort =
         E.wrap (RedPrlAst.getAnnotation ast,
                 fn () => AstToAbt.convertOpen (metactx, metactxToNameEnv metactx) (env, NameEnv.empty) (ast, sort)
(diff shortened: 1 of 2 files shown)


wilcoxjay:
I've updated this branch to reflect the recent changes to the parser.

@jonsterling I'd appreciate a review on this (and RedPRL/sml-typed-abts#78) when you get a chance! Thanks!

wilcoxjay:
Just to be clear, the build is failing because travis builds against the master branch of RedPRL/sml-typed-abts.

jonsterling:
looks good! I've merged your PR to the abt library. I think you will need to update the submodule to point at the latest commit. (And in the future, if you want Travis to test your library changes, you can just temporarily change the submodule to point at them, and then update to master once the dependency PR is merged.)

Thanks for doing this!!

wilcoxjay:
Oh, I see, that makes sense. Thanks! 

I just updated the submodule, so it should build now.

PR #173, opened 2016-12-03
Build under sml-nj

### LINKED ISSUE #172: Support building executable with SML-NJ
When hacking on RedPRL, it's nice to play with your changes using one of the interactive editor modes, but these require a built executable rather than just an SML-NJ REPL with the bindings loaded.

Purely out of impatience for compile times under MLton, it would be nice to add support for building an executable with SML-NJ.

As a first cut, I think we can basically port the infrastructure from JonPRL.

wilcoxjay:
Fixes #172.

WIP: this works by commenting out the binding mentioned in #172, which is not a long term solution. In particular, the exit code should be set, and the command line arguments passed in. And this should be done in a way that is compatible with SML-NJ and MLton.

### DIFF
--- a/script/go-nj.sml
+++ b/script/go-nj.sml
@@ -1,0 +1,3 @@
+CM.make "src/frontend.cm";
+SMLofNJ.exportFn ("bin/.heapimg", Main.main);
+
--- a/script/mkexec.sh
+++ b/script/mkexec.sh
@@ -1,0 +1,13 @@
+#!/bin/sh
+#
+# Batch File Creator
+#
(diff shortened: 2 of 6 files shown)


jonsterling:
I think the right approach is to just have separate `main` files, one being for MLton and the other being for SML/NJ. Hopefully they can share most of their code though...

wilcoxjay:
Ah, good idea. I moved the top-level binding for MLton into its own file, and then added it `frontend.mlb` (but not `frontend.cm`).

PR #175, opened 2016-12-06
Add proof rules for hcom

wilcoxjay:
This is a work in progress on adding hcom rules to RedPRL's proof theory. These are summarized on pg 46 of "Dependent cubical realizability".

- [x] Add hcoms to the syntax view
- [x] Structural equality rule for hcoms
- [x] "Cap" rule
- [x] "Tube" rule
- [x] Figure out a better way of getting HCom.Eq rule to be called by auto
- [x] Add test for Cap rule
- [x] Add test for Tube rule

### DIFF
--- a/src/redprl/refiner.fun
+++ b/src/redprl/refiner.fun
@@ -1073,7 +1073,7 @@ struct
     (* Produce the list of goals requiring that tube aspects agree with each other. *)
     fun intraTubeGoals H w ty group =
       let
-        fun intraTube ext0 eps0 (u, tube0) ext1 eps1 (v, tube1) =
+        fun intraTube (ext0, eps0, (u, tube0)) (ext1, eps1, (v, tube1)) =
           let
             val tube0 = substSymbol (P.ret w, u) tube0
             val tube1 = substSymbol (P.ret w, v) tube1
@@ -1084,24 +1084,22 @@ struct
           end
       in
         listToTel
-          (ListMonad.bind (fn (ext0, eps0, tube0) =>
-             ListMonad.bind (fn (ext1, eps1, tube1) =>
-                 intraTube ext0 eps0 tube0 ext1 eps1 tube1)
-               group)
+          (ListMonad.bind
+             (fn x => ListMonad.bind (intraTube x) group)
              group)
       end
 
     (* Produce the list of goals requiring that tube aspects agree with the cap. *)
     fun tubeCapGoals H ty r cap group =
       let
-        fun tubeCap ext eps (u, tube) =
+        fun tubeCap (ext, eps, (u, tube)) =
           let
             val J = H >> CJ.EQ ((substSymbol (r,u) tube, cap), ty)
           in
-            List.map (fn j => #1 (makeGoal j)) (Restriction.One J ext eps)
+            List.map (#1 o makeGoal) (Restriction.One J ext eps)
           end
       in
-        listToTel (ListMonad.bind (fn (ext, eps, tube) => tubeCap ext eps tube) group)
+        listToTel (ListMonad.bind tubeCap group)
       end
 
     fun Eq alpha jdg =
@@ -1133,11 +1131,11 @@ struct
         val group0 = groupTubes exts0 tubes0
         val group1 = groupTubes exts1 tubes1
 
-        val interTubeGoals = listToTel
-          (ListMonad.bind (fn ((ext, eps, tube0), (_, _, tube1)) => interTube ext eps tube0 tube1)
-                          (ListPair.zipEq (group0, group1)))
-
-
+        val interTubeGoals =
+          listToTel
+            (ListMonad.bind
+               (fn ((ext, eps, tube0), (_, _, tube1)) => interTube ext eps tube0 tube1)
+               (ListPair.zipEq (group0, group1)))
       in
         T.append (T.empty >: goalTy >: goalCap)
                  (T.append interTubeGoals (T.append (intraTubeGoals H w ty group0)
--- a/src/redprl/refiner.fun
+++ b/src/redprl/refiner.fun
@@ -1015,33 +1015,43 @@ struct
 
   structure Restriction =
   struct
-    fun neq_dims (P.APP P.DIM0, P.APP DIM1) = true
-      | neq_dims (P.APP P.DIM1, P.APP DIM0) = true
-      | neq_dims _ = false
+    (* This structure provides functions that automate the restriction
+       judgement rules given in "Dependent Cubical Realizability",
+       page 46.
 
-    fun One jdg ext eps =
-      if P.eq Sym.eq (ext, eps) then [jdg]
-      else if neq_dims (ext, eps) then []
+       Such automation is possible because the rules are "syntax
+       directed" on the restriction being performed.
+     *)
+
+    fun neqDims (P.APP P.DIM0, P.APP DIM1) = true
+      | neqDims (P.APP P.DIM1, P.APP DIM0) = true
+      | neqDims _ = false
+
+    (* Restrict a judgement by a single equation `ext = eps`. *)
+    fun One (jdg : abt jdg) (ext : param) (eps : param) =
+      if P.eq Sym.eq (ext, eps) then [jdg] (* combining rules in first row *)
+      else if neqDims (ext, eps) then [] (* second row, left rule *)
       else
         let
           val (P.VAR v, P.APP eps) = (ext, eps)
         in
-          [Seq.map (substSymbol (P.APP eps, v)) jdg]
+          [Seq.map (substSymbol (P.APP eps, v)) jdg] (* bottom left rule *)
         end
 
-    fun Two jdg ext0 eps0 ext1 eps1 =
-      if P.eq Sym.eq (ext0, eps0) then One jdg ext1 eps1
-      else if P.eq Sym.eq (ext1, eps1) then One jdg ext0 eps0
-      else if neq_dims (ext0, eps0) then []
-      else if neq_dims (ext1, eps1) then []
+    (* Restrict a judgement by two equations `ext0 = eps0` and `ext1 = eps1`. *)
+    fun Two (jdg : abt jdg) (ext0 : param) (eps0 : param) (ext1 : param) (eps1 : param) =
+      if P.eq Sym.eq (ext0, eps0) then One jdg ext1 eps1 (* top right rule *)
+      else if P.eq Sym.eq (ext1, eps1) then One jdg ext0 eps0 (* top right rule *)
+      else if neqDims (ext0, eps0) then [] (* second row, left rule *)
+      else if neqDims (ext1, eps1) then [] (* second row, left rule *)
       else
         let
           val (P.VAR u, P.APP _) = (ext0, eps0)
           val (P.VAR v, P.APP _) = (ext1, eps1)
         in
-          if Sym.eq (u, v) andalso not (P.eq Sym.eq (eps0, eps1)) then []
+          if Sym.eq (u, v) andalso not (P.eq Sym.eq (eps0, eps1)) then [] (* second row, right rule *)
           else
-            [Seq.map (substSymbol (eps0, u)) (Seq.map (substSymbol (eps1, v)) jdg)]
+            [Seq.map (substSymbol (eps0, u)) (Seq.map (substSymbol (eps1, v)) jdg)] (* bottom right rule *)
         end
   end
 
--- a/src/redprl/refiner.fun
+++ b/src/redprl/refiner.fun
@@ -1136,6 +1136,7 @@ struct
 
   structure Computation =
   struct
+    exception UnsafeStep
 
     local
       open Machine.S.Cl infix <: $
@@ -1149,7 +1150,7 @@ struct
           case out m of
              th $ _ =>
                if List.exists (fn (_, sigma) => sigma = P.DIM) @@ Abt.O.support th then
-                 NONE
+                 raise UnsafeStep
                else
                  Machine.next sign st
            | _ => NONE
--- a/src/redprl/refiner.fun
+++ b/src/redprl/refiner.fun
@@ -1141,6 +1141,27 @@ struct
                                                     (tubeCapGoals H ty r0 cap0 group0)))
         #> trivial
       end
+
+    fun CapEq alpha jdg =
+      let
+        val _ = RedPrlLog.trace "HCom.CapEq"
+        val H >> CJ.EQ ((lhs, rhs), ty) = jdg
+        val Syn.HCOM (exts, (r, r'), ty0, cap, tubes) = Syn.out lhs
+        val () = assertParamEq "HCom.CapEq source and target of direction" (r, r')
+        val () = assertAlphaEq (ty0, ty)
+        val () = assertAlphaEq (cap, rhs)
+        val group = groupTubes exts tubes
+
+        val (goalTy, _) = makeGoal @@ H >> CJ.TYPE ty
+        val (goalCap, _) = makeGoal @@ H >> CJ.MEM (cap, ty)
+
+        val w = alpha 0
+      in
+        T.append (T.empty >: goalTy >: goalCap)
+                 (T.append (intraTubeGoals H w ty group)
+                           (tubeCapGoals H ty r cap group))
+        #> trivial
+      end
   end
 
 
--- a/src/redprl/refiner.fun
+++ b/src/redprl/refiner.fun
@@ -1166,7 +1166,6 @@ struct
         val _ = RedPrlLog.trace "Computation.EqHeadExpansion"
         val H >> CJ.EQ ((m, n), ty) = jdg
         val Abt.$ (theta, _) = Abt.out m
-        val hasFreeDims = List.exists (fn (_, sigma) => sigma = P.DIM) @@ Abt.O.support theta
         val m' = Machine.unload sign (safeEval sign (Machine.load m))
         val (goal, _) = makeGoal @@ H >> CJ.EQ ((m', n), ty)
       in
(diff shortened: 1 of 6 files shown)


wilcoxjay:
Okay, I've pushed a prototype of the structural equality rule as HCom.Eq.

One issue I ran into is that in the implementation of `auto`, `StepEq` is convinced that all my `hcom`s are redexes, but no step ever makes progress. This may be related to the way the syntax view "loses information" as discussed on IRC. 

I hacked around this for testing purposes by just calling the `HCom.Eq` rule on all redexes :) But we should fix this before merging.

@jonsterling Could you give me some early feedback on this? And also do you have any thoughts about the redex problem mentioned above?

wilcoxjay:
Oh, and don't forget to checkout the new `test/success/bool-hcom.prl`, which contains the first nontrivial cubical theorem proved in RedPRL! 🛠

jonsterling:
@wilcoxjay Regarding the auto-step rule, I can make a few hopefully helpful comments:

1. This may have something to do with the information loss we discussed, but there's a more fundamental issue.

2. The extent to which "direct computation rules" are possible for Cubical Type Theory is different from in Nuprl-style type theory. As a result, we have a very (*very*) conservative approximation which says *Don't do any direct computation if the term has any free dimensions in it*. This is captured in `safeEval` or whatever it is called.

3. When we are checking whether a term is a redex, we should ideally also be checking that condition there—since we should not attempt to reduce terms that have free dimensions.

Does this help?

test/success/bool-hcom.prl
```diff
@@ -0,0 +1,12 @@
+Thm BasicHcom : [
+  (a : bool) -> (b : bool) -> (c : bool) -> (d : bool) ->
+  paths({_}.bool; a; b) ->
+  paths({_}.bool; a; c) ->
+  paths({_}.bool; b; d) ->
+  paths({_}.bool; c; d) true
+] by [
+  { lam a. lam b. lam c. lam d.
+    lam pab. lam pac. lam pbd.
+    <i> 'hcom{i; 0 ~> 1}(bool; ,pab @ i; {j}. ,pac @ j; {j}. ,pbd @ j)
```
jonsterling: wow! the first hcom!

jonsterling:
More generally, though, these hcoms really will always look like redexes to the machine, because `hcom{NONE,...}(bool;...)` reduces to `hcom{BOOL,...}(...)`.

src/redprl/syntax.sml
```diff
@@ -15,6 +17,7 @@ struct
    | DFUN of 'a * variable * 'a | LAM of variable * 'a | AP of 'a * 'a
    | DPROD of 'a * variable * 'a | PAIR of 'a * 'a | FST of 'a | SND of 'a
    | ID_TY of (symbol * 'a) * 'a * 'a | ID_ABS of symbol * 'a | ID_AP of 'a * param
+   | HCOM of param list (* extents *) * dir * 'a (* type *) * 'a (* cap *) * (symbol * 'a) list (* tubes *)
```
jonsterling: one way to make this conversion lossless (I think) would be to include a boolean parameter here that says whether the hcom is canonical—which would only be meaningful assuming that the type parameter is a canonical type. Or maybe there is a cleaner way...

wilcoxjay:
Okay, I added some documentation about restriction and fixed style issues. 

@jonsterling I'm still not sure how to proceed getting `auto` to call the hcom rule. As you mentioned, the ABT machinery will always see source-level hcom's as redexes, but `safeEval` will refuse to do anything with them. Say we extend `auto`'s redex check to look for free dimensions. I'm still not sure which lower level tactic should be extended with the hcom rule.

By the way, would it make sense to have `safeEval` raise an exception instead of returning its argument in the case where a free dimension is present? This would at least prevent one kind of infinite loop, caused by disagreement about whether the term can step between the machine and `safeEval`.

jonsterling:
I will try to answer this as soon as I can (exams on Monday, so it may take a few days). thanks again for your hard work!

jonsterling:
1. Somehow, I think that it should be fine to make `safeEval` raise an exception rather than returning its argument; it seems like ideally we would not be collapsing failure in the rules themselves like we are doing now. Feel free to try it—I can't guarantee that depends on the current behavior, but hopefully nothing does.

2. As for adding a call to the hcom structural equality rule, I think that hcoms should be dealt with specially at the level of the automation tactics. Therefore, maybe the right thing to do is to make `StepEq` have an outer case statement that checks for hcoms, and then delegates to the appropriate rule; if it is not an hcom, then, pass to the existing canonicity-based dispatch.

cangiuli:
I haven't been watching closely, but what exactly is the intent of `safeEval` right now? I think we decided that one can evaluate safely so long as the principal argument's head constructor has no dimension names as indices, no?

jonsterling:
@cangiuli That would be quite nice, but the last time we spoke about it, you said that we could only evaluate if the entire term had no free dimension names. Previously, it was implemented such that only the head operator's indices were checked, but I seem to remember that we discussed it and you thought that would be unsound. Perhaps we had a miscommunication, though!

wilcoxjay:
You two are the experts, but to my eye it looks like `safeEval` is actually only checking the head operator's indices. See [here](https://github.com/RedPRL/sml-redprl/blob/8cf6e816720cd7a8fe147a6f412d0300be015a76/src/redprl/refiner.fun#L1032)

jonsterling:
@wilcoxjay Huh, I must have got what happened backwards or messed up 😄  @cangiuli

jonsterling:
Ah, I see what `safeEval` is doing now; I should have just looked at it before I commented from memory 15 minutes ago, haha.

It checks the head operator at every *step*; this should be sound, but it may not be quite ideal. As @cangiuli mentioned, it would be better to be checking the indices of the principal argument; however, I'd have to think a bit about how to achieve that generically (it ought to be possible with the machine apparatus that's in place, just have to think about it). For now, we should probably just leave it the way it is until we propose the proper way to do it.

cangiuli:
@jonsterling It doesn't matter whether or not the term has dimensions internally; the indices of the principal argument are, I think, the only things that pose a problem.

Let me write `R` for the "cubically-stable stepping judgment", which I think is the intent of `safeEval`. For example, `R((\a.M) N , M[N/a])` regardless of free dimensions, since dimensions in `M`,`N` can't possibly affect this step. Also note that `R(M@x , M'@x)` when `R(M,M')`, even though `x` is the index of the head operator, because the identity of that index does not affect the computation until `M` is a value. Does this sound right?

jonsterling:
@cangiuli This is very very helpful!

But it sounds like it contradicts some of the things we'd discussed in the past, which is the idea of being (not) able to step an `hcom` which is non-canonical if it has a dimension variable in its indices. Can I read your comment as a confirmation that it is actually safe to step `hcom{x...}(A -> B; ...)` to `\a. hcom{x...}` or whatever?

jonsterling:
@cangiuli Aha, I just re-read your comment and now I understand what you were saying. So, that would be a resounding "no" to my last question 😄 

I think the relation you describe makes a lot of sense, and we should try to implement it.

cangiuli:
@jonsterling Right, when `hcom{x...}` is the principal argument, this is precisely the scenario in which stepping is not safe. But I think it would be prudent for us to discuss this more carefully before implementing something fancy.

jonsterling:
@cangiuli OK, sounds good! Soon (once I get onto vacation), I could come up with a proposal for such a relation, with respect to our machine dynamics, which you could review and we could discuss.

src/redprl/refiner.fun
```diff
@@ -1015,8 +1013,190 @@ struct
...
   end
 
+  structure Restriction =
+  struct
+    (* This structure provides functions that automate the restriction
+       judgement rules given in "Dependent Cubical Realizability",
+       page 46.
+
+       Such automation is possible because the rules are "syntax
+       directed" on the restriction being performed.
```
jonsterling: great! This is just the sort of thing that Carlo and I had envisioned, and thanks for putting the reference here.

src/redprl/refiner.fun
```diff
@@ -1015,8 +1013,190 @@ struct
...
+     *)
+
+    fun neqDims (P.APP P.DIM0, P.APP DIM1) = true
+      | neqDims (P.APP P.DIM1, P.APP DIM0) = true
+      | neqDims _ = false
```
jonsterling: What is the meaning of `neqDims`? I'm curious why it always returns false on "generic" / variable dimensions.

src/redprl/refiner.fun
```diff
@@ -1015,8 +1013,190 @@ struct
...
+      end
+
+    (* Produce the list of goals requiring that tube aspects agree with the cap. *)
+    fun tubeCapGoals H ty r cap group =
+      let
+        fun tubeCap ext eps (u, tube) =
+          let
+            val J = H >> CJ.EQ ((substSymbol (r,u) tube, cap), ty)
+          in
+            List.map (fn j => #1 (makeGoal j)) (Restriction.One J ext eps)
```
jonsterling: might as well write `#1 o makeGoal`

wilcoxjay:
Okay, I've finished prototyping all the rules. 

I'd appreciate another pair of eyes looking at the rules on paper and making sure the code matches. 

I'd also appreciate any ideas for examples/tests that use the cap and tube rules. 

One thing I'm not sure about for the cap and tube rules is how useful they are when stated the way they are, since they require the right hand side to already be the right thing. The proof theory doesn't (yet?) have any sort of substitution or rewrite rule, so I imagine it's rather difficult to get your goal into a form where these rules apply.

src/redprl/refiner.fun
```diff
@@ -1015,8 +1013,190 @@ struct
...
+          end
+      in
+        listToTel (ListMonad.bind (fn (ext, eps, tube) => tubeCap ext eps tube) group)
```
jonsterling: If you change `tubeCap` to be curried, this could be written `listToTel (ListMonad.find tubeCap group)`, which might be a little cleaner.

src/redprl/refiner.fun
```diff
@@ -1015,8 +1013,190 @@ struct
...
+            List.map (fn j => #1 (makeGoal (([(w, P.DIM)], []) |> j)))
+                     (Restriction.One J ext eps)
+          end
+
+        val group0 = groupTubes exts0 tubes0
+        val group1 = groupTubes exts1 tubes1
+
+        val interTubeGoals = listToTel
+          (ListMonad.bind (fn ((ext, eps, tube0), (_, _, tube1)) => interTube ext eps tube0 tube1)
+                          (ListPair.zipEq (group0, group1)))
```
jonsterling: I don't really care that much (so feel free to ignore this comment), but for reference, I prefer the following indentation style (which avoids leading whitespace that depends on lexical width of other identifiers):

```sml
val interTubeGoals = 
  listToTel
    (ListMonad.bind 
      (fn ((ext, eps, tube0), (_, _, tube1)) => interTube ext eps tube0 tube1)
      (ListPair.zipEq (group0, group1)))
```

jonsterling:
@wilcoxjay Good point about the "judgmental computation rules" for cap & tube... What I've done in the past is taken the rules on paper and "transposed" them, so that one side of the equation doesn't have any constraints on it. Essentially, this is like building transitivity into each rule, which is a bit better for refinement proof for the reasons you describe. An example of this in practice is `Path.ApComputeConst`.

src/redprl/refiner.fun
wilcoxjay: `neqDims` is used to implement the part of the restriction judgment that "finds a contradiction" such as `0 = 1`. Since `x = 0` is not a contradiction, it is important that both sides are constants, and that they are different. I'll clarify this.

src/redprl/syntax.sml
wilcoxjay: @jonsterling Coming back to this, do you think making the conversion losless is important to do before merging this PR?

src/redprl/refiner.fun
jonsterling: Ah, ok! Thanks!

Maybe it could be renamed to something like `dimsContradictory` or something else?

src/redprl/syntax.sml
jonsterling: I don't think it's crucial for this PR; already, the syntax view is fairly fast-and-loose. We might make cleaning it up a separate task, rather than burdening you with it right now.

wilcoxjay:
@jonsterling Hmm, okay, I'll try to implement transposed versions of those rules. But I don't really understand what's going on in `Path.ApComputeConst`. What does the `Synth` judgment mean? Something like "give me a type for this term"?

If we translate the code back to an on paper rule, would it be something like (horrible mix of RedPRL and TeX syntax...)
```
    m in paths(x.A; P_0; P_1)
    A<\epsilon / x> = A' type
    P_\epsilon = P in A'
------------------------------
  m @ \epsilon = P in A'
```

jonsterling:
@wilcoxjay Yeah, the implementation there is a bit arcane 😄 `R synth` is indeed a judgment that means `tell me what type R is`... This is a nice trick that is enabled by dependent refinement, since when we add this subgoal, we bind a variable which ranges over whatever type `R` turns out to be; then, we can drill down into this type using the `MATCH` judgment. This gives us a way to sort of "asynchronously" depend on stuff from an question that we may not yet have fully answered.

I think your translation of the rule looks about right.

The main idea of this "transposition" is that you take an equational rule `M = N : A` and replace it with something like `M = L : A ====> N = L : A`. Then, the rule can be applied without the user having to get the goal into just the right shape.

wilcoxjay:
Okay, I think I got it; that was easier than I expected :) 

@jonsterling does that look right?

src/redprl/refiner.fun
```diff
@@ -1015,8 +1013,190 @@ struct
...
+        val _ = RedPrlLog.trace "HCom.CapEq"
+        val H >> CJ.EQ ((lhs, rhs), ty) = jdg
+        val Syn.HCOM (exts, (r, r'), ty0, cap, tubes) = Syn.out lhs
+        val () = assertParamEq "HCom.CapEq source and target of direction" (r, r')
+        val () = assertAlphaEq (ty0, ty)
+        val group = groupTubes exts tubes
+
+        val (goalTy, _) = makeGoal @@ H >> CJ.TYPE ty
+        val (goalCap, _) = makeGoal @@ H >> CJ.MEM (cap, ty)
+        val (goalEq, _) = makeGoal @@ H >> CJ.EQ ((cap, rhs), ty)
```
wilcoxjay: So I just added this goal when mechanically transposing the rule, but it looks to me like it makes the previous goal redundant. 

Isn't it the case that if `a = b in A` then `a in A`?

jonsterling:
Great work @wilcoxjay! I'll take some time today to review the rules and compare them with what's on paper.

src/redprl/refiner.fun
```diff
@@ -1015,8 +1015,224 @@ struct
...
+      let
+        val _ = RedPrlLog.trace "HCom.CapEq"
+        val H >> CJ.EQ ((lhs, rhs), ty) = jdg
+        val Syn.HCOM (exts, (r, r'), ty0, cap, tubes) = Syn.out lhs
+        val () = assertParamEq "HCom.CapEq source and target of direction" (r, r')
+        val () = assertAlphaEq (ty0, ty)
+        val group = groupTubes exts tubes
+
+        val (goalTy, _) = makeGoal @@ H >> CJ.TYPE ty
+        val (goalCap, _) = makeGoal @@ H >> CJ.MEM (cap, ty)
```
jonsterling: As you described, let's delete this goal since it is redundant.

src/redprl/refiner.fun
```diff
@@ -1015,8 +1015,224 @@ struct
...
+        fun go acc [] = raise List.Empty
+          | go acc (x :: l) =
+            if p x then acc
+            else go (acc + 1) l
+      in
+        go 0
+      end
+
+    (* Search for an index of a constant extent in an hcom, to pass to HCom.TubeEq. *)
+    fun FindTubeEq alpha jdg =
```
jonsterling: Good idea, factoring this out and having the primitive rule take `i` as a parameter.

src/redprl/refiner.fun
```diff
@@ -1015,8 +1015,224 @@ struct
...
+        val () = assertParamEq "HCom.Eq source of direction" (r0, r1)
+        val () = assertParamEq "HCom.Eq target of direction" (r'0, r'1)
+        val _ = ListPair.mapEq (assertParamEq "HCom.Eq extents") (exts0, exts1)
+
+        val (goalTy, _) = makeGoal @@ H >> CJ.EQ_TYPE (ty0, ty1)
+        val (goalCap, _) = makeGoal @@ H >> CJ.EQ ((cap0, cap1), ty)
+
+        val w = alpha 0
+
+        fun interTube ext eps (u, tube0) (v, tube1) =
```
jonsterling: Can you add a comment here which describes what this does, analogous to the comment on `intraTubeGoals`? In these comments, it might also be nice to write (in ascii) an example premise that they correspond to from the paper.

src/redprl/refiner.fun
```diff
@@ -1015,8 +1015,224 @@ struct
...
+       tube's position in the list *)
+    fun groupTubes [] [] = []
+      | groupTubes (ext :: exts) (tube0 :: tube1 :: tubes)  =
+        (ext, P.APP P.DIM0, tube0) :: (ext, P.APP P.DIM1, tube1) :: groupTubes exts tubes
+      | groupTubes _ _ = raise Fail "groupTubes"
+
+    fun listToTel l = List.foldl (fn (g, l) => l >: g) T.empty l
+
+    (* Produce the list of goals requiring that tube aspects agree with each other. *)
+    fun intraTubeGoals H w ty group =
```
jonsterling: Am I right to understand that this corresponds to the premise that says `forall i, j, eps, eps'. N_i^eps = N_j^eps' in A [Psi, y | r_i = eps, r_j = eps']` (the fourth premise of the structural equality rule)?

If so, does this do the right thing? From the rule it looks like we need to try every possible combination of tube faces or something, but I can't quite make out how that's happening here. But I might be misunderstanding either the paper rule or the implementation.

wilcoxjay: Yes, that's right. We do need to consider all pairs of tube faces. The local function `intraTube` produces the (one or zero) goals corresponding to a particular choice of tube faces. Then we wrap that up in a couple `ListMonad.bind`s to get all of them. 

Does that seem right?

src/redprl/refiner.fun
wilcoxjay: Good idea, I just pushed some comments!

src/redprl/refiner.fun
jonsterling: Ahhhh, it's the `bind` that makes it all work 😄 Thanks for explaining!

PR #177, opened 2016-12-12
Handle new Duplicate error from telescopes

wilcoxjay:
This PR handles the new Duplicate exception from the telescope library. See RedPRL/sml-telescopes#10. 

I also took the opportunity to do a little local refactoring of the elaborator's error reporting mechanism.

### DIFF
--- a/src/redprl/signature.sml
+++ b/src/redprl/signature.sml
@@ -169,6 +169,8 @@ struct
 
     structure Err = RedPrlError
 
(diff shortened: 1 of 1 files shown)


wilcoxjay:
The build will fail until I can update the telescope dependency. (Can't use the usual trick since the PR is from a branch in a fork.)

jonsterling:
looks good! I've just merged the other PR, and fixed the access issue.

PR #178, opened 2016-12-13
Add sequents to frontend

### LINKED ISSUE #176: Support sequents in the frontend
All of RedPRL's internals manipulate sequents, but there is no way for a user to give a nontrivial sequent explicitly. It would be nice if there was such a way.

For example, this would let you prove "polymorphic theorems", which would traditionally be stated by quantifying over a universe, by instead stating, eg, `A type >> A -> A true`.

There are some things to be worked out:
* Represent sequents in the abstract syntax; there are some tricky issues with the binding structure of sequents. (@jonsterling could you expand on what you meant here?)
* Represent sequents in the concrete syntax. This should be fairly straightforward once we figure out the abstract counterpart. I'm happy to do the requisite parser hacking.
* Generalize the elaborator's implementation of the `Thm` command to tolerate nontrivial sequents, if necessary.
* Generalize the Lemma rule to nontrivial sequents, if necessary.
* Think carefully about the well formedness invariants required by the proof theory and ensure that these are checked on "untrusted" sequents.

wilcoxjay:
Fixes #176. 

There are some outstanding problems with this branch.

## Concrete Syntax
The concrete syntax for membership is now 'in', instead of ':'. This is
in order to fix an ambiguity in the grammar between binding a name to a
hypothesized judgement, or hypothesizing a (nameless) membership
judgement.

For consistency, we might want to also change the concrete syntax of the
equality judgment to be `a = b in A`, but I haven't done this yet.

## Converting ABTs to sequents

I don't understand functors very well, so I just banged my head against
it until it compiled. This should be fixed up.

There are at least two issues.

1) Sequents are functorized over the categorical judgments without
   constraints. But in order to convert abts to sequents, we obviously
   need to know that our abts have the operators we expect, and not some
   arbitrary operators. Thus we restrict Tm.O.t to be RedPrlOperator.t.
   I'm pretty confident that something like this is necessary.

2) Are hypothesis names symbols or variables? The current infrastructure
   treats them as symbols (eg, hyp{x}), but they feel like variables in
   the context of the binding structure of sequents (eg, they can appear
   as terms, whereas symbols cannot: b : bool true >> b = b : bool). In
   order to get things to compile, I introduced the completely bogus
   sharing constraint Sym.t = Var.t, but this should be revisited.

As an aside, could we just not functorize sequents over categorical
judgments? I'm not sure exactly what the benefits of the functorization
are, so I can't comment on this further.

## Elaborator

I am not convinced that I did the right thing in the elaborator. What is the purpose of the `REFINE` operator? How should it be generalized to handle nontrivial sequents?


jonsterling:
Thanks very much for getting this started! I will try to answer your questions as best I can.

### concrete syntax
You should feel free to make whatever changes to the concrete syntax that you feel are warranted. The nice thing about having no users is that we can do whatever we like 😄 

### conversions

1. The constraint you describes sounds appropriate

2. Excellent question about hypothesis names in a sequent; this is a very subtle issue, and negotiating it has led to a situation in which symbols and variables come from the same type in RedPRL. I will summarize what is going on in a section below.

3. I think I have tried to functorize sequents in that way before, but ran into some issues that I didn't want to deal with. If you want to try it out you are welcome to do so, but I seem to remember trying it before and being unsatisfied with the result (I might have just done it wrong though!).

### are hypotheses symbols or variables?

Hypotheses have two different aspects in RedPRL; first, there is their place within the proof system, which is essentially a form of sequent calculus. The main distinction between natural deduction and sequent calculus is that the primary motive of computation in natural deduction is *substitution*, whereas in sequent calculus, *cut* plays this role instead. As a result, in sequent calculus, hypotheses are not subject to substitution (rather, only renaming), and in many interpretations, they behave like symbols rather than like variables.

All this is to say, in the proof system and tactic language, hypotheses are treated as symbols. However, in the *programs*/extracts, hypotheses are turned into variables and are again subject to substitution. This is, by the way, the reason for some of the subtlety concerning the "unquoting" of hypotheses as variables in tactic scripts (e.g. `,x`).

I believe that sequents should bind variables rather than symbols, because the terms with these names bound are subject to substitutions.

### elaborator / `REFINE`

Good question about `REFINE`. Basically, `REFINE` turns a goal and a tactic script into a "theorem" (that is, an object of sort `THM`). A `REFINE` expression optionally has as a subterm an extract, i.e. a realizer for its theorem; this term exists depending on one of the indices to the operator. If the realizer is present, then the theorem has been refined and can be used; if it is not present, then the theorem has not been refined (and isn't really a theorem).

Using "poor man's data abstraction" (i.e. the parser), we ensure that it is not possible for a user to construct one of these objects except by using the refiner.

#### what needs to be done to the elaborator to support hypotheses in the initial goal?

An interesting question comes up here, since I'm not certain what is to be done about the names that are bound in a sequent—how can they be used in the proof script? In some sense, we are again coming up against a tension between symbol and variable. In some sense, this part would be easier if sequents had no binding structure, and instead the names of hypotheses were *free* symbols, bound as *parameters* to the theorem in question (under this version of sequents, sequents themselves would have no binding structure, but would be an association of symbols to judgments). Then, they'd be accessible in the proof script, etc. in the appropriate way.

It's not clear to me what the right answer is here. Maybe what ought to be done is that we have to ensure that the theorem has a sufficient number of parameters of sort `HYP`, and then we will use THESE specific symbols *in order* as a namestore when translating the syntactic representation of the sequent into the programmatic representation. To make this concrete, I'm imagining something like this being what a user writes:

```
Thm Jon[a:hyp,b:hyp] : (x : J1, y : J2 >> J3) by [
  // Then, the user sees a goal like "a : J1, y : J2 >> J3", 
  // and write things like "elim[a]" or "hyp[a]".
].
```

I'm not certain the above is the best idea, but I think it could work. Do you have any thoughts or feedback?

wilcoxjay:
Thanks for your detailed answers.

## Concrete syntax 

> You should feel free to make whatever changes to the concrete syntax that you feel are warranted. The nice thing about having no users is that we can do whatever we like 😄

Alright then, I'll wreak whatever havoc I desire 

## Conversions

I'll take a look at unfunctorizing sequents and report back. 

## Symbols vs variables and using hypotheses in initial goals

> Maybe what ought to be done is that we have to ensure that the theorem has a sufficient number of parameters of sort HYP, and then we will use THESE specific symbols in order as a namestore when translating the syntactic representation of the sequent into the programmatic representation

I think this is actually kind of reasonable. I want to suggest a kind of concrete syntactic hack to make it a bit more user friendly, and you can tell me what you think. 

Instead of explicitly writing out the parameters of the theorem in addition to the hypothesis names bound in the sequent, we just **reuse** the hypothesis names from the sequent as additional bound parameters to the theorem. This will require some slightly funky stuff in the parser, but I think I know how to do it. 

So concretely, your example would like 

```
Thm Jon : [x : J1, y : J2 >> J3] by [
  // The user sees a goal like "x : J1, y : J2 >> J3", 
  // and can use the hypothesis names as symbols in the proof script, 
  // like "elim[x]" or "hyp[x]".
].
```

What do you think? Too hacky? 

Another way to ask this question is: sure, the implementors of RedPRL are going to have to know that hypothesis names are both symbols and variables, but the (as-yet nonexistent) *users* have to know?

jonsterling:
@wilcoxjay thanks for your answers! I'm about to be traveling (quite an ordeal), so I'll have to be super brief for now, but let me say that I like your idea about the concrete syntax for the theorem declaration w/ sequents.

These fancy theorem declarations elaborate to "pure" definitions, so my suggestion would be to make it so that in the elaboration, the names that were chosen there get added as parameters of sort `HYP` during elaboration.

wilcoxjay:
Sounds good, I'll prototype it.

Have a good trip!

jonsterling:
> Have a good trip!

Thanks!! Hopefully I'll be in touch again soon, but the place I'm going has been without electricity for a few days and there's no sign it's going to be repaired shortly, so if you don't hear from me for a few days, it's because I'm tending the fire and reading Lenin, lol.

jonsterling:
@wilcoxjay  Hi! I just wanted to check back in about this issue. I think that the conclusion we came to was a good one, and that we should begin pursuing it soon. The ability to prove arbitrary sequents will be really important for RedPRL, and it's a step along the way toward enabling MetaPRL-style proofs of entire inference rules.

I imagine you're pretty busy these days, so I'd be happy to start putting in some of the preliminary stuff if you wanted...

One thing I could use your feedback on is, do you think that there is a way to do this work incrementally? It's not necessary that any part of the feature actually be usable in between PRs, but it might be nice to break the work into smaller pieces and do those one by one.

wilcoxjay:
@jonsterling Hi! I am indeed busy teaching undergrad PL at the moment. All my cycles are going into making interesting assignments for the kiddos. I suppose I could make RedPRL an assignment, but I don't think that would go over very well 😄 

I double checked and I don't have any unpushed commits on this branch, but I do have a small incomplete uncommitted change to `elabRefine`. I don't remember exactly what I was doing, but it looks like something to do with sequents, so here's the diff, just in case it's helpful. (Feel free to ignore.)
```
diff --git a/src/redprl/signature.sml b/src/redprl/signature.sml
index 4169de0..ca68d38 100644
--- a/src/redprl/signature.sml
+++ b/src/redprl/signature.sml
@@ -322,9 +322,8 @@ struct
 
       fun names i = Sym.named ("@" ^ Int.toString i)
 
-      fun elabRefine sign (goal, script) =
+      fun elabRefine sign (seqjdg, script) =
         let
-          val seqjdg = RedPrlSequent.fromAbt goal
           val (_, tau) = RedPrlJudgment.sort seqjdg
           val pos = getAnnotation script
         in
@@ -348,7 +347,16 @@ struct
           val (params', symctx, env) = elabDeclParams sign params
           val names = fn i => Sym.named ("@" ^ Int.toString i)
         in
-          convertToAbt (metactx, symctx, env) goal SEQ <&> convertToAbt (metactx, symctx, env) script TAC
+          convertToAbt (metactx, symctx, env) goal SEQ
+            >>= (fn seq =>
+              let
+                open RedPrlSequent
+                val hyps >> concl = RedPrlSequent.fromAbt seq
+              in
+              end
+
+
+ <&> convertToAbt (metactx, symctx, env) script TAC
             >>= elabRefine sign
             >>= (fn definiens => E.ret @@ EDEF {sourceOpid = opid, params = params', arguments = arguments', sort = sort definiens, definiens = definiens})
         end
```

In terms of incrementalizing this work, I'm not really sure if it's possible. On the other hand, I don't think the most basic functionality requires too huge of a change, so hopefully it wouldn't be that bad to do in one go. That said, I haven't thought about this stuff in 6 weeks, and won't be able to again for at least another 6 weeks or so, so take all that with a grain of salt.

jonsterling:
This was easier to finish up than I expected! Will merge in soon. Thanks @wilcoxjay for your hard work!

PR #181, opened 2016-12-31
update to new dependent lcf

jonsterling:
Begin switching over to the built-in support for generic judgment in dependent lcf. It's still broken, but this will help me fix it.


jonsterling:
It is a bit strange that I had to switch to the depth-first version of `auto` to make this work; it clearly has something to do with the way that the PROGRESS multitactical works, which is not yet completely how I want it. However, I don't want to bother changing it just yet until I understand more what is going on. In the meanwhile, I think I will merge this.

PR #183, opened 2017-01-04
Fix/issue 182

### LINKED ISSUE #182: Sequencing tactic parsed incorrectly?
Something strange is happening with the sequencing tactical `x <- t1; t2`, but I'm not yet sure what. Just putting this here so that I will remember to look into it.

(It doesn't affect normal RedPRL use, since users are meant to use the high-level development calculus and the `auto` tactic, rather than do low-level proofs manually.)

jonsterling:
fix #182

### DIFF
--- a/src/redprl/redprl.grm
+++ b/src/redprl/redprl.grm
@@ -374,7 +374,7 @@ atomicRawMultitac
   | MTAC_PROGRESS LBRACKET multitac RBRACKET (Ast.$$ (O.MONO O.MTAC_PROGRESS, [\ (([], []), multitac)]))
   | MTAC_REC VARNAME IN LBRACKET multitac RBRACKET (Ast.$$ (O.MONO O.MTAC_REC, [\ (([],[VARNAME]), multitac)]))
   | LBRACKET multitac RBRACKET (multitac)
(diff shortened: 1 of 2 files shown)


src/redprl/redprl.grm
```diff
@@ -374,7 +374,7 @@ atomicRawMultitac
   | MTAC_PROGRESS LBRACKET multitac RBRACKET (Ast.$$ (O.MONO O.MTAC_PROGRESS, [\ (([], []), multitac)]))
   | MTAC_REC VARNAME IN LBRACKET multitac RBRACKET (Ast.$$ (O.MONO O.MTAC_REC, [\ (([],[VARNAME]), multitac)]))
   | LBRACKET multitac RBRACKET (multitac)
-  | hypBindings LEFT_ARROW atomicRawMultitac SEMI multitac %prec LEFT_ARROW (Tac.makeSeq multitac hypBindings atomicRawMultitac)
+  | hypBindings LEFT_ARROW atomicRawMultitac SEMI multitac %prec LEFT_ARROW (Tac.makeSeq atomicRawMultitac hypBindings multitac)
```
wilcoxjay: oops :)

jonsterling: haha, no worries 😆 Glad it was something simple!!

Issue #37, opened 2016-02-06
MetaPRL-style sequent schemata and derived rules


jonsterling:
This would make our lives so much easier when dealing with things like wellformedness lemmas; it also seems necessary if we want to allow data abstraction.

jonsterling:
Now that #176 has been dealt with, we are one step closer.

jonsterling:
It's done! Exciting. Though, we've hit the limit until we are able to switch to LF for the syntactic basis.

PR #185, opened 2017-03-28
rough start on implementing rule schemes

jonsterling:
#37 [very rough, don't look at this code (lol!)]

The basic idea is that a rule is just like a theorem, except you can close it without completing all the subgoals. Then, when you use a rule as a lemma, it "restores" the appropriate missing subgoals into the current state. This is a very simple idea, whose technical realization is a bit more involved.

When I add a declaration form for rules (right now I'm just piggy-backing on theorems), this could include a specification of the 'final' subgoal telescope by the user, so that the definitions succeeds only if this specification unifies with the actual state of affairs.

One sucky fact is that we really need a *second-order binder* in the syntax of theorem objects / extractions, to account for the fact that a piece of evidence now can have holes in it. Currently we just fake the binding, which is sad. If we can move to a higher-order logical framework, this would be resolved.

### DIFF
--- a/src/redprl.cm
+++ b/src/redprl.cm
@@ -6,6 +6,7 @@ Library
   structure RedPrlLog
   structure RedPrlError
   structure RedPrlLrVals
+  structure RedPrlAbt
 is
   $/basis.cm
   $/ml-yacc-lib.cm
--- a/src/redprl.cm
+++ b/src/redprl.cm
@@ -25,6 +25,7 @@ is
 
   debug.cm
 
+  redprl/symbol.sml
   redprl/parameter.sml
   redprl/operator.sml
   redprl/abt.sml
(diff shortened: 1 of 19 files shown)


jonsterling:
I need to add syntax for the following:

1. generic judgments
2. rule schemes (a list of generic judgments, and then a sequent)

Then, it should be 'straightforward' (but laborious) to write the code that unifies a rule scheme / specification with the proof state, to check that a rule declaration is valid. At that point, we should have the full power of MetaPRL's sequent schemata.

jonsterling:
I've made some progress, but I need to think more carefully about how hypothesis names should be dealt with. The current setup leads to an incorrect result.

jonsterling:
Superseded by #186

PR #186, opened 2017-04-12
refactor elaborator around derived rules

jonsterling:
this will lead the way toward derived rules (#37)


jonsterling:
I had a thought, which is that we could go even further than this!

One way to unify everything conceptually is to say that *all* construction occurs through the refiner, even definitions (which are currently treated differently from theorems). Currently, definitions are "fundamental" and theorems elaborate to special definitions which add to our state of knowledge.

But I think it may make more sense to have theorems be the fundamental notion, and have definitions be treated like theorems (their statements would be something silly like "there exists an expression"). All this is of course internal, so it makes no difference to the user.

But it would allow us to simplify a lot of stuff.

jonsterling:
BTW, this also gives us a way to construct raw terms incrementally/interactively (no different from how the fancy development language works).

jonsterling:
TODO:

- [ ] add sequent 'unification': when we apply a lemma or derived rule, we need to merge the sequents together in the appropriate way so that the premises have access to the right hypotheses. This is justified semantically by the "uniformity/naturality" property of type theoretic constructions.

jonsterling:
It would be sweet to be able to infer the parameters to a rule application using unification!

src/redprl/signature_mini.sml
```diff
@@ -0,0 +1,165 @@
...
+      for hypotheses qua variables. *)
+    fun hypothesisRenaming (entry : entry) (ps : Tm.param list) : Tm.varenv = 
+      let
+        fun handleHyp ((u, psort), ptm, ctx) = 
+          case psort of 
+            RedPrlParamData.HYP =>
+              let
+                val RedPrlParameterTerm.VAR v = ptm
+              in
+                Var.Ctx.insert ctx u (Tm.check (Tm.`v, RedPrlOpData.EXP))
```
jonsterling: This is not quite right; we need to somehow find out what judgment the hypothesis is for, and use its synthesis instead of `EXP`.

PR #195, opened 2017-06-12
Clean-up and bugfixes, stage 0

favonia:
I was trying to clean up some files and suddenly fixed two bugs, therefore I think it is a good thing. I further rewrote `eqPoly` to prevent this kind of mistakes (which can now be detected by a disciplined programming style or a coverage checker).

* Add more comments to the constructors.
* (Try to) unify the order of constructors.
* Fix the bug that `eqPoly` missed `PATH_AP`.
* Rewrite `eqPoly` to exploit coverage checking in order to prevent this kind of mistakes.
* Fix the bug that `toString` for `sort` missed `SEQ`.

### DIFF
--- a/src/redprl/operator.sml
+++ b/src/redprl/operator.sml
@@ -83,6 +83,7 @@ struct
      | MTAC => "mtac"
      | JDG => "jdg"
      | TRIV => "triv"
+     | SEQ => "seq"
 end
 
 structure RedPrlArity = ListAbtArity (structure PS = RedPrlParamSort and S = RedPrlSort)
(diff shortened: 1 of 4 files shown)


jonsterling:
@favonia this looks great! Nice work.

src/redprl/operator.sml
```diff
@@ -347,34 +367,46 @@ struct
...
+       | (RULE_UNFOLD a, t) =>
+           (case t of RULE_UNFOLD b => f (a, b) | _ => false)
+       | (DEV_BOOL_ELIM a, t) =>
+           (case t of DEV_BOOL_ELIM b => f (a, b) | _ => false)
+       | (DEV_S1_ELIM a, t) =>
+           (case t of DEV_S1_ELIM b => f (a, b) | _ => false)
+       | (DEV_DFUN_ELIM a, t) =>
+           (case t of DEV_DFUN_ELIM b => f (a, b) | _ => false)
+       | (DEV_DPROD_ELIM a, t) =>
+           (case t of DEV_DPROD_ELIM b => f (a, b) | _ => false)
```
jonsterling: I'm OK with this change, but I wouldn't make other ones like it in the future. I understand why you prefer it, and I think that your reasoning makes a lot of sense, but it is not idiomatic for ML.

favonia: Well, I am actually up for rewriting it back once the RedPRL core calculus is more or less stable (and that we have enough tests). Now, given that we already made a mistake, I prefer to stay with ugliness while we are still tempering with the heart of RedPRL.

jonsterling:
that sounds reasonable! you have my endorsement. 

Von meinem iPad gesendet

favonia:
@jonsterling By the way, I renamed `DEV_FUN_INTRO` to `DEV_DFUN_INTRO` but forgot to mention it in the commit message, and just realized it could be a mistake after reading more code. Should I revert it back?

favonia:
Talked to @jonsterling and the renaming to `DEV_DFUN_INTRO` seems fine.

jonsterling:
@favonia no need to revert, you're doing fine :)

Issue #193, opened 2017-06-11
Rename ID types to PATH types.


favonia:
Due to various concerns I think paths are a better name for the current ID types in RedPRL. I can do the renaming if there is no objection.

jonsterling:
Good idea, agreed.

favonia:
@jonsterling It seems there is not enough testing for ID types, because my changes did not break anything! Well, I kind of expected this because it is not mentioned in the lexer. I am not sure if I can close this thread if it was not properly tested.

jonsterling:
We don't worry too much about thorough testing in RedPRL, given our lack of engineering power. It's too hard to test a proof assistant in a thorough way, especially one that is constantly changing. Instead, we just make sure some basic examples continue to work.

PR #196, opened 2017-06-12
Fix a bug introduced by 9b803ee. See #193 and #194.

favonia:
I somehow missed the `pretty.sml` file.


jonsterling:
good catch!

This is one of those areas where I wish SML had a special lexical syntax for constructors (as in Haskell), so that they could not be confused with pattern variables.

PR #199, opened 2017-06-12
Clean-ups and bugfixes, stage 1

favonia:
This pull request will be a collection of further clean-ups. The stage ends either when a critical bug is found or when I feel the collection is large enough. A review will be requested when it is time.

### DIFF
--- a/src/redprl/signature_mini.sml
+++ b/src/redprl/signature_mini.sml
@@ -158,7 +158,7 @@ struct
       fun extract (Lcf.|> (subgoals, validation)) = 
         case outb validation of 
            ([],[]) \ term => term
-         | _ => raise Fail "Extract term has unexpected binder. Can this relaly happen?"
+         | _ => raise Fail "Extract term has unexpected binder. Can this really happen?"
     end
 end
 
--- a/src/redprl/signature_mini.sml
+++ b/src/redprl/signature_mini.sml
@@ -118,14 +118,14 @@ struct
       let
         fun handleHyp ((u, psort), ptm, H) =
           case psort of
-             RedPrlSortData.HYP _ =>
-               let
-                 val v = case ptm of RedPrlParameterTerm.VAR v => v
-                            | _ => raise Fail "Hypothesis relabeling failed: not a variable"
-               in
-                 relabelHyp (u, v) H
-               end
-           | _ => H
+            RedPrlSortData.HYP _ =>
+              let
+                val v = case ptm of RedPrlParameterTerm.VAR v => v
+                           | _ => raise Fail "Hypothesis relabeling failed: not a variable"
+              in
+                relabelHyp (u, v) H
+              end
+          | _ => H
       in
         ListPair.foldl handleHyp H (#params entry, ps)
       end
--- a/src/redprl/signature_mini.sml
+++ b/src/redprl/signature_mini.sml
@@ -96,7 +96,8 @@ struct
           case psort of 
             RedPrlSortData.HYP tau =>
               let
-                val RedPrlParameterTerm.VAR v = ptm
+                val v = case ptm of RedPrlParameterTerm.VAR v => v
+                           | _ => raise Fail "Hypothesis renaming failed: not a variable"
               in
                 Var.Ctx.insert ctx u (Tm.check (Tm.`v, tau))
               end
@@ -116,10 +117,11 @@ struct
     fun relabelHyps (entry : entry) (ps : Tm.param list) H = 
       let
         fun handleHyp ((u, psort), ptm, H) =
-          case psort of 
+          case psort of
              RedPrlSortData.HYP _ =>
                let
-                 val RedPrlParameterTerm.VAR v = ptm
+                 val v = case ptm of RedPrlParameterTerm.VAR v => v
+                            | _ => raise Fail "Hypothesis relabeling failed: not a variable"
                in
                  relabelHyp (u, v) H
                end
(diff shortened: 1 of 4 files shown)


jonsterling:
@favonia I'm excited that you are working your way through the codebase and making improvements! 😄

src/redprl/signature_mini.sml
```diff
@@ -137,7 +139,7 @@ struct
         val entry = lookup sign opid
         val paramsSig = #params entry
         val argsSig = #arguments entry
-        val SOME goal = #spec entry
+        val goal = case #spec entry of SOME goal => goal | _ => raise Fail "Reviving theorem failed: goal missing"
```
favonia: @jonsterling I need your help in writing good error messages.

src/redprl/signature_mini.sml
```diff
@@ -96,7 +96,8 @@ struct
           case psort of 
             RedPrlSortData.HYP tau =>
               let
-                val RedPrlParameterTerm.VAR v = ptm
+                val v = case ptm of RedPrlParameterTerm.VAR v => v
+                           | _ => raise Fail "Hypothesis renaming failed: not a variable"
```
favonia: @jonsterling I need your help here in writing good error messages.

src/redprl/syntax.sml
```diff
@@ -111,7 +111,9 @@ struct
               | (O.TAG_DPROD, (A :: xB :: args)) => (O.MONO O.DPROD $$ [A, xB], args)
               | (O.TAG_PATH, (uA :: a0 :: a1 :: args)) => (O.MONO O.PATH_TY $$ [uA, a0, a1], args)
               | _ => raise Fail "Syntax.out hcom: Malformed tag"
-           val (_ \ cap) :: args = args
+           val (cap , args) = (case args of
+                                    ((_ \ cap) :: args) => (cap, args)
+                                  | _ => raise Fail "Syntax.out hcom: Missing cap")
```
favonia: @jonsterling I also need your help here to come up with a good error message.

src/redprl/signature.sml
```diff
@@ -449,21 +449,24 @@ struct
...
+                      ((x, RedPrlSortData.HYP tau) :: ps, Tm.Sym.Ctx.insert ctx x (RedPrlSortData.HYP tau), NameEnv.insert env (Sym.toString x) x))
+                    (params', symctx, env)
+                    hyps
+              in
+                convertToAbt (metactx, symctx', env') script TAC
+                  >>= (fn scriptTm => elabRefine sign (seqjdg, scriptTm))
+                  >>= checkProofState (pos, subgoalsSpec)
+                  >>= (fn state => E.ret @@ EDEF {sourceOpid = opid, params = params'', arguments = arguments', sort = tau, spec = SOME seqjdg, state = state})
+              end
+             | _ => raise Fail "Elaboration failed: no context in the derived rule")
```
favonia: @jonsterling I believe this is the reason why the elaboration should fail?

favonia:
After these changes the source should compile cleanly (no warnings).

src/redprl/signature_mini.sml
jonsterling: This is good.

src/redprl/syntax.sml
jonsterling: I would leave the partial binding as-is, rather than adding this case statement, since it is an invariant that this will never happen; I prefer the simple code. If you look at the refiner / rules, there are hundreds of partial matches which are essential for the rules being readable, for instance.

Also, I prefer the following indentation style:

```sml
val (cap , args) = 
  (case args of
      ((_ \ cap) :: args) => (cap, args)
    | _ => raise Fail "Syntax.out hcom: Missing cap")
 ```

The main idea of the indentation style that I prefer is that the indentation of some code should never depend on the *lexical* width of some other identifier; this means that I don't try and "line things up", but rather use indentation as a way to signify syntactic hierarchy. I think that there are some inevitable exceptions, due to ML's keywords, but in general I try to follow that rule.

And a number of redundant parentheses could also be removed, leading to:

```sml
val (cap , args) = 
  case args of
     (_ \ cap) :: args => (cap, args)
   | _ => raise Fail "Syntax.out hcom: Missing cap"
 ```

File: src/redprl/signature.sml (hunk 4 above)
Comment (jonsterling) [inline src/redprl/signature.sml]: Let's leave the partial binding here rather than adding the wildcard case, as this is another invariant which is ensured elsewhere.

File: src/redprl/signature_mini.sml (hunk 2 above)
Comment (favonia) [inline src/redprl/signature_mini.sml]: I assume you also want to keep this partial binding.

Comment (jonsterling) [inline src/redprl/signature_mini.sml]: Let's keep this one, since this is a place where I could imagine a bug happening. Your error message looks good.

Comment (jonsterling) [inline src/redprl/signature_mini.sml]: Oops, I mean "let's keep the code the way *you* wrote it". So, I mean that I like your adding the case statement.

File: src/redprl/signature_mini.sml [hunk 5]
```diff
@@ -116,10 +117,11 @@ struct
     fun relabelHyps (entry : entry) (ps : Tm.param list) H = 
       let
         fun handleHyp ((u, psort), ptm, H) =
-          case psort of 
+          case psort of
              RedPrlSortData.HYP _ =>
                let
-                 val RedPrlParameterTerm.VAR v = ptm
+                 val v = case ptm of RedPrlParameterTerm.VAR v => v
+                            | _ => raise Fail "Hypothesis relabeling failed: not a variable"
```
favonia: Forgot to label this one, but I will keep this change as well.

jonsterling: looks good.

jonsterling:
feel free to merge at will!

PR #201, opened 2017-06-13
Improved interaction model, and bugfixes

jonsterling:
I'll be ready to get this merged in soon, but I need to coordinate the release of a new version of the VS Code editor mode together with this.


favonia:
@jonsterling Question: how should I print out the goals when I saw "N Remaining Obligations"?

jonsterling:
@favonia At any point in a proof, you can write `?hello` (any word can be written instead of `hello`, only the question mark is mandatory); this will cause the goals that will be affected at that location to be printed. 

We used to just print all the goals at all times, but when you have a lot of them, it becomes difficult to work with as a user. so, now we let users control what goals they want to see.

PR #202, opened 2017-06-13
Fix SML/NJ build

jonsterling:
Apparently, in SML/NJ it is an error rather than a warning to have redundant cases in pattern matches. I have updated the Travis script to test SML/NJ as well as MLton.

@favonia By the way, this suggests to me that you have been developing RedPRL with MLton, which could be a little frustrating considering how long it takes for MLton to build RedPRL. It may be more fun for you if you use SML/NJ during development, since RedPRL takes only a few seconds to build there. The script `script/smlnj.sh` should get you started.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,15 @@
 language: sml
+env:
+  - COMPILER=smlnj PACKAGES="smlnj ml-yacc ml-ulex"
+  - COMPILER=mlton PACKAGES="mlton"
 before_install:
   - sudo apt-get update -qq
-  - sudo apt-get install -y --force-yes smlnj mlton
+  - sudo apt-get install -y --force-yes ${PACKAGES}
 install:
   - git submodule init
   - git submodule update --init --recursive
 script:
-  - ./script/test-mlton.sh
+  - ./script/test-${COMPILER}.sh
   - cat 'build.log'
 
 notifications:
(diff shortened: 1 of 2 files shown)


favonia:
Thanks, though I have been well-trained by Agda. FYI: The #199 pull request will remove all redundant rules.

favonia:
I wonder why travis suddenly cannot find the ML lex and yacc. By the way, maybe we should print out the build.log for both MLton and SML/NJ in test.sh, otherwise the travis script only prints out one of them? We could also save the log into two different files.

jonsterling:
@favonia Yeah...

For me right now, it is not a huge priority to get this working, but if you want to do so, feel free.

In the meanwhile, we should at least make sure the SML/NJ build works locally in each branch before merging.

favonia:
Now travis shows
```
Fatal error -- Uncaught exception SysErr with <unknown> raised at <export-fun.c>
```
I have no ideas what it is.

favonia:
@jonsterling For some reason I cannot request your review. So here is my comment. This is done!

favonia:
> Fatal error -- Uncaught exception SysErr with <unknown> raised at <export-fun.c>

This is because `bin/` does not exist during the first run.

jonsterling:
great job, thanks!!

PR #213, opened 2017-06-17
Update cmlib and sml-json.

favonia:


jonsterling:
@favonia So that it is less bureaucratic for you to make these kinds of changes across all the repositories, I hereby authorize you to self-merge any library upgrades 😄

Issue #192, opened 2017-06-11
Remove type_tag and redesign hcom in the stack machine semantics


favonia:
Currently `type_tag` is used to represent (the head symbol of) a value type and determine the behavior of the hcom being indexed, but one can skip these intermediate steps and directly dispatch on the value types in focus. After this optimization, `type_tag` can be completely removed. Note that we might need a separate hcom for base types where hcom's may be regarded as values.

jonsterling:
@favonia I have been thinking about this, and what can actually be done at this time is not so exciting. Here is the best that can be done:

We can add a special hcom for base types, but it will not necessarily be a value. It is a value if its extents all variables, and if `r,r'` are distinct. But this hcom will still need `type_tag`—in the operational semantics it is not necessary to know what type the hcom is from at this point, but if one of these base type hcoms shows up in a typing judgment, we would need to know the type. It would be possible to get around this by adding special cased typing rules for value hcoms in `bool` and in `S1`, but I would like to avoid this duplication.

What this means it that `type_tag` can be removed from "general hcoms" (the ones where the type is not yet resolved/computed), but that `type_tag` would have to be kept for base type hcoms. The only difference is that there would be no `TAG_NONE`.

All in all, these changes sound very superficial and unlikely to result in any conceptual improvement. I would suggest that we defer making any such changes until we can remove the `type_tag` ENTIRELY.

favonia:
@jonsterling In light of #207 I feel the separation may be useful.

favonia:
@jonsterling Related to this issue, is it possible to allow the atomic hcom stepper to return a new value after the substitution is (partially) forced? For the new hcom in #206, I expect that equations of hcom values in base types will have lots of 0 and 1. And if they are forced during the checking, I think we should keep the resolved values to save time in further checking.

jonsterling:
@favonia 

> In light of #207 I feel the separation may be useful

Maybe, but it seems like we could just as well implement #207 in the way you suggest, and then branch on the existing tags. I'm not totally opposed to what you say here, but I'd like to make sure that it doesn't cause a blow-up of code & bureaucracy.

>  Related to this issue, is it possible to allow the atomic hcom stepper to return a new value after the substitution is (partially) forced?

This sounds interesting, but I don't think I understand what you mean yet. Could you explain it to me in a little more detail?

favonia:
[...]

PR #216, opened 2017-06-18
Implement the stack machine without tags.

favonia:
This pull request will be a complete solution to issue #192.

There are still bugs and the code needs further clean-ups, but I think I should put the current progress on-line. @jonsterling How should I debug the machine? Even something as ugly as `printf` would help.


jonsterling:
@favonia `printf` sounds reasonable

favonia:
@jonsterling Hmm I want a more civilized method. How should I try a particular tactic and see why it fails? For example trying `HCom.CapEq` and seeing what went wrong?

jonsterling:
@favonia I'm afraid that `printf` debugging may really be the best you can do. You can instrument a particular rule with print statements to find out why it went wrong.

favonia:
@jonsterling A separate issue is that currently I put something like `intoCoe` in `Syntax` to generate raw expressions. Should they actually be in `Machine` or `OpData`?

jonsterling:
@favonia I think it's ok in `Syntax`.

src/redprl/syntax.sml
```diff
@@ -67,17 +82,45 @@ struct
...
+      O.POLY (O.COE dir) $$ [([u],[]) \ a, ([],[]) \ m]
+
+    fun intoCom (dir as (r, r'), eqs) ((u, a), cap, tubes) =
+      let
+        fun coe v m = intoCoe (v, r') ((u, a), m)
+        fun goTube (([v],_) \ n) = ([v],[]) \ coe (P.ret v) n
+          | goTube _ = raise Fail "malformed tube"
+      in
+        intoHcom (dir, eqs) (substSymbol (r', u) a, coe r cap, List.map goTube tubes)
+      end
```
favonia: @jonsterling A separate issue is that currently I put something like `intoCoe`, `intoCom`, etc in `Syntax` to generate raw expressions. Should they actually be in `Machine` or `OpData`?

jonsterling: `Syntax` is good.

favonia: Alright. So we have syntax-ish builder in `Syntax`, and others in `Tac` in the `Machine`. (I deleted your other response since I deleted my question there.)

jonsterling: @favonia Yeah... I guess the `Tac` things are in `Machine` because they aren't used anywhere else, and there's no reason to expect that they'd be used anywhere else. But if you prefer / find it more systematic, you can feel free to move any helpers that you want into `Syntax`.

I think that there's not yet much rhyme nor reason to how things are arranged, so if you find something that you wish were elsewhere, you can feel free to move it.

favonia:
@jonsterling I think you can save me some time by showing me how to run a particular tactic in the script. So far I only know `auto`, `lam`, `hyp`, `elim`, `?` and `'`.

jonsterling:
@favonia By design, users don't have access to primitive rules; instead, rules are applied through high-level tactics (like the ones you mention). In general, rules for equality are applied through `auto-step` (which is `auto` but without repeating at all). There's a few reasons for this design:

1. Like Nuprl, we want tactic scripts to be high-level and not sensitive to small changes. So, we prefer to have a few high-level tactics that call primitive rules on the basis of context.

2. There are dozens of primitive rules (and ultimately, there may be hundreds of them), so it is FAR too big an engineering burden to hook them all up to the parser, the syntax, etc. It would be intractable to allow all rules to be called by hand.

If you really need to debug this specific rule, and cannot access it indirectly (e.g. through `auto-step`), my suggestion is to temporarily add it to the syntax. But I hope you don't have to do that (I have never needed to in the past, and I've done a lot of debugging.)

favonia:
An interesting discovery is that SML/NJ fails to compile but MLton does. Without knowing further information I would guess SML/NJ is at fault.

favonia:
@jonsterling I think I know what is going on. A meta-question: I suppose you want users to be able to reason about `coe` and `hcom`, but `coe` was not in `Syntax.view` and refiners seem to rely on `Syntax`. Do you have any suggestion? Should we add `coe` back?

jonsterling:
Can you explain what the "bug" has to do with the syntax view? that seems like it should be a completely orthogonal issue, but I have clearly missed something.

favonia:
@jonsterling Previously, `hcom` is used as a operator (like `coe`) and also a value. Now `hcom` is never a value. I did not see `coe` in the view and wonder whether I should remove `hcom` from the syntax as well. However, I then realized the refiners should be able to reason about Kan types without knowing what the types are. The bug is that the refiners need to handle both `hcom` and `fhcom`, and probably `coe` as well, and I fixed it by writing code for both `hcom` and `fhcom`.

jonsterling:
@favonia Yes, the refiner needs to transparently handle both `hcom` and `hfcom`—one of the things I tried to get across in our previous discussion is that it would be nice if we didn't have to duplicate all the `hcom` rules for base-type-hcom. This is one of the reasons that we were holding onto the tags for base type compositions, so that we could "reify" them into a single form in the `Syntax` view; that way the refiner would work with that, and not need to know the difference.
 
As things are, it looks like we will need to pursue a different approach in the refiner. I'm open to ideas...

I just checked and it looks like we forgot to implement rules for `coe` all along, so we'll have to add these (what an oversight!).

favonia:
If you checked my code, you would see `ComKit` is largely shared, and more sharing can probably be done. By the way, the code for `fhcom` is not completed, but I feel we can fix it along with `coe` in another pull request or something.

jonsterling:
> If you checked my code, you would see ComKit is largely shared, and more sharing can probably be done.

Ah, I see now! I think I got confused when reading the diff. It looks reasonable 👍 

I agree that we can do whatever other things are necessary in another pull request, and that it's worth getting your very good work merged in as soon as we can.

PR #217, opened 2017-06-18
Really update the libs.

favonia:
Hashes were changed when merge nodes were created or commits were reworded.


jonsterling:
I didn't check this before, but in the future, we should try to avoid making `RedPRL` depend on a branch other than `master` of any of our libraries. The reason for this is that sometimes when debugging, it can be useful to try and build older versions of RedPRL and see when an issue was introduced. Since these library nodes don't point to anything anymore, it means that it is not possible to consistently build some versions of RedPRL.

favonia:
I was aware of this issue but thought the merging on GitHub would be a simple fast-forwarding.

PR #218, opened 2017-06-18
Prove theorems about coe and fhcom.

favonia:
Work in progress.

Goals:
- [x] Add coe.
- [x] Add rules fhcom except for the issues that should be fixed after `safeStep` is redone.
- [x] Add testing for fhcom.
- [x] Add rules for coe.
- [x] Add testing for coe.

Dream:
- [ ] Revamp `safeStep`.

### DIFF
--- a/src/redprl/refiner.fun
+++ b/src/redprl/refiner.fun
@@ -1177,6 +1177,54 @@ struct
     end
   end
 
+  structure Coe =
+  struct
+    fun Eq alpha jdg =
+      let
+        val _ = RedPrlLog.trace "Coe.Eq"
+        val H >> CJ.EQ ((lhs, rhs), ty) = jdg
+        val Syn.COE {dir=(r0, r'0), ty=(u, ty0u), coercee=m0} = Syn.out lhs
+        val Syn.COE {dir=(r1, r'1), ty=(v, ty1v), coercee=m1} = Syn.out rhs
+        val () = assertParamEq "Coe.Eq source of direction" (r0, r1)
+        val () = assertParamEq "Coe.Eq target of direction" (r'0, r'1)
+
+        val ty01 = substSymbol (r'0, u) ty0u
+        val (goalTy1, _) = makeGoal @@ ([],[]) || H >> CJ.EQ_TYPE (ty01, ty)
+
+        val w = alpha 0
+        val ty0w = substSymbol (P.ret w, u) ty0u
+        val ty1w = substSymbol (P.ret w, v) ty1v
+        val (goalTy, _) = makeGoal @@ ([(w, P.DIM)],[]) || H >> CJ.EQ_TYPE (ty0w, ty1w)
+
+        val ty00 = substSymbol (r0, u) ty0u
+        val (goalCoercees, _) = makeGoal @@ ([],[]) || H >> CJ.EQ ((m0, m1), ty00)
+      in
+        T.empty >: goalTy1 >: goalTy >: goalCoercees #> trivial
+      end
+
+    fun CapEq alpha jdg =
+      let
+        val _ = RedPrlLog.trace "Coe.CapEq"
+        val H >> CJ.EQ ((lhs, rhs), ty) = jdg
+        val Syn.COE {dir=(r, r'), ty=(u, tyu), coercee=m} = Syn.out lhs
+        val () = assertParamEq "Coe.CapEq source and target of direction" (r, r')
+
+        val ty0 = substSymbol (r, u) tyu
+        val (goalTy0, _) = makeGoal @@ ([],[]) || H >> CJ.EQ_TYPE (ty0, ty)
+
+        val (goalTy, _) = makeGoal @@ ([(u, P.DIM)],[]) || H >> CJ.TYPE tyu
+        val (goalEq, _) = makeGoal @@ ([],[]) || H >> CJ.EQ ((m, rhs), ty)
+      in
+        T.empty >: goalTy0 >: goalTy >: goalEq #> trivial
+      end
+
+    local
+      infix orelse_
+    in
+      (* Try all the fhcom rules. *)
+      val AutoEq = Eq orelse_ CapEq
+    end
+  end
 
   structure Computation =
   struct
@@ -1450,6 +1498,8 @@ struct
         case (Syn.out m, Syn.out n) of
            (Syn.HCOM _, _) => HCom.AutoEq
          | (_, Syn.HCOM _) => Equality.Symmetry
+         | (Syn.COE _, _) => Coe.AutoEq
+         | (_, Syn.COE _) => Equality.Symmetry
          | _ => StepEqCanonicity sign ((m, n), ty)
 
       fun StepSynth sign m =
--- a/src/redprl/refiner.fun
+++ b/src/redprl/refiner.fun
@@ -1371,19 +1371,19 @@ struct
     local
       fun StepTrue sign ty =
         case Syn.out ty of
-           Syn.DFUN _ => DFun.True
+           Syn.PATH_TY _ => Path.True
+         | Syn.DFUN _ => DFun.True
          | Syn.DPROD _ => DProd.True
-         | Syn.PATH_TY _ => Path.True
          | _ => raise E.error [E.% "Could not find introduction rule for", E.! ty]
 
       fun StepEqTypeVal (ty1, ty2) =
         case (Syn.out ty1, Syn.out ty2) of
            (Syn.BOOL, Syn.BOOL) => Bool.EqType
          | (Syn.S_BOOL, Syn.S_BOOL) => StrictBool.EqType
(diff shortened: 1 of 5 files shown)


favonia:
@jonsterling What would be your suggestion about fixing `safeStep`? I see several possibilities:

- Easy but worse approach: the user-defined global state `sign` has a toggle `stable`/`unstable`, and under the `stable` mode, `fhcom`, `loop`, and other stuff will be neutral if the valueness depends on knowing something is a variable. The `unstable` mode is the current dynamics.

- More refined but complicated approach: in addition to the above approach, when we reach `coe_{x.ia_x}` and try to evaluate `ia_x`, we also mark that `x` is bound. That is, we further partition free parameter variables into "unbound" and "bound", and that `loop_x` and `ia_x` are still values when `x` is bound even under the `stable` mode.

I believe the second approach will be easier with some support from the libs. What do you think?

jonsterling:
I need to examine the machine a little bit to see what I think makes sense...

jonsterling:
@favonia Do you think that (1,2) can be done separately? Maybe what makes sense is to start with (1) and then we can add (2) soon.

favonia:
@jonsterling Maybe, but not when you are refactoring the entire refiner. I am not so sure about the flow of `sign` in the code and whether it can really be used as a mode switcher.

favonia:
@jonsterling I feel it might be easier for you to write some examples for `coe` for testing if we want to get this merged faster.

jonsterling:
@favonia We can defer the refactoring of the refiner until your stuff is done. In the meanwhile, I can come up with a few very basic test cases for coe.

src/redprl/refiner.fun
```diff
@@ -1177,6 +1177,54 @@ struct
...
+        val ty1w = substSymbol (P.ret w, v) ty1v
+        val (goalTy, _) = makeGoal @@ ([(w, P.DIM)],[]) || H >> CJ.EQ_TYPE (ty0w, ty1w)
+
+        val ty00 = substSymbol (r0, u) ty0u
+        val (goalCoercee0, _) = makeGoal @@ ([],[]) || H >> CJ.MEM (m0, ty00)
+        val ty10 = substSymbol (r0, v) ty1v
+        val (goalCoercee1, _) = makeGoal @@ ([],[]) || H >> CJ.MEM (m1, ty10)
+      in
+        T.empty >: goalTy1 >: goalTy >: goalCoercee0 >: goalCoercee1 #> trivial
+      end
```
jonsterling: @favonia I'm having trouble understanding this rule. When trying it out, I found that it induced some unexpected subgoals, and after examining the code, I am trying to figure out how it relates to the rule in the paper:

screen shot 2017-06-18 at 3 40 01 pm

Am I missing something?

jonsterling: This is the code I would have expected for the rule from the paper:

```sml
    fun Eq alpha jdg =
      let
        val _ = RedPrlLog.trace "Coe.Eq"
        val H >> CJ.EQ ((lhs, rhs), ty) = jdg
        val Syn.COE {dir=(r0, r'0), ty=(u, ty0u), coercee=m0} = Syn.out lhs
        val Syn.COE {dir=(r1, r'1), ty=(v, ty1v), coercee=m1} = Syn.out rhs
        val () = assertParamEq "Coe.Eq source of direction" (r0, r1)
        val () = assertParamEq "Coe.Eq target of direction" (r'0, r'1)

        val w = alpha 0
        val ty0w = substSymbol (P.ret w, u) ty0u
        val ty1w = substSymbol (P.ret w, v) ty1v
        val (goalTy, _) = makeGoal @@ ([(w, P.DIM)],[]) || H >> CJ.EQ_TYPE (ty0w, ty1w)

        val ty00 = substSymbol (r0, u) ty0u
        val (goalCoercees, _) = makeGoal @@ ([],[]) || H >> CJ.EQ ((m0, m1), ty00)
      in
        T.empty >: goalTy >: goalCoercees #> trivial
      end
```

favonia: I was thinking that maybe it makes more sense to generate the goal for `m1` from `ty1v`, but the downside is that there would be duplicate goals.... actually I now think the binary version is better. However you still have to show `CJ.EQ_TYPE (ty01, ty)`. The conclusion may or may not match `ty `.

src/redprl/refiner.fun
```diff
@@ -1177,6 +1177,54 @@ struct
...
+        val H >> CJ.EQ ((lhs, rhs), ty) = jdg
+        val Syn.COE {dir=(r, r'), ty=(u, tyu), coercee=m} = Syn.out lhs
+        val () = assertParamEq "Coe.CapEq source and target of direction" (r, r')
+
+        val ty0 = substSymbol (r, u) tyu
+        val (goalTy, _) = makeGoal @@ ([],[]) || H >> CJ.EQ_TYPE (ty0, ty)
+        val (goalEq, _) = makeGoal @@ ([],[]) || H >> CJ.EQ ((m, rhs), ty)
+      in
+        T.empty >: goalTy >: goalEq #> trivial
+      end
```
jonsterling: Likewise, I don't understand this rule. I would have expected:

```sml
    fun CapEq alpha jdg =
      let
        val _ = RedPrlLog.trace "Coe.CapEq"
        val H >> CJ.EQ ((lhs, rhs), ty) = jdg
        val Syn.COE {dir=(r, r'), ty=(u, tyu), coercee=m} = Syn.out lhs
        val () = assertParamEq "Coe.CapEq source and target of direction" (r, r')

        val ty0 = substSymbol (r, u) tyu
        val (goalTy, _) = makeGoal @@ ([(u, P.DIM)],[]) || H >> CJ.TYPE ty
        val (goalEq, _) = makeGoal @@ ([],[]) || H >> CJ.EQ ((m, rhs), ty0)
      in
        T.empty >: goalTy >: goalEq #> trivial
      end
```

src/redprl/refiner.fun
jonsterling: @favonia Now I see what you're saying... I still think the type equality subgoal is slightly incorrect, however. Let me fiddle with it  little bit.

src/redprl/refiner.fun
jonsterling: Now I see what you were doing. nvm!

src/redprl/refiner.fun
jonsterling: I think I found the issue—the wrong `r` was substituted in. I'll push a fix.

jonsterling:
@favonia OK, I've added a test and got it working. Can you look over my change to make sure you think it's ok?

src/redprl/refiner.fun
jonsterling: @favonia Agreed; except that `ty01` is using the wrong `r`, which I have fixed in https://github.com/RedPRL/sml-redprl/pull/218/commits/eafc9f3840861b888e29240b121d83b8c034a520.

favonia: @jonsterling No, it should be either `r'1` or `r1` but not `r0`. `r'0` is `r` on paper.

favonia: @jonsterling Ah, good catch. I was confused in so many ways.

favonia:
@jonsterling Yay! Would you mind if I clean up the commit history before merging? I am worried that some of your authorship might disappear.

jonsterling:
@favonia Feel free! I don't mind.

favonia:
Hmm I want to add another test with the assumption `#A{r} = #B` to see whether it can figure out that the `coe` is in `#B`, but I could not figure out the syntax fast enough.

jonsterling:
@favonia I can predict that it would not be able to do this automatically; I'm not even sure whether all the necessary rules are in place for doing that, but it wouldn't be hard to implement. (Rules for things like this have been implemented on an as-needed basis: when we want to prove something that requires a rule, we add it 😄 )

favonia:
@jonsterling Hmm. The strange thing is that the precondition `#A{r} = #B` disappeared when I was trying, but I guessed it is because I did not learn the syntax yet, and so I gave up.

favonia:
@jonsterling I wanted to add a rule for `CapEq` but the goal it shows is
```
  Goal 30.
    m : #A{r} true
    ⊢ #A{r} = #A{r} type
```
I am not sure how to proceed. Also if this is too difficult I would want to give up for now.

jonsterling:
@favonia The problem was that the system checks whether the expected premises are EXACTLY what is written in the derived rule. (The "derived rules" mechanism is still a little janky; it will be replaced with something very elegant once we are able to switch to the higher order logical framework.) So, I fixed your test by removing the unused premise.

favonia:
@jonsterling Ah, that means there is a bug in the rule. BTW I actually forgot Carlo already put a proof theory there.

favonia:
I think we have suffered enough for this pull request. If there is any bug we can fix it later.

Issue #214, opened 2017-06-17
Get rid of fancy treatment of 'generic judgment'


jonsterling:
I am using an experimental treatment of generic judgments (in https://github.com/RedPRL/sml-dependent-lcf), which enables us to write rules in "local" style and have all the necessary abstractions and instantiations dealt with in a semi-automatic style.

However, I don't think that the implementation is correct (based on some bugs I have seen), and it is complex enough that I am not confident that I can fix it.

Lately, I have been experimenting elsewhere, and I think that it makes sense to go back to what I did originally, which is require explicit abstractions of the *full context* in refinement rules. This is a little bit more bureaucratic, but because it is simple-minded, it is easy to do correctly.

This is a lot of work, but I think it will be important to do it sooner rather than later.

jonsterling:
This is possibly related to #209.

jonsterling:
To do this, we will need to switch back to the "vanilla" version of Dependent LCF, and then make the following changes in RedPRL:

1. The parameter/dimension context will need to be part of the sequent.
2. All metavariables must be fully applied to the spine of dimensions and terms; we can provide some "shortcuts" for making this mostly non-bureaucratic.
3. All extracts of judgments must be fully abstracted wrt the current dimension & hypothesis context.

This is a fair bit of work, but it may solve a couple of bugs. Then, we should delete the "generic judgment" version of Dependent LCF, which I believe is not correct.

PR #222, opened 2017-06-18
Fixing generic judgments

jonsterling:
(Issue #214)

This was surprisingly less painful than expected; at least, it was way more painful to get everything into the state that it was than to undo that.

- [x] The only thing that doesn't work yet is the derived rules machinery. I'll fix that.


jonsterling:
I'm going to merge this, as it seems to work (and doesn't conflict with any other ongoing work). The only dark places are in the signature elaborator, having to do with things like symbol parameters... This will require some further fixing, but for now I am content to merge.

PR #225, opened 2017-06-19
Clean-ups, stage 3.

favonia:
Work in progress.

### DIFF
--- a/src/redprl/redprl.grm
+++ b/src/redprl/redprl.grm
@@ -149,10 +149,6 @@ end
 
  | symBinding of string * P.param_sort
  | symBindings of (string * P.param_sort) list
- | varBinding of string * O.sort
- | varBindings of (string * O.sort) list
- | annotatedBinder of (string * P.param_sort) list * (string * O.sort) list
-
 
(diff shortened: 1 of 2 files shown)


jonsterling:
Good catch. By the way, how are you finding all this unused code? Is there a flag I can turn on when I build RedPRL so that I see warnings for this?

favonia:
@jonsterling MLton supports annotations in the `mlb` file and ml-yacc will output the `desc` file (because we put `%verbose` in the `grm` file). Try this:

```patch
diff --git a/src/redprl.mlb b/src/redprl.mlb
index 1877e85..196bb27 100644
--- a/src/redprl.mlb
+++ b/src/redprl.mlb
@@ -1,3 +1,7 @@
+ann
+  "forceUsed"
+  "warnUnused true"
+in
 local
   $(SML_LIB)/basis/basis.mlb
   $(SML_LIB)/mlyacc-lib/mlyacc-lib.mlb
@@ -73,3 +77,4 @@ in
   structure RedPrlError
   structure RedPrlLrVals
 end
+end
```

PR #226, opened 2017-06-19
Add COM, part one.

favonia:
- [x] Add the `COM` operator.
- [x] Add the view for `COM`.
- [x] Add rules for `COM`.
- [ ] Add concrete user syntax for `COM`.
- [ ] Make sure `COM`, `HCOM` and `FHCOM` reuse the same code as much as possible.
- [x] Clean up the code more.
- [ ] Clean up the code even more.


favonia:
@cangiuli I am slightly worried about the proof theory of `COM`. I guess the right order of proof searching (with no backtracking after a rule is applied) is

1. Check the cap equation. If it matches, trigger the cap rule.
2. Check the tube equations. If the first one matches, trigger the tube rule.
3. Otherwise, trigger the structural eq rule.

The first two rules are absolutely invertible. Not 300% sure about the third one even if we know the first two fail.

jonsterling:
@favonia My suggestion is that we write the rules that make sense first, and see how they work out in practice before thinking too hard about things about invertibility. That is, we should avoid obvious violations, but we should not try to predict in advance all contingencies.

Concepts like "invertibility" make sense only after a careful analysis of a collection of rules as a mathematical object, which we have not yet undertaken. So, I think that we should be driven by practice and empirical testing at this point. If we find that the combination of rules and tactics lead to unrecoverable situations in practice, then we should take a closer look.

favonia:
@jonsterling Once again I am not able to satisfy `coe.prl` even if I thought I fed it with exactly the same judgment it was looking for.

jonsterling:
@favonia I can have a look at it.

jonsterling:
@favonia The test can be made to pass by removing a premise from the "derived rules" as follows:

```
Rule CoeTest{r:dim, r':dim}(#A : {dim}.exp) : [
  {x:dim} | m : #A{r} >> #A{x} = #A{x} type
  ==>
  m : #A{r} >> coe{r ~> r'}({x}.#A{x}; m) in #A{r'}
] by [
  auto;
  ?hole
].

Rule CoeTestCap{r:dim}(#A : {dim}.exp) : [
  {x:dim} | m : #A{r} >> #A{x} = #A{x} type
  ==>
  m : #A{r} >> coe{r ~> r}({x}.#A{x}; m) = m in #A{r}
] by [
  auto
].
```

I think the premise that used to be there was necessary under the old rules, but was semantically redundant. So, it looks like you must have made a nice improvement!

favonia:
@jonsterling Wait... so RedPRL is actually complaining about some unused premise? I certainly did not think of this possibility.

jonsterling:
The (currently fragile) derived rules mechanism which is used in that test works as follows:

1. you specify a rule, in the form of premises and a conclusion
2. write a tactic script
2. redprl checks if that tactic script takes that conclusion to exactly those premises

The usual notion of theorem is the special case of a derived rule where no premises are given.

jon

favonia:
So I think I would like to have this merged before adding the concrete syntax and examples. There are some structural improvements that I think are important even without `COM`.

jonsterling:
Sounds good! I'll review this PR directly after my meeting with bob. nice work!

favonia:
@jonsterling Thanks. By the way, I allowed more equal types in the same spirit of #228---those rules should be applicable even if alpha-equivalence on types fails. All the alpha-equivalence assertion on terms is replaced by conditional goal making.

src/redprl/machine.sml
```diff
@@ -168,6 +168,19 @@ struct
          S.CUT
            @@ (coe `$ [([u],[]) \ S.HOLE, ([],[]) \ S.% m], a)
            <: env
+     | O.POLY (O.COM (dir as (r, r'), eqs)) `$ (([u],_) \ a) :: (_ \ cap) :: tubes <: env =>
+         let
+           fun coe v m = Syn.intoCoe (v, r') ((u, a), m)
```
jonsterling: It may be clearer to write something like `s` instead of `v` here, because we usually use `u,v` as schematic variables that range over dimension variables only.

src/redprl/refiner_composition_kit.fun
```diff
@@ -138,23 +136,26 @@ struct
...
         val eqs1 = List.map #1 tubes1
-        val _ = ListPair.mapEq (assertEquationEq "HCom.Eq equations") (eqs0, eqs1)
-        val _ = assertTautologicalEquations "HCom.Eq tautology checking" eqs0
+        val _ = ListPair.mapEq (Assert.equationEq "HCom.Eq equations") (eqs0, eqs1)
+        val _ = Assert.tautologicalEquations "HCom.Eq tautology checking" eqs0
 
-        val (goalTy, _) = makeGoal @@ (I, H) >> CJ.EQ_TYPE (ty0, ty1)
-        val (goalCap, _) = makeGoal @@ (I, H) >> CJ.EQ ((cap0, cap1), ty)
+        (* type *)
+        val goalTy0 = makeEqTypeIfDifferent (I, H) (ty0, ty)
```
jonsterling: good idiom, nice.

src/redprl/refiner_composition_kit.fun
```diff
@@ -188,16 +191,114 @@ struct
...
+    val TubeEqR = catJdgFlipWrapper TubeEqL
+
+    (* Try all the hcom rules.
+     * Note that the EQ rule is invertible only when the cap and tube rules fail. *)
+    val AutoEqLR = CapEqL orelse_ CapEqR orelse_ TubeEqL orelse_ TubeEqR orelse_ Eq
+    val AutoEqL = CapEqL orelse_ TubeEqL orelse_ Eq
+    val AutoEqR = CapEqR orelse_ TubeEqR orelse_ Eq
+  end
+
+  structure Com =
```
jonsterling: Are these rules written down somewhere, or did you just derive what they should be by looking at the rules for `coe` and `hcom`? If the latter, that's fine, and I'll just take your word for it that they are correct 😄

favonia: I just derived what they should be. They were not tested so I was actually a little bit worried... but then I thought we can fix it later anyways.

jonsterling: @favonia I agree that we can fix them later if they are not quite right... Once we add tests, it may become clear whether these are right or wrong, so we can wait until then.

Ultimately, I expect that @cangiuli would write something down about this at some point, so we can just compare the rules that he writes down with what is implemented, when it is time.

favonia: @jonsterling Actually @cangiuli already has a theorem about `COM`, which is different from proof rules, but let me check it...

favonia:
@jonsterling Apparently I just found some bugs in the COM rule. :laughing:

PR #227, opened 2017-06-19
Try Path.ApConstCompute whenever possible. Fix #220.

### LINKED ISSUE #220: Tactics for path application
@jonsterling This
```
Thm Test : [
     paths({_}.bool; tt; tt)
  -> paths({_}.bool; tt; tt)
] by [
  { lam p. <i> 'fhcom{i=0, i=1; 0~>1}(,p @ i; {_}. ,p @ 0; {_}. ,p @ 1) };
  <| auto |>;
  ?hello
].
```
generated two goals not solved by `auto`:
```
test/bool-fhcom-regular.prl:7.3-7.9 [Info]:
  hello.
  Goal 26.
  [p : exp]. 
    p : paths({_}.bool;tt;tt) true
    ⊢ p @ 0 = tt : bool
  
  Goal 20.
  [p : exp]. 
    p : paths({_}.bool;tt;tt) true
    ⊢ p @ 1 = tt : bool
```
I think the problem is in the tactics for paths, but I am not touching the refiner now.

favonia:
`StepEqNeuVal` should try `ApComputeConst`.

### DIFF
--- a/src/redprl/refiner.fun
+++ b/src/redprl/refiner.fun
@@ -1485,7 +1485,7 @@ struct
            (_, Syn.DPROD _) => DProd.Eta
          | (_, Syn.DFUN _) => DFun.Eta
          | (_, Syn.PATH_TY _) => Path.Eta
-         | _ => raise E.error [E.% "Could not simplify neutral term or find eta expansion rule for type", E.! ty]
+         | _ => raise E.error [E.% "Could not expand neutral term of type", E.! ty]
 
       fun StepEqCanonicity sign ((m, n), ty) =
         case (Machine.canonicity sign m, Machine.canonicity sign n) of
--- a/src/redprl/refiner.fun
+++ b/src/redprl/refiner.fun
@@ -737,9 +737,9 @@ struct
           #> (I, H, trivial)
       end
 
-    fun ApComputeConst alpha jdg =
+    fun ApConstCompute alpha jdg =
       let
-        val _ = RedPrlLog.trace "Path.ApComputeConst"
+        val _ = RedPrlLog.trace "Path.ApConstCompute"
         val (I, H) >> CJ.EQ ((ap, p), a) = jdg
         val Syn.PATH_AP (m, P.APP r) = Syn.out ap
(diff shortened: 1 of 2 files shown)


src/redprl/refiner.fun
```diff
@@ -1477,25 +1477,27 @@ struct
...
          | (_, Syn.PATH_AP (_, P.APP _), _) => Equality.Symmetry
          | _ => raise E.error [E.% "Could not find neutral equality rule for", E.! m, E.% "and", E.! n, E.% "at type", E.! ty]
 
-      fun StepEqEta ty =
-        case Syn.out ty of
-           Syn.DPROD _ => DProd.Eta
-         | Syn.DFUN _ => DFun.Eta
-         | Syn.PATH_TY _ => Path.Eta
-         | _ => raise E.error [E.% "Could not find eta expansion rule for type", E.! ty]
+      fun StepEqNeuVal (m, ty) =
```
jonsterling: These changes look good; my only concern is the name, since `ApConstCompute` doesn't turn a neutral into a value, but a neutral into something that may be a redex.

Maybe we should rename `StepEqNeuVal` to `StepComputeNeut`, and `StepEqValVal` back to `StepEqVal`, and `StepEqNeuNeu` to `StepEqNeu`. My thought is that eta and beta principles can be captured under things with the word `...Compute...` in them, and other *structural*/symmetric principles can be captured under `...EqNeu` and `...EqVal`.

favonia: Are you suggesting a new name for `ApConstCompute`? How about reserving `Compute` to beta rules and make up some other word for non-structural rules?

jonsterling: @favonia Nah, I was thinking of new names for `StepEqNeuVal`, `StepEqValVal`. Anyway, the code as-written looks correct. Reserving `Compute` for beta-like rules quite sensible, so maybe we can think of something else for other principles, as you say.

In any case, if you don't want to think about names just yet, feel free to merge whenever you like, since this looks like a good change.

Issue #209, opened 2017-06-17
Make connections work!


favonia:
It will be great if we can type check this with only `auto`? (Currently RedPRL is quite slow and looks buggy on this, not to mention proving it.)

```
Thm Connection(#A) : [
  a/type : #A type >>
     (a : #A)
  -> (b : #A)
  -> (p : paths({_}.#A; a; b))
  -> paths({i}.(paths({_}.#A; a; p@i)); <_>a; p)
] by [
  {
  lam a. lam b. lam p.
    <i> <j> 'hcom{i=0,i=1,j=0,j=1;0~>1}(#A; ,a;
     {k}.hcom{k=0,k=1;1~>j}(#A; hcom{k=0,k=1;1~>0}(#A; ,p @ k; {_}.,a; {j}.,p @ j); {_}.,a; {_}.,a);
     {k}.hcom{k=0,k=1;1~>j}(#A; ,p @ k; {_}.,a; {j}.,p @ j);
     {k}.hcom{k=0,k=1;1~>i}(#A; hcom{k=0,k=1;1~>0}(#A; ,p @ k; {_}.,a; {i}.,p @ i); {_}.,a; {_}.,a);
     {k}.hcom{k=0,k=1;1~>i}(#A; ,p @ k; {_}.,a; {j}.,p @ j))
  };
].
```

jonsterling:
@favonia I experimented with this and there is just a blowup of proof obligations.

jonsterling:
And there must be some bugs somewhere too, because I am seeing incorrect subgoals arising. It is possible that either the bug is in the rules, or in the refinement apparatus (the latter is probably too complex for you to debug, but I can look into it.)

jonsterling:
I have at least mostly fixed the performance problems which made this kind of impossible to test (the blow-up of subgoals was partially caused by using a "parallel" version of tactic sequencing, in which substitutions don't propagate along the subgoals telescope until after all subgoals have been refined.)

I still get goals now which look either non-trivial or untrue. But at least we can try and figure out what is going on.

favonia:
Fixed a bug in the example. (Finally it's RedPRL helping me debugging.) However there are still four goals left unsolved.

jonsterling:
@favonia Awesome! And once I get around to adding *refinement* rules for compositions (so that you don't have to use `'`/`exact`), it will be easier to write correct compositions anyway.

[...]

PR #228, opened 2017-06-19
Allow equal types in Hyp.

favonia:
- [x] Wait for #227.
- ~~Review every assertion of alpha-equivalence of types. In general it is too strict.~~ Ill-conceived.

This kills one of the lingering goals in #209.

### DIFF
--- a/src/redprl/refiner.fun
+++ b/src/redprl/refiner.fun
@@ -931,10 +931,17 @@ struct
              CJ.TRUE ty => ty
            | _ => raise E.error [E.% "Equality.Hyp: expected truth hypothesis"]
       in
+        (* If the types are identical, there is no need to create a new subgoal (which would amount to proving that 'ty' is a type).
+           This is because the semantics of sequents is that by assuming that something is a member of a 'ty', we have
+           automatically assumed that 'ty' is a type. *)
         if Syn.Tm.eq (ty, ty') then
           T.empty #> (I, H, trivial)
         else
-          raise E.error [E.% @@ "Expected type of hypothesis " ^ Var.toString x ^ " to be", E.! ty, E.% "but found", E.! ty']
+          let
+            val (goalTy, _) = makeGoal @@ (I, H) >> CJ.EQ_TYPE (ty, ty')
+          in
+            T.empty >: goalTy #> (I, H, trivial)
+          end
       end
       handle Bind =>
         raise E.error [E.% "Expected variable-equality sequent"]


favonia:
@jonsterling This again breaks the fragile coe testing.... I will fix it later.

favonia:
@jonsterling To double check, I think we should replace every `assertAlphaEq` except for the `CEquiv.Refl`, right? Actually, what is `CEquiv.Refl` for?

jonsterling:
@favonia That doesn't really make any sense... We actually do need to use alpha equivalence; this is how PRLs work. Rules should not be doing non-trivial equational reasoning *ever*, instead it is the responsibility of tactics to get the goals into such a state that rules can apply.

src/redprl/refiner.fun
```diff
@@ -954,11 +954,9 @@ struct
           case catjdg of
              CJ.TRUE ty => ty
            | _ => raise E.error [E.% "Equality.Hyp: expected truth hypothesis"]
+        val (goalTy, _) = makeGoal @@ (I, H) >> CJ.EQ_TYPE (ty, ty')
```
jonsterling: The reason we should force `hyp` to use alpha equivalence rather than making a new subgoal is primarily that we want `hyp` to fail synchronously (in the proof theoretic sense). This is better for automation.

The behavior you want can be emulated in a tactic.

jonsterling:
@favonia I just double-checked, and I had apparently mis-read your code. Your code does update the correct rule, so we can think about incorporating your change.

jonsterling:
@favonia If you get this branch up to date with master, I can look at it again and we should try to get it merged. We can also try to figure out what went wrong with the test.

favonia:
@jonsterling Not sure how to fix the coe example.

jonsterling:
I can look into it and see what can be done

favonia:
Okay, so if I understand it correct, you are proposing (1) testing alpha eq first (2) if they match, shut up and let it go (3) if not, then generate a goal. Did I miss something? I can minimize the diff to make this pull request beautiful.

jonsterling:
@favonia Yeah, that's a correct summary. The idea is that if they match and we generate a subgoal, that subgoal will be true, but harder to prove (and ultimately unnecessary to prove).

favonia:
@jonsterling I think it is a nice thing. In fact, I am going to experiment new refiners in another pull request that do not generate unnecessary goals for the cases where alpha-equivalence is enough.

favonia:
@jonsterling I am going to destroy the commit history of this pull request. You have been warned.

PR #231, opened 2017-06-19
Try the cap and tube rules on both sides.

### LINKED ISSUE #209: Make connections work!

favonia:
Together with #228, this will fix #209.

### DIFF
--- a/src/redprl/refiner.fun
+++ b/src/redprl/refiner.fun
@@ -29,6 +29,8 @@ struct
 
   fun orelse_ (t1, t2) alpha = Lcf.orelse_ (t1 alpha, t2 alpha)
 
+  fun swapEqJdg ((lhs, rhs), ty) = ((rhs, lhs), ty)
+
   structure S1 =
   struct
     fun EqType _ jdg =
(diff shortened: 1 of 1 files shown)


favonia:
I deleted the example so that we can cleanly merge this first while fixing the coe example in #228.

jonsterling:
@favonia Sorry for the delay; I'll look at this now.

favonia:
@jonsterling I can refactor the code even further---what we want (as a ugly hack) is a wrapper to flip the top equational judgment. In fact, the EQ rule should try both sides too, but so far it is sufficient to prove the current examples.

jonsterling:
@favonia If it works, let's leave it the way it is for now, since I think the code is reasonable looking.

PR #238, opened 2017-06-21
Immortal pretty printer and sexpr-style concrete notation

jonsterling:
This is the first step toward migrating to use the ~immortal pretty printer~ Immortal Pretty Printer. This PR changes the *term printer* to use the ~immortal library~ Immortal Library, and removes our dependency on that fixity unparsing library.

In later stages, we will make the use of the ~immortal library~ Immortal Library pervasive throughout RedPRL. Another thing we'll aim for is that the rendering only ever occur at the edges of RedPRL, so that we never have to compute these documents to a string, which is probably expensive.

### DIFF
--- a/src/redprl/judgment.fun
+++ b/src/redprl/judgment.fun
@@ -12,7 +12,8 @@ struct
 
   val subst = S.map o Tm.substMetaenv
   val eq = S.eq
-  val toString = S.toString TermPrinter.toString
+  fun toString _ = "TODO"
+  (* S.toString TermPrinter.toString*)
 
   local
     open S
@@ -31,6 +32,6 @@ struct
            end
   end
 end
-
+ 
 structure RedPrlSequent = Sequent (structure CJ = RedPrlCategoricalJudgment)
 structure RedPrlJudgment = SequentJudgment (structure S = RedPrlSequent and TermPrinter = TermPrinter)
--- a/src/redprl/judgment.fun
+++ b/src/redprl/judgment.fun
@@ -1,6 +1,6 @@
 functor SequentJudgment
   (structure S : SEQUENT where type 'a CJ.Tm.O.Ar.Vl.Sp.t = 'a list and type CJ.Tm.Sym.t = Sym.t and type CJ.Tm.Var.t = Sym.t
-   structure TermPrinter : sig type t = S.CJ.Tm.abt val ppTerm : t -> FinalPrinter.doc end) : LCF_JUDGMENT  =
+   structure TermPrinter : sig type t = S.CJ.Tm.abt val ppTerm : t -> Fpp.doc end) : LCF_JUDGMENT  =
 struct
   structure CJ = S.CJ
   structure Tm = CJ.Tm
--- a/src/redprl/judgment.fun
+++ b/src/redprl/judgment.fun
@@ -1,6 +1,6 @@
 functor SequentJudgment
   (structure S : SEQUENT where type 'a CJ.Tm.O.Ar.Vl.Sp.t = 'a list and type CJ.Tm.Sym.t = Sym.t and type CJ.Tm.Var.t = Sym.t
-   structure TermPrinter : SHOW where type t = S.CJ.Tm.abt) : LCF_JUDGMENT  =
+   structure TermPrinter : sig type t = S.CJ.Tm.abt val ppTerm : t -> FinalPrinter.doc end) : LCF_JUDGMENT  =
 struct
   structure CJ = S.CJ
   structure Tm = CJ.Tm
@@ -12,8 +12,7 @@ struct
 
   val subst = S.map o Tm.substMetaenv
   val eq = S.eq
-  fun toString _ = "TODO"
-  (* S.toString TermPrinter.toString*)
+  val toString = FppRenderPlainText.toString o FinalPrinter.execPP o S.pretty TermPrinter.ppTerm
 
   local
     open S
--- a/src/redprl/judgment.fun
+++ b/src/redprl/judgment.fun
@@ -20,7 +20,7 @@ struct
   in
     val rec sort =
       fn (I, H) >> catjdg =>
-           ((List.map #2 I, Hyps.foldr (fn (_, jdg, r) => CJ.synthesis jdg :: r) [] H), 
+           ((List.map #2 I, Hyps.foldr (fn (_, jdg, r) => CJ.synthesis jdg :: r) [] H),
             CJ.synthesis catjdg)
        | MATCH (th, k, _, _, _) =>
            let
@@ -31,6 +31,6 @@ struct
            end
   end
 end
- 
+
 structure RedPrlSequent = Sequent (structure CJ = RedPrlCategoricalJudgment)
 structure RedPrlJudgment = SequentJudgment (structure S = RedPrlSequent and TermPrinter = TermPrinter)
--- a/src/redprl/lcf.sml
+++ b/src/redprl/lcf.sml
@@ -33,7 +33,7 @@ struct
             val jdg' = J.subst env jdg
             val env' = Metavar.Ctx.insert env x (LcfLanguage.var x' (J.sort jdg'))
           in
-            {doc = Fpp.seq [doc, prettyGoal (x, jdg), Fpp.hardLine],
+            {doc = Fpp.seq [doc, prettyGoal (x, jdg), Fpp.newline],
              env = env',
              idx = idx + 1}
           end)
--- a/src/redprl/lcf.sml
+++ b/src/redprl/lcf.sml
@@ -13,14 +13,14 @@ struct
 
   (* TODO: clean up all this stuff with vsep *)
 
+  fun @@ (f, x) = f x
+  infixr 0 @@
+
   fun prettyGoal (x, jdg) =
-    Fpp.seq
-      [Fpp.text "Goal",
-       Fpp.space 1,
-       Fpp.text ".",
-       Fpp.newline,
-       Fpp.nest 2 (RedPrlSequent.pretty TermPrinter.ppTerm jdg),
-       Fpp.hardLine]
+    Fpp.nest 2 @@ 
+      Fpp.vsep 
+        [Fpp.seq [Fpp.hsep [Fpp.text "Goal", Fpp.text (Metavar.toString x)], Fpp.text "."],
+         RedPrlSequent.pretty TermPrinter.ppTerm jdg]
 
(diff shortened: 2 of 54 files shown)


favonia:
@jonsterling I believe the capitalization in our comments is wrong.

favonia:
@jonsterling I believe the capitalization in your comments is wrong.

jonsterling:
I think since there is no urgency, I'll try and get a bit more of the task done in this PR. In particular, getting everything to use the new pretty printer, and deferring rendering to the outer edges.

jonsterling:
What would be sweet is to extend the pretty printing monad to have an environment of names to use for printing... This would enable a much more efficient implementation of goal printing.

But I won't do this now.

favonia:
@jonsterling Are you talking about something like `alpha` in tactics/rules?

jonsterling:
@favonia Kind of, but what I have in mind would be a mapping from metavariables to names.

favonia:
Will adding `COM` into `Syntax` now create new conflicts? (I have tons of other things to do. ;-))

jonsterling:
@favonia I would not worry about it; if you add `COM`, that's fine, and I am happy to deal with any merge conflicts on my end, since I'm the one blowing everything up 😄

jonsterling:
I'm much happier with the new concrete notation for the CHiTT term language. Now, what remains is to fix all the reduce/reduce and shift/reduce errors in the grammar.

jonsterling:
I'm mostly happy with this PR; the remaining task that I want to complete prior to merging is to get rid of the "fragments" stuff in `REDPRL_ERROR` and just use the pretty printer type for that.

src/redprl/judgment.fun
```diff
@@ -1,6 +1,6 @@
 functor SequentJudgment
   (structure S : SEQUENT where type 'a CJ.Tm.O.Ar.Vl.Sp.t = 'a list and type CJ.Tm.Sym.t = Sym.t and type CJ.Tm.Var.t = Sym.t
-   structure TermPrinter : SHOW where type t = S.CJ.Tm.abt) : LCF_JUDGMENT  =
+   structure TermPrinter : sig type t = S.CJ.Tm.abt val ppTerm : t -> Fpp.doc end) : LCF_JUDGMENT  =
```
favonia: Should we give `sig type t = S.CJ.Tm.abt val ppTerm : t -> Fpp.doc end` a name?

src/redprl/judgment.fun
```diff
@@ -12,7 +12,7 @@ struct
 
   val subst = S.map o Tm.substMetaenv
   val eq = S.eq
-  val toString = S.toString TermPrinter.toString
+  val toString = FppRenderPlainText.toString o FinalPrinter.execPP o S.pretty TermPrinter.ppTerm
```
favonia: I wonder in what cases `toString` is ever needed...

jonsterling: @favonia I consider it a bug if we ever use that from within RedPRL; the `toString` stuff is required by the LCF library, in order to make it possible to debug.

src/redprl/judgment.fun
jonsterling: I think these sharing constraints are gnarly enough that we ought to instead consider "defunctorizing" this module if possible... I wonder if there were some circularities which made me do it this way?

src/redprl/lcf.sml
```diff
@@ -3,72 +3,43 @@ structure LcfLanguage = LcfAbtLanguage (RedPrlAbt)
...
+        [Fpp.seq [Fpp.hsep [Fpp.text "Goal", Fpp.text (Metavar.toString x)], Fpp.text "."],
+         RedPrlSequent.pretty TermPrinter.ppTerm jdg]
 
-  val prettyGoals : jdg eff Tl.telescope -> {doc : PP.doc, env : J.env, idx : int} =
+  val prettyGoals : jdg Tl.telescope -> {doc : Fpp.doc, env : J.env, idx : int} = 
     let
       open RedPrlAbt
     in
-      Tl.foldl
+      Tl.foldl
```
favonia: I wonder if you want to enable git's feature to kill all whitespaces automatically. 🤔

src/redprl/operator.sml
```diff
@@ -483,6 +483,8 @@ struct
              ^ "["
              ^ dirToString f dir
              ^ "]"
+       | CUST (opid, [], _) => 
+           f opid
        | CUST (opid, ps, _) =>
            f opid ^ "{" ^ paramsToString f ps ^ "}"
```
favonia: Shouldn't this be `unlessEmpty` or something?

src/redprl/lcf.sml
jonsterling: That's a great idea! I have this enabled in `vim`, but I was using a different editor. Doing it in `git` hooks would be much better. Do you know the setting to use?

src/redprl/operator.sml
jonsterling: @favonia That's defined locally in some other module, I think. I guess we could define it here, but it is probably not worth it.

test/success/derived-rule.prl
```diff
@@ -17,42 +17,42 @@ Rule MyRule : [
 // it replaces the current goal with the premises induced by the derived rule. The
 // hypothetical contexts are also unified together in the appropriate way (i.e. we 
 // do not require strict equality of contexts, but only compatibility).
-Thm MyCorollary : [ bool -> bool ] by [
+Thm MyCorollary : [ (-> bool bool) ] by [
```
favonia: I wonder why `(` and `)` are needed?

jonsterling: @favonia Because in this PR I have switched RedPRL's term language to use an s-expression style notation. The reason for this is that sexprs are possible to layout in a civilized way when pretty-printing, whereas the notation we used before was basically hopeless in this respect.

src/redprl/lcf.sml
jonsterling: I think I've got it set up now.

favonia: Try [appl.whitespace](https://git-scm.com/docs/git-config#git-config-applywhitespace) and [core.whitespace](https://git-scm.com/docs/git-config#git-config-corewhitespace).

test/success/derived-rule.prl
favonia: Oh, I was thinking that maybe the outermost parentheses can be omitted, but keeping them is indeed more s-expression-ish.

favonia:
@jonsterling As a style issue, I could have written
```
(X {_}Y {_}Z)
```
Do the spaces between binders and terms have any significant meaning?

jonsterling:
@favonia The spaces don't mean anything; it's just a matter of convention.

Now, with that said, I think that this suggests an ambiguity in the grammar, since operator heads are written like `op{params}`, so `(op {x} a)` could mean two different things. So, we should think of a good notation to discharge this ambiguity.

favonia:
@jonsterling `<params>`?

favonia:
@jonsterling ~~`<params>`?~~ Is `:` used?

jonsterling:
@favonia `:` is not used yet

favonia:
I was thinking `loop:x` or something crazy like that, but then am unsure of `hcom:r~>r' ...`.

jonsterling:
@favonia That's an interesting idea... one problem is that there can be multiple parameters (the biggest issue to deal with is user-defined operators, since we can always give a non-ambiguous notation for the built-in ones).

favonia:
Should we recycle `[` and `]` then?

jonsterling:
The problem w/ using square brackets here is that there is still an ambiguity, since those are used for object-variable bindings (as opposed to dimension bindings).

jonsterling:
Focusing for now just on the general case for custom operators, I'm thinking that maybe it would make sense to adopt the sexpr style even for the interior of an operator head. This would mean that you would write `({my-theta p1 ... pn} {u...} [x...] m ...)` for the custom operation that we write `my-theta{p1...pn}({u...}[x...].m)`.

For the built-in operations, we can just give a special concrete notation which is unambiguous.

favonia:
@jonsterling I think the problem is, given `{x y z} m` how would you know `x` is an operator not a bound parameter? By the way, if you have liberated `'`, maybe you can use that to indicate custom operators?

jonsterling:
@favonia This is unambiguous because it would occur only at the beginning of an sexpr.

Re `'`, I have no problem with liberating this, but I'd like custom operators to be a bit more seamless, so I prefer not to require a sigil. Custom operators are already distinguished lexically because they begin with an uppercase letter, whereas variables do not.

favonia:
@jonsterling To double check, so we will have special syntax like `tt` which is the same as `{tt}`. Another crazy idea: how about putting a bar or even dollar sign? Like `(Something {x} $ {_} ...)` or `(Something {x} | {_} ...)`?

favonia:
Or maybe `(Something {x} ' {_} ...)`. All of these extra notations can be omitted if the operator head takes nothing.

jonsterling:
@favonia 

> To double check...

Yes, but it is up to us as to whether we will allow that; it might make sense to disallow such cases when there are no parameters...

> Another crazy idea:

I kind of like this idea!

jonsterling:
I played around a bit with the idea of using a pipe or something to disambiguate, but I'm afraid it looks too bureaucratic and too much like a hack. I think I like the braces idea better (`({op params} args)`).

jonsterling:
I think that some of the inconsistencies in the notation are going to disappear when we switch to the new logical framework. In that setting, I think it may make sense to just have a "type" of dimensions and be done with it, rather than dealing with all this bureaucracy for separating dimensions from other things. [As far as semantics are concerned, this "type" would simply be representable.]

So, my feeling is that we may tolerate some inconsistencies and confusing aspects of notation, until we are ready to switch.

jonsterling:
I'm going to change the `RedPrlError` stuff to use the new pretty printer only after merging @favonia's big refactoring PR in. That way we can avoid merge conflicts.

favonia:
@jonsterling I can work on the telescope library. My to-do list contains adding COM to the parser and then the testing suit after this PR.

jonsterling:
@favonia Sounds good. My thought about the order in which we can get things merged in is as follows (I think this is what you have already said anyway):

1. Your PR #226, which looks good.
2. This PR.
3. Concrete syntax for `com` and tests.

favonia:
@jonsterling #226 is merged. Now it is your turn.

PR #240, opened 2017-06-21
Refactor the refiner.

favonia:
@jonsterling Currently I am not working on this. I would like to hear your comments.

- [x] Rewrite the goal-appending process in `ComKit`.
- [x] Partition the refiners into more modules. Possibly one to for composition, ~~one for structural rules, one for base types, and another one for functions, pairs and identifications etc.~~ another one for type-specific modules.
- [x] Fix the wrong `HCom` and `Coe` tactics which wrongly assume the structural equality rule is invertible.
- [x] Fix the wrong `FCom` rule which does not check whether the type is a (weak) base types.

Needs more thinking:
- [ ] ~~Each base type should provides an `EqVal`, if appropriate, and that `Auto` should not dispatch on value terms. Leave the value determination to each `EqVal` rule. I also doubt the usefulness of exposing `EqTT` and `EqFF` when `EqVal` is available.~~ Needs discussion.
- [ ] ~~Use the `flipWrapper` hack to skip all `Equality.Symmetry`.~~ An ill-concieved idea.

Dream:
- [ ] Investigate why `Telescope.Internal.ConsView.out` is taking so much time.

### DIFF
--- a/src/redprl/refiner_kit.fun
+++ b/src/redprl/refiner_kit.fun
@@ -31,14 +31,30 @@ struct
       end
   end
 
+  fun abstractEvidence (I : (sym * psort) list, H) m =
+    let
+      val (us, sigmas) = ListPair.unzip I
+      val (xs, taus) = Hyps.foldr (fn (x, jdg, (xs, taus)) => (x::xs, CJ.synthesis jdg::taus)) ([],[]) H
+    in
+      Abt.checkb (Abt.\ ((us, xs), m), ((sigmas, taus), Abt.sort m))
+    end
+
+  fun #> (psi, (I, H, m)) =
+    Lcf.|> (psi, abstractEvidence (I, H) m)
+  infix #>
+
+  val trivial = Syn.into Syn.AX
+
+  fun orelse_ (t1, t2) alpha = Lcf.orelse_ (t1 alpha, t2 alpha)
+  infix orelse_
 
   fun >:+ (tel, (l : (label * 'a) list)) : 'a telescope =
     List.foldl (fn (g, t) => t >: g) tel l
   infix 5 >:+
 
+  (* this is a hack till we have a nice way to pre-compose Equality.Symmetry *)
   fun catJdgFlip (CJ.EQ_TYPE (a, b)) = CJ.EQ_TYPE (b, a)
     | catJdgFlip (CJ.EQ ((a, b), ty)) = CJ.EQ ((b, a), ty)
-
   fun catJdgFlipWrapper tactic alpha (IH >> jdg) =
     tactic alpha (IH >> catJdgFlip jdg)
 
--- a/src/redprl/refiner_kit.fun
+++ b/src/redprl/refiner_kit.fun
@@ -32,8 +32,15 @@ struct
   end
 
 
-  fun appendListOfGoals (tel, (l : (label * 'a) list)) : 'a telescope =
-    List.foldl (fn (g, l) => l >: g) tel l
+  fun >:+ (tel, (l : (label * 'a) list)) : 'a telescope =
+    List.foldl (fn (g, t) => t >: g) tel l
+  infix 5 >:+
+
+  fun catJdgFlip (CJ.EQ_TYPE (a, b)) = CJ.EQ_TYPE (b, a)
+    | catJdgFlip (CJ.EQ ((a, b), ty)) = CJ.EQ ((b, a), ty)
+
+  fun catJdgFlipWrapper tactic alpha (IH >> jdg) =
+    tactic alpha (IH >> catJdgFlip jdg)
 
   fun hypsToSpine H =
     Hyps.foldr (fn (x, jdg, r) => Abt.check (Abt.`x, CJ.synthesis jdg) :: r) [] H
--- a/src/redprl/refiner_types.fun
+++ b/src/redprl/refiner_types.fun
@@ -1,0 +1,922 @@
+(* type-specific modules *)
+functor RefinerTypeRules (Sig : MINI_SIGNATURE) =
+struct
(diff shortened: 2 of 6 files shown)


jonsterling:
Re `ConsView.out`, the implementation of the telescopes is very crappy and non-optimized. I bet we can improve it quite a bit if necessary...

jonsterling:
> Each base type should provides a EqVal, if appropriate, and that Auto should not dispatch on value terms. Leave the value determination to each EqVal rule. I also doubt the usefulness of exposing EqTT and EqFF when EqVal is available.

I disagree. Primitive PRL rules should not try to do fancy things like you are proposing. We should have one PRL rule for `tt = tt` and one for `ff = ff`; it is the job of tactics to do more clever dispatch.

favonia:
> I disagree. Primitive PRL rules should not try to do fancy things like you are proposing. We should have one PRL rule for tt = tt and one for ff = ff; it is the job of tactics to do more clever dispatch.

Are you against each type having an `EqVal` then? They can provide all the primitive rules *and* some useful tactics so that `Auto` is more modular.

jonsterling:
> I see your points. Are you against each type having an EqVal then?

Yes, I think I'm opposed to that, at least for now. My preference is for the rules of the PRL to be highly modular, and not at all to embed any decision procedures inside these rules.

jonsterling:
My preference is to write a tactic to capture any of the things that you are currently proposing to embed in the rules. If it is currently difficult or not possible to write a tactic that you need, then I can fix that.

favonia:
Hmm I feel I never used the terminology correctly, and we should probably clarify that: by "rules" are you talking about everything currently implemented in the refiner except for `AutoEq` and `Auto`? Or, is `AutoEq` in various types actually a rule? I think I am confused because I thought `AutoEq` is not a primitive rule and yet we have it for almost every type, and then concluded that it is okay to put more stuff in. If you also said "we should move `AutoEq` out of those modules" then I would have understood it better. I also like a complete separation of "rules" and "tactics", which is different from what I thought but also makes perfect sense.

BTW, to avoid confusion, I will use the following terminology:
- rule: any rule (scheme) that can be directly traced back to some rule (scheme) in the proof theory on paper.
- tactic: any other admissible theorems based on the rules.

favonia:
While we are trying to reach some agreement, I suppose you are fine with me fixing the wrong `Auto` rule which assumes too much invertibility and that `FCom.Eq` which fails to check whether it is in a base type.

src/redprl/refiner_kit.fun
```diff
@@ -31,9 +31,32 @@ struct
...
+  fun #> (psi, (I, H, m)) =
+    Lcf.|> (psi, abstractEvidence (I, H) m)
+  infix #>
+
+  val trivial = Syn.into Syn.AX
+
+  fun orelse_ (t1, t2) alpha = Lcf.orelse_ (t1 alpha, t2 alpha)
+  infix orelse_
+
+  fun >:+ (tel, (l : (label * 'a) list)) : 'a telescope =
```
favonia: @jonsterling Are you okay with this notation?

jonsterling: Looks good!

src/redprl/refiner_types.fun
```diff
@@ -0,0 +1,1013 @@
...
+        val Syn.SND m = Syn.out tm
+        val (goalTy, holeTy) = makeGoal @@ (I, H) >> CJ.SYNTH m
+        val (goalB, holeB) = makeGoal @@ MATCH (O.MONO O.DPROD, 1, holeTy, [], [Syn.into @@ Syn.FST m])
+      in
+        T.empty >: goalTy >: goalB
+          #> (I, H, holeB)
+      end
+  end
+
+  structure Match =
```
jonsterling: I think the idea of breaking up the refiner is fine, but this structure (`Match`) probably shouldn't go in this file, since it is not specific to types, but has to do with the general refinement apparatus. Maybe it could go into the main `Refiner` functor or something? Perhaps the same should be done with `Coe` below.

favonia: Sorry, this is a mistake. I did not intend to put these two structures here.

favonia:
@jonsterling There is still something I want to do, like having a SYNTH rule for strict bools, but I am not sure how to do that... I hope the breaking down makes it obvious what's lacking now.

jonsterling:
@favonia You're doing a great job! 

Re: `SYNTH` for strict bools, let's chat about what you have in mind. I think it might be a little tricky to have type synthesis for the `if/s(e1; e2; e3)` form, because this does not have an annotation of the motive, unlike the usual `if(a.A; e1; e2; e3)` form for the weak booleans.

src/redprl/refiner_kit.fun
```diff
@@ -31,9 +31,32 @@ struct
...
+    List.foldl (fn (g, t) => t >: g) tel l
+  infix 5 >:+
 
-  fun appendListOfGoals (tel, (l : (label * 'a) list)) : 'a telescope =
-    List.foldl (fn (g, l) => l >: g) tel l
+  (* this is a hack till we have a nice way to pre-compose Equality.Symmetry *)
+  fun catJdgFlip (CJ.EQ_TYPE (a, b)) = CJ.EQ_TYPE (b, a)
+    | catJdgFlip (CJ.EQ ((a, b), ty)) = CJ.EQ ((b, a), ty)
+  fun catJdgFlipWrapper tactic alpha (IH >> jdg) =
+    tactic alpha (IH >> catJdgFlip jdg)
```
jonsterling: After getting the pretty printer set up, I can make my next task to make it possible to write real composed tactics in the RedPRL source. This is not a technically difficult thing to do—it just means rearranging some modules.

PR #241, opened 2017-06-22
Clean-ups, stage 4

favonia:


src/redprl/operator.sml
```diff
@@ -100,7 +100,7 @@ struct
    * MONO: the Standard ML built-in equality properly compares the operators.
    * POLY: we have to compare the operators manually. *)
   datatype mono_operator =
-   (* axioms *)
+   (* the only element in equality *)
```
favonia: @jonsterling The NuPRL documentation did say "axiom" but after some thinking I feel this could be confusing... I don't really know what to say here.

jonsterling: Yeah, I think `axiom` is confusing too. Maybe we could make the comment say something like `the trivial realizer`.

jonsterling:
(btw, feel free to merge at any time)

PR #246, opened 2017-06-24
Update sml-telescopes.

favonia:


jonsterling:
I think one of these shas must point to the delete branch rather than master, or something like that.

favonia:
@jonsterling Right. I probably failed to reset everything, but then I discovered other issues...

favonia:
The fixing script I am using now:
```sh
git submodule update --recursive --init
git submodule foreach --recursive git fetch origin --force
git submodule foreach --recursive git checkout origin/master --force
```

PR #247, opened 2017-06-24
Another round of refiner cleanups and bugfixes.

favonia:
Future to-do:
- [ ] A good combinator or wrapper for rules that do not use `alpha`.

### DIFF
--- a/src/redprl/refiner.fun
+++ b/src/redprl/refiner.fun
@@ -1,3 +1,9 @@
+(* 2017/06/24
+ * As a note on programming style, the most important or most interesting
+ * subgoals should go first, as long as telescopes are well-formed.
+ *
+ * Rules violating this principle should be updated.
+ *)
 functor Refiner (Sig : MINI_SIGNATURE) : REFINER =
 struct
   structure Kit = RefinerKit (Sig)
--- a/src/redprl/refiner.fun
+++ b/src/redprl/refiner.fun
@@ -17,7 +17,7 @@ struct
 
   structure Hyp =
   struct
-    fun Project z alpha jdg =
+    fun Project z _ jdg =
       let
         val _ = RedPrlLog.trace "Hyp.Project"
         val (I, H) >> catjdg = jdg
@@ -34,11 +34,11 @@ struct
 
   structure TypeEquality =
   struct
-    fun Symmetry alpha jdg =
+    fun Symmetry _ jdg =
     let
       val _ = RedPrlLog.trace "Equality.Symmetry"
       val (I, H) >> CJ.EQ_TYPE (ty1, ty2) = jdg
-      val (goal, hole) = makeGoal @@ (I, H) >> CJ.EQ_TYPE (ty2, ty1)
+      val goal = makeGoal' @@ (I, H) >> CJ.EQ_TYPE (ty2, ty1)
     in
       T.empty >: goal
         #> (I, H, trivial)
@@ -47,7 +47,7 @@ struct
 
   structure Truth =
   struct
-    fun Witness tm alpha jdg =
+    fun Witness tm _ jdg =
       let
         val _ = RedPrlLog.trace "True.Witness"
         val (I, H) >> CJ.TRUE ty = jdg
@@ -62,7 +62,7 @@ struct
 
   structure Synth =
   struct
-    fun FromWfHyp z alpha jdg =
+    fun FromWfHyp z _ jdg =
       let
         val _ = RedPrlLog.trace "Synth.FromWfHyp"
         val (I, H) >> CJ.SYNTH tm = jdg
@@ -74,7 +74,7 @@ struct
           raise Fail "Did not match"
       end
 
-    fun Hyp alpha jdg =
+    fun Hyp _ jdg =
       let
         val _ = RedPrlLog.trace "Synth.Hyp"
         val (I, H) >> CJ.SYNTH tm = jdg
@@ -84,7 +84,7 @@ struct
         T.empty #> (I, H, a)
       end
 
-    fun If alpha jdg =
+    fun If _ jdg =
       let
         val _ = RedPrlLog.trace "Synth.If"
         val (I, H) >> CJ.SYNTH tm = jdg
@@ -97,7 +97,7 @@ struct
           #> (I, H, cm)
       end
 
-    fun S1Elim alpha jdg =
+    fun S1Elim _ jdg =
       let
         val _ = RedPrlLog.trace "Synth.S1Elim"
         val (I, H) >> CJ.SYNTH tm = jdg
@@ -110,7 +110,7 @@ struct
           #> (I, H, cm)
       end
 
-    fun Ap alpha jdg =
+    fun Ap _ jdg =
       let
         val _ = RedPrlLog.trace "Synth.Ap"
         val (I, H) >> CJ.SYNTH tm = jdg
@@ -124,20 +124,19 @@ struct
           #> (I, H, holeCod)
       end
 
-    fun PathAp alpha jdg =
+    fun PathAp _ jdg =
       let
         val _ = RedPrlLog.trace "Synth.PathAp"
         val (I, H) >> CJ.SYNTH tm = jdg
         val Syn.PATH_AP (m, r) = Syn.out tm
         val (goalPathTy, holePathTy) = makeGoal @@ (I, H) >> CJ.SYNTH m
         val (goalLine, holeLine) = makeGoal @@ MATCH (O.MONO O.PATH_TY, 0, holePathTy, [r], [])
-        val psi = T.empty >: goalPathTy >: goalLine
       in
         T.empty >: goalPathTy >: goalLine
           #> (I, H, holeLine)
       end
 
-    fun Fst alpha jdg =
+    fun Fst _ jdg =
       let
         val _ = RedPrlLog.trace "Synth.Fst"
         val (I, H) >> CJ.SYNTH tm = jdg
@@ -149,7 +148,7 @@ struct
           #> (I, H, holeA)
       end
 
-    fun Snd alpha jdg =
+    fun Snd _ jdg =
       let
         val _ = RedPrlLog.trace "Synth.Snd"
         val (I, H) >> CJ.SYNTH tm = jdg
@@ -164,7 +163,7 @@ struct
 
   structure Match =
   struct
-    fun MatchOperator alpha jdg =
+    fun MatchOperator _ jdg =
       let
         val _ = RedPrlLog.trace "Match.MatchOperator"
         val MATCH (th, k, tm, ps, ms) = jdg
@@ -172,7 +171,7 @@ struct
         val Abt.$ (th', args) = Abt.out tm
         val true = Abt.O.eq Sym.eq (th, th')
 
-        val (vls, _) = Abt.O.arity th
+        val _ = Abt.O.arity th (* is this needed? *)
         val (us, xs) \ arg = List.nth (args, k)
         val srho = ListPair.foldrEq (fn (u,p,rho) => Sym.Ctx.insert rho u p) Sym.Ctx.empty (us, ps)
         val vrho = ListPair.foldrEq (fn (x,m,rho) => Var.Ctx.insert rho x m) Var.Ctx.empty (xs, ms)
@@ -187,7 +186,7 @@ struct
 
   structure Equality =
   struct
-    fun Hyp alpha jdg =
+    fun Hyp _ jdg =
       let
         val _ = RedPrlLog.trace "Equality.Hyp"
         val (I, H) >> CJ.EQ ((m, n), ty) = jdg
@@ -210,7 +209,7 @@ struct
       handle Bind =>
         raise E.error [Fpp.text "Expected variable-equality sequent"]
 
-    fun Symmetry alpha jdg =
+    fun Symmetry _ jdg =
       let
         val _ = RedPrlLog.trace "Equality.Symmetry"
         val (I, H) >> CJ.EQ ((m, n), ty) = jdg
@@ -246,7 +245,7 @@ struct
         T.empty >: goalTy >:? goalTy0 >: goalCoercees #> (I, H, trivial)
       end
 
-    fun CapEqL alpha jdg =
+    fun CapEqL _ jdg =
       let
         val _ = RedPrlLog.trace "Coe.CapEq"
         val (I, H) >> CJ.EQ ((coe, other), ty) = jdg
@@ -310,7 +309,7 @@ struct
          | _ => m
     end
 
-    fun Unfold sign opid alpha jdg =
+    fun Unfold sign opid _ jdg =
       let
         val _ = RedPrlLog.trace "Computation.Unfold"
         val unfold = safeUnfold sign opid o Abt.deepMapSubterms (safeUnfold sign opid)
@@ -320,11 +319,11 @@ struct
         T.empty >: goal #> (I, H, hole)
       end
 
-    fun EqHeadExpansion sign alpha jdg =
+    fun EqHeadExpansion sign _ jdg =
       let
         val _ = RedPrlLog.trace "Computation.EqHeadExpansion"
         val (I, H) >> CJ.EQ ((m, n), ty) = jdg
-        val Abt.$ (theta, _) = Abt.out m
+        val Abt.$ _ = Abt.out m (* is this needed? *)
         val m' = Machine.unload sign (safeEval sign (Machine.load m))
         val goal = makeGoal' @@ (I, H) >> CJ.EQ ((m', n), ty)
       in
@@ -333,11 +332,11 @@ struct
       end
       handle _ => raise E.error [Fpp.text "EqHeadExpansion!"]
 
-    fun EqTypeHeadExpansion sign alpha jdg =
+    fun EqTypeHeadExpansion sign _ jdg =
       let
         val _ = RedPrlLog.trace "Computation.EqTypeHeadExpansion"
         val (I, H) >> CJ.EQ_TYPE (ty1, ty2) = jdg
-        val Abt.$ (theta, _) = Abt.out ty1
+        val Abt.$ _ = Abt.out ty1 (* is this needed? *)
         val ty1' = Machine.unload sign (safeEval sign (Machine.load ty1))
         val goal = makeGoal' @@ (I, H) >> CJ.EQ_TYPE (ty1', ty2)
         in
@@ -351,7 +350,7 @@ struct
       val _ = RedPrlLog.trace "Cut"
       val (I, H) >> catjdg' = jdg
       val z = alpha 0
-      val tau = CJ.synthesis catjdg
+      val _ = CJ.synthesis catjdg (* what is this for? it is not used. *)
       val (goal1, hole1) = makeGoal @@ (I, H @> (z, catjdg)) >> catjdg'
       val (goal2, hole2) = makeGoal @@ (I, H) >> catjdg
     in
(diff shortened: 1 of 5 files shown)


favonia:
When comparing two abt under binders, say `(u, s)` and `(v, t)`, I guess we do not have to do substitution on both sides but just rename one of them to the other, right? Of course, getting a fresh variable is free of any shadowing/capturing problem.

src/redprl/refiner_composition_kit.fun
```diff
@@ -145,16 +145,20 @@ struct
...
+         * and thus alpha-equivalence is sufficient. *)
         val goalTy0 = makeEqTypeIfDifferent (I, H) (ty0, ty)
-        val goalTy01 = makeGoal' @@ (I, H) >> CJ.EQ_TYPE (ty0, ty1)
+        val goalTy1 = makeEqTypeIfAllDifferent (I, H) (ty1, ty) [ty0]
 
         (* cap *)
-        val goalCap01 = makeGoal' @@ (I, H) >> CJ.EQ ((cap0, cap1), ty)
+        (* the cap-tube adjacency premise guarantees that [cap0] is in [ty],
+         * and thus alpha-equivalence is sufficient. *)
+        val goalCap01 = makeEqIfDifferent (I, H) ((cap0, cap1), ty)
```
favonia: Actually this is wrong. Sorry.

src/redprl/refiner.fun
```diff
@@ -164,15 +163,15 @@ struct
...
+    fun MatchOperator _ jdg =
       let
         val _ = RedPrlLog.trace "Match.MatchOperator"
         val MATCH (th, k, tm, ps, ms) = jdg
 
         val Abt.$ (th', args) = Abt.out tm
         val true = Abt.O.eq Sym.eq (th, th')
 
-        val (vls, _) = Abt.O.arity th
+        val _ = Abt.O.arity th (* is this needed? *)
```
favonia: is this needed?

src/redprl/refiner.fun
```diff
@@ -320,11 +319,11 @@ struct
         T.empty >: goal #> (I, H, hole)
       end
 
-    fun EqHeadExpansion sign alpha jdg =
+    fun EqHeadExpansion sign _ jdg =
       let
         val _ = RedPrlLog.trace "Computation.EqHeadExpansion"
         val (I, H) >> CJ.EQ ((m, n), ty) = jdg
-        val Abt.$ (theta, _) = Abt.out m
+        val Abt.$ _ = Abt.out m (* is this needed? *)
```
favonia: is this needed?

src/redprl/refiner.fun
```diff
@@ -351,7 +350,7 @@ struct
       val _ = RedPrlLog.trace "Cut"
       val (I, H) >> catjdg' = jdg
       val z = alpha 0
-      val tau = CJ.synthesis catjdg
+      val _ = CJ.synthesis catjdg (* what is this for? it is not used. *)
```
favonia: is this needed?

src/redprl/refiner.fun
```diff
@@ -333,11 +332,11 @@ struct
       end
       handle _ => raise E.error [Fpp.text "EqHeadExpansion!"]
 
-    fun EqTypeHeadExpansion sign alpha jdg =
+    fun EqTypeHeadExpansion sign _ jdg =
       let
         val _ = RedPrlLog.trace "Computation.EqTypeHeadExpansion"
         val (I, H) >> CJ.EQ_TYPE (ty1, ty2) = jdg
-        val Abt.$ (theta, _) = Abt.out ty1
+        val Abt.$ _ = Abt.out ty1 (* is this needed? *)
```
favonia: is this needed?

src/redprl/refiner.fun
```diff
@@ -383,11 +382,6 @@ struct
      | UPDATE of hyp * catjdg
      | INSERT of hyp * catjdg
 
-    val diffToString =
-      fn DELETE x => "DELETE " ^ Sym.toString x
-       | UPDATE (x,_) => "UPDATE " ^ Sym.toString x
-       | INSERT (x,_) => "INSERT " ^ Sym.toString x
-
```
favonia: Maybe this was for debugging? But it does not seem to be used anywhere.

jonsterling: yeah, it was for debugging.

jonsterling:
@favonia To answer your question, I think it is not necessary since 'unbinding' will ensure that each of those variables is globally fresh. I tend to pick a common fresh variable anyway at times like that because it makes the code a bit more symmetric, but probably not necessary.

jonsterling:
Regarding the seemingly dead code you pointed out, I think you can use the following rule of thumb:

1. If the behavior cannot be changed by removing the line, then feel free to remove the line. This includes things like `val _ = ...` where `...` is some code that does not have an observable side effect.

2. If the behavior could be changed by removing the line, probably don't remove it for now; one example would be `val _ $ _ = out t`: lines like this probably exist to ensure a rule only applies when `t` is an application-form. But I haven't looked carefully at each example...

src/redprl/refiner.fun
```diff
@@ -17,7 +17,7 @@ struct
 
   structure Hyp =
   struct
-    fun Project z alpha jdg =
+    fun Project z _ jdg =
```
jonsterling: to be honest, I'm not sure how I feel about hiding all these "unused variables". I know it's nice for when you have the warnings turned on, but it also makes it a lot harder to understand the code. It is helpful if when looking at the code, a programmer can see what this argument is supposed to do—and if they need to edit the rule to use the namestore, it will be clear how to do so without having to search around.

So, I kind of feel that we should leave things like this in place. Unused variables is not always a bug.

src/redprl/refiner.fun
favonia: Should we keep them though?

src/redprl/refiner.fun
favonia: The problem I was trying to solve is that the log is too long. How about I do what I did previously---writing `_(*alpha*)`?

jonsterling: @favonia I don't want to solve problems with the MLton tooling by changing my code...

We should probably turn off the unused variable warning, because they appear to be negatively useful.

jonsterling: I'm also not a big fan of `_(*alpha*)`, as it seems very bureaucratic and is kind of confusing to look at; moreover, it makes the code more sensitive to small changes. Having to comment out a variable binding non-locally if I remove a line that is the only use of some variable is kind of odious to me.

src/redprl/refiner.fun
jonsterling: I would keep it; it's kind of annoying to have to rewrite the same code again and again when I need to debug something.

Ultimately I think that this falls into the same category as our other discussion thread: not every bit of unused code is a bug. Sometimes it is the right thing to have unused code or unused variables.

src/redprl/refiner.fun
favonia: @jonsterling The warnings are very useful because I actually fixed bugs by reviewing all the unused variables. For example, this time, one of the goals in S1-elim was created but forgotten in the returned expression, which was detected by the warnings. I also feel the ugliness of the code, but since bugs were found I feel the price is small.

src/redprl/refiner.fun
favonia: If this is written with the pretty printer I would be happier.

src/redprl/refiner.fun
jonsterling: If you feel very strongly about it, then `_` is at least preferable to `_(*alpha*)`.

I will just say that I understand what you are saying, but don't agree in the general case. For instance, in some places I define a dummy structure that ascribes a signature, to check if some other structure conforms to that signature (as a sanity check):

```sml
structure _MyStruct : SIG = MyStruct
```

jonsterling: This is helpful because it lets me ensure that `MyStruct` implements `SIG`, but without making me ascribe it directly (which would destroy things that I need to be visible externally, such as datatype declarations).

favonia: Maybe I can add that I did not blindly delete all unused code---you can see the build log. The `alpha` is an annoying case because it clutters the log when I tried to carefully review every case, and I have not came up with a good combinator or style to both problems. Could we have another wrapper for the "pure" rules that do not use the name generator? The `makeGoal'` etc were created to solve readability and warnings at the same time.

favonia:
> If the behavior could be changed by removing the line, probably don't remove it for now; one example would be val _ $ _ = out t: lines like this probably exist to ensure a rule only applies when t is an application-form. But I haven't looked carefully at each example...

Indeed, and I would like to leave a comment to indicate my suspicion.

favonia:
@jonsterling I want to say that even the lines like `val _ = ...` can have side effects. Maybe it is checking something. I keep all the lines when I was not sure what they do.

src/redprl/refiner.fun
jonsterling: @favonia That's a good idea!

For now, let's replace `alpha` with `_` and we can think of helper functions in a future PR. Ultimately, it may be helpful to have some kind of wrapper around rules anyway.

favonia:
@jonsterling I have questions about the importance principle you proposed---for the elimination rules, do you think the motive is more important/interesting or the "actual" proofs?

jonsterling:
@favonia Re: a comment, good idea; definitely feel free to add a comment with the question, and then at some point, either a comment should be added to explain it or that line should be deleted. (Or, it could be replaced with an assertion that made semantic sense.)

jonsterling:
@favonia I'm not sure I follow... In sequent formulation, the motive is automatic and is not a subgoal: when eliminating hypothesis `z` with goal `C[z]`, the motive is just `C`. Or am I misunderstanding?

I think that it might be easier to go through these questions about premise ordering in person when you get back from Big Proof... Of course, I'm happy to discuss now or answer any questions, but these are kind of intricate issues so it is easy to miscommunicate over text. 

One thing I would note is that in the equality rules, it is possible that the principle which I outlined (of putting the most important premises first) is not fully implemented; there may be cases where the order is not what it should be, in which case you should feel free to change it.

favonia:
@jonsterling Along these lines I gathered that these principles should be written down somewhere. Even just a comment in the header. Often I am not aiming at a principle different from yours, but somehow inferred a different principle from the current inconsistent state. I then changed everything for consistency.

I basically want to hear your opinion about this ordering, but yes we can discuss it later. The closures in my opinion are more important and we should get the machine working (almost) without forcing.
```sml
val goalCz = makeGoal' @@ (I, H @> (z, CJ.TRUE S1)) >> CJ.EQ_TYPE (c0z, c1z)
val goalM = makeGoal' @@ (I, H) >> CJ.EQ ((m0, m1), S1)
val goalCM = makeEqTypeIfDifferent (I, H) (c0m0, c)
val goalB = makeGoal' @@ (I, H) >> CJ.EQ ((b0, b1), cbase)
val goalL = makeGoal' @@ (I, H) >> CJ.EQ ((l0u, l1u), cloop)
val goalL00 = makeEqIfAllDifferent (I, H) ((l00, b0), cbase) [b1]
val goalL01 = makeEqIfAllDifferent (I, H) ((l01, b0), cbase) [l00, b1]

val psi = T.empty >: goalCz >: goalM >:? goalCM >: goalB >: goalL >:? goalL00 >:? goalL01
```

jonsterling:
> Along these lines I gathered that these principles should be written down somewhere. Even just a comment in the header. Often I am not aiming at a principle different from yours, but somehow inferred a different principle from the current inconsistent state. I then changed everything for consistency.

Excellent point; these things should be documented, because as you say, the guiding principles are not really consistently implemented as is. Documenting stuff like that is an important step in the transition from a system with a single implementer to a collaborative effort.

I also agree about replacing the machine being a critical task.

Briefly glancing at the subgoals that you wrote down, I think it looks sensible.

src/redprl/refiner.fun
favonia: To brainstorm the ideas, I have tried `ignoreAlpha @@ fn jdg =>...` but SML wants parentheses around the `fn` expression, which in my opinion is ugly. An also ugly compromise is to define `EqType'` and then `fun EqType jdg = ignoreAlpha EqType' jdg`.

src/redprl/refiner.fun
```diff
@@ -1,3 +1,9 @@
+(* 2017/06/24
+ * As a note on programming style, the most important or most interesting
+ * subgoals should go first, as long as telescopes are well-formed.
+ *
+ * Rules violating this principle should be updated.
+ *)
```
favonia: @jonsterling Is this what you were trying to say?

favonia:
@jonsterling Questions: If a `TRUE A` goal is satisfied and its hole is `M`, can I assume `A` is a type and `M` must be in `A`? On the other hand, I guess I cannot assume anything about the result of `SYNTH`, right? What can I know from a `MATCH` goal when it is completed? These pre- and post-conditions should probably be documented in a separate file, I think. Double checking the `EQ` and `EQ_TYPE` judgments:

- An `EQ ((m,n),ty)` in the context implies `TYPE ty`.
- An `EQ ((m,n),ty)` goal assumes nothing, and will imply `TYPE ty` when satisfied.
- An `EQ_TYPE` goal assumes nothing.

I wish to know similar information about `SYNTH`, `TRUE` and `MATCH`.

favonia:
Waiting for #248.

src/redprl/refiner.fun
jonsterling: Good!

src/redprl/refiner.fun
jonsterling: Yeah... I too have thought about something like that; I think that maybe we can do something better later on—for now, I think we can leave the `_` in, but ultimately there are some alternatives that may turn out to be preferable in the end. Let's keep thinking about this, but put it on the back-burner.

favonia:
@jonsterling I don't quite understand the following part in `ApEq`:
```
val (goalDFun0, holeDFun0) = makeGoal @@ (I, H) >> CJ.SYNTH m0
val (goalDFun1, holeDFun1) = makeGoal @@ (I, H) >> CJ.SYNTH m1
val (goalDFunEq, _) = makeGoal @@ (I, H) >> CJ.EQ_TYPE (holeDFun0, holeDFun1)
```
Why do we want to synthesize the type twice instead of type checking?

jonsterling:
@favonia An alternative would be to synthesize the left hand side and check the right hand side, but that is less symmetric and doesn't make much sense for structural equality rules.

src/redprl/refiner_types.fun
```diff
@@ -565,13 +554,13 @@ struct
...
         val Syn.SND m1 = Syn.out snd1
 
-        val (goalTy, holeTy) = makeGoal @@ (I, H) >> CJ.SYNTH m0
-        val (goalTyB, holeTyB) = makeGoal @@ MATCH (O.MONO O.DPROD, 1, holeTy, [], [Syn.into @@ Syn.FST m0])
-        val (goalEq, _) = makeGoal @@ (I, H) >> CJ.EQ ((m0, m1), holeTy)
-        val (goalEqTy, _) = makeGoal @@ (I, H) >> CJ.EQ_TYPE (holeTyB, ty)
+        val (goalTy, holeTy) = makeSynth (I, H) m0
+        val (goalTyB, holeTyB) = makeMatch (O.MONO O.DPROD, 1, holeTy, [], [Syn.into @@ Syn.FST m0])
+        val goalEq = makeEqIfDifferent (I, H) ((m0, m1), holeTy) (* m0 well-typed *)
+        val goalEqTy = makeEqType (I, H) (holeTyB, ty) (* holeTyB type? *)
```
favonia: I wonder if we could assume `TYPE holeTyB` here?

src/redprl/refiner_types.fun
```diff
@@ -708,30 +693,29 @@ struct
...
         val Syn.PATH_AP (m1, r1) = Syn.out ap1
         val () = Assert.paramEq "Path.ApEq" (r0, r1)
-        val (goalSynth, holeSynth) = makeGoal @@ (I, H) >> CJ.SYNTH m0
-        val (goalMem, _) = makeGoal @@ (I, H) >> CJ.EQ ((m0, m1), holeSynth)
-        val (goalLine, holeLine) = makeGoal @@ MATCH (O.MONO O.PATH_TY, 0, holeSynth, [r0], [])
-        val (goalTy, _) = makeGoal @@ (I, H) >> CJ.EQ_TYPE (ty, holeLine)
+        val (goalSynth, holeSynth) = makeSynth (I, H) m0
+        val goalMem = makeEqIfDifferent (I, H) ((m0, m1), holeSynth) (* m0 well-typed *)
+        val (goalLine, holeLine) = makeMatch (O.MONO O.PATH_TY, 0, holeSynth, [r0], [])
+        val goalTy = makeEqType (I, H) (holeLine, ty) (* holeLine type? *)
```
favonia: I wonder if we could assume `TYPE holeLine` here?

src/redprl/refiner_composition_kit.fun
```diff
@@ -285,25 +310,34 @@ struct
...
         val (I, H) >> CJ.EQ ((com, other), ty) = jdg
         val Syn.COM {dir=(r, r'), ty=(u0,ty0), cap, tubes} = Syn.out com
-        val (eq, (u, tube)) = Option.valOf (List.find (fn (eq, _) => P.eq Sym.eq eq) tubes)
+
+        (* equations. they must be tautological because one of them is true. *)
+        val (_, (u, tube)) = Option.valOf (List.find (fn (eq, _) => P.eq Sym.eq eq) tubes)
 
         (* type *)
-        val goalTy0 = makeGoal' @@ (I, H) >> CJ.EQ_TYPE (substSymbol (r', u0) ty0, ty)
+        val goalTy = makeType (I @ [(u0,P.DIM)], H) ty0
```
favonia: I believe `@ [(u0,P.DIM)` was forgotten.

src/redprl/refiner_types.fun
jonsterling: Yes, I think that you can assume that `holeTyB` is a type, if I am reading these premises correctly, since `SYNTH` is meant to guarantee that its output is a type; and if `(x : A) -> B[x]` is a type, then so is `B[x]`.

src/redprl/refiner_composition_kit.fun
jonsterling: It's a little hard for me to follow the diff, so I'll take your word for it.

src/redprl/refiner_types.fun
jonsterling: Yes, I think it should be possible to assume this as well, for the same reason as the other one.

src/redprl/refiner_types.fun
```diff
@@ -343,14 +339,15 @@ struct
...
-        val (goalL, _) = makeGoal @@ (I, H) >> CJ.EQ ((l0u, l1u), cloop)
-        val (goalL00, _) = makeGoal @@ (I, H) >> CJ.EQ ((l00, b0), cbase)
-        val (goalL01, _) = makeGoal @@ (I, H) >> CJ.EQ ((l01, b0), cbase)
+        val goalCz = makeEqType (I, H @> (z, CJ.TRUE S1)) (c0z, c1z)
+        val goalM = makeEq (I, H) ((m0, m1), S1)
+        val goalCM = makeEqTypeIfDifferent (I, H) (c0m0, c) (* c0m0 type *)
+        val goalB = makeEq (I, H) ((b0, b1), cbase)
+        val goalL = makeEq (I @ [(u,P.DIM)], H) ((l0u, l1u), cloop)
+        val goalL00 = makeEqIfAllDifferent (I, H) ((l00, b0), cbase) [b1]
+        val goalL01 = makeEqIfAllDifferent (I, H) ((l01, b0), cbase) [l00, b1]
```
favonia: @jonsterling Sorry I meant this line:
```sml
val goalL = makeEq (I @ [(u,P.DIM)], H) ((l0u, l1u), cloop)
```

favonia: I added the `@ [(u,P.DIM)]`.

jonsterling: @favonia Great! Thanks, very good catch 😄

PR #250, opened 2017-06-26
Documentation on categorical judgments. Close #248.

### LINKED ISSUE #248: Documentation on the spec of refining rules
Specification of the telescopes and the categorical judgments that every rule needs to follow:

- Telescopes
    - An `EQ ((m,n),ty)` in the context implies `TYPE ty`.

- Goals to prove:
    - All new goals assume nothing. For example, while proving `EQ ((m,n),ty)` it is not known that `TYPE ty`.

- Satisfied goals:
    - A satisfied `EQ ((m,n),ty)` goal implies `TYPE ty`.
    - A satisfied `TRUE ty` goal with the hole `m` implies `TYPE ty` and `MEM (m, ty)`.
    - A satisfied `SYNTH ty` with the hole `m` goal implies `TYPE ty` and `MEM (m, ty)`.

By the way, a `SYNTH` for `if/s` can be done by `if/s` itself, but the generated type may not be what the user want.

(This issue will be closed once the documentation is completed.)

favonia:


favonia:
@jonsterling We need a label for documentation.

jonsterling:
@favonia 

 > We need a label for documentation.

done

Issue #256, opened 2017-06-27
Unify RedPRL theorems & definitions


jonsterling:
I want to add categorical judgments for each sort; then, we can get rid of this idea of having "arguments" & arities as a primitive property of signature declarations, and capture this in the `spec` thing.

This may also lead the way to incremental/refinement-style construction of raw terms, which would be pretty sweet.

PR #257, opened 2017-06-27
starting to address #256

jonsterling:


jonsterling:
I'm back-peddling away from derived rules and allowing users to instantiate lemmas whose statement is a non-empty sequent. That stuff is very iffy and I don't think it ever worked correctly.

In the short-term this means that there will be some things that we cannot test or prove (see the tests that I have deleted)... But as we start to think about pretypes, we can recover this stuff in a simpler way.

PR #258, opened 2017-06-28
some experiments with pretypes

jonsterling:
Highly experimental. Since everything in RedPRL is currently Kan, this shouldn't break anything semantically. But this might give us the opportunity to start experimenting with non-Kan types ("cubical pretypes")—we would have subgoals for demonstrating that a type is Kan in (e.g.) the elimination rules for weak bool and S1.

### DIFF
--- a/src/redprl/categorical_judgment.sml
+++ b/src/redprl/categorical_judgment.sml
@@ -1,9 +1,11 @@
 structure RedPrlCategoricalJudgment :
 sig
+  datatype type_kind = KAN_TYPE | PRETYPE
+
   datatype 'a redprl_jdg =
      EQ of ('a * 'a) * 'a
    | TRUE of 'a
-   | EQ_TYPE of 'a * 'a
+   | EQ_TYPE of type_kind * 'a * 'a
    | SYNTH of 'a
    | TERM of RedPrlSort.t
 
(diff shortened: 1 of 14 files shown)


favonia:
There's also something I want to experiment: coe types, hcom types, pretypes and Kan types (coe+hcom).

favonia:
Ah, and also "strict" Kan types, with a notable property that its equality is Kan!

jonsterling:
@favonia That's a very interesting one!

favonia:
Did you define (meet-)semilattice signature somewhere? These classes form a semilattice, and I believe we can implement everything at once. Also I want `EQ`, `TRUE` and `SYNTH` to carry the information around so that various optimizations would work.

jonsterling:
Re: carrying around this information on `EQ`, `TRUE` and `SYNTH`, I don't think that will work out for a number of reasons that we can discuss.

My suggestion is that we try to do the simplest possible version of this idea that can work, and then we will improve it *incrementally*, and add sensible optimizations as we go.

favonia:
@jonsterling The current code is already exploiting the invariant that `EQ ((m,n),ty)` actually means `EQ ((m,n),ty)` and `KAN ty`, etc, and so I think carrying this information actually leads to the minimal/simplest change.

jonsterling:
@favonia I think it will be better to *start* with a situation where all the usual judgments only guarantee pretypehood. It is easier to add to this than it is to subtract...

I think that it ultimately may make sense to adopt an idea like you are saying, where we have `A true_{Kan/DiscreetlyKan/Pretype}` judgment whatever—and then the semantics of this judgment are that if it can be derived, than `A` is itself Kan/DiscreetlyKan/Pretype or whatever.

But I would like to start with something simpler and work our way up to this.

favonia:
Should I start a new PR for the super fancy implementation?

jonsterling:
@favonia Yes, feel free. We should close this PR.

favonia:
I believe this branch does not have useful commits left unmerged. @jonsterling Agree?

jonsterling:
Agreed, I've deleted the branch.

Issue #230, opened 2017-06-19
Summary of discussion on machine


jonsterling:
After @favonia's productive chat with Bob, @favonia and I have discussed the status of the machine, and it seems largely correct except for one very bad bug which we need to fix. In the course of fixing it, we hope that we can achieve a more efficient and well-structured machine simultaneously.

The issue is that the machine seems to be correct for code which does not compute under dimension binders, but it is definitely not correct for computation which does go underneath a binder. The problem is in the `unload` phase, which is invoked when a **neutral** term is in focus: the focused term gets plugged into the nearest stack frame, and the environment of the focused term gets wrapped around the resulting term, which becomes the focus. This is correct behavior for computation that does not go underneath a binder, but it is not correct when we are underneath a binder: since by plugging the term into the binder, we lose the identity of a variable in the context, since the variable becomes bound.

***Note***: The `unload` phase *only* can occur when we are computing with open terms. With closed terms, it cannot occur. The `up` phase (which is the ordinary way in which a *value* is pushed up the stack) seems to have no such flaw.

The way to fix this is to use a *second-order* environment, so that we can perform a substitution which can see the bound variable. In older literature, it is common to see this as a kind of "capture-inducing substitution", but as far as we are concerned, bound variables don't have names, so it makes more sense to think of it as the substitution of a binder for a second-order variable. (I believe that this perspective comes from either Hamana or Fiore, though I can't remember.)

/cc @RobertHarper @cangiuli

RobertHarper:
cool!

bob

(c) Robert Harper  All Rights Reserved.

favonia:
Related issues: #221, https://github.com/RedPRL/sml-typed-abts/issues/86. I guess we can close all of these once the more powerful environments are implemented.

jonsterling:
@favonia Please don't feel obligated to respond to this soon, since I know you're quite busy at Big Proof! Let me just write down something I was thinking...

I was thinking a bit more about the case of evaluating under a dimension binder (e.g. the commutative case for `coe`). Even given the current weird behavior of `unload`, in which the closure's environment gets re-wrapped around the innermost stack frame, I don't think that this can lead to incorrect behavior.

[...]

PR #262, opened 2017-06-30
New machine

### LINKED ISSUE #207: `safeEval` is too conservative and inefficient
@cangiuli @jonsterling `safeEval` seems way too conservative and inefficient to me. Ideas:

It seems we only have to check whether it is hcom and coe in base types. All other rules are either safe (independent of dim) or safety-preserving (which should already be an operator on the stack, except some hcom/coe holding non-base tags until issue #192 is resolved). Therefore, we should not force the delayed substitution unless it is critical for us to determine safety, and should bravely step in all other cases.

Forgot to say that, even in base types, the cap rule should fire whenever possible.

Related discussion is in #175.

jonsterling:
Starting to implement the new abstract machine.

To resolve #207, #230


favonia:
Since we are creating a new machine, could we pass around some user-defined local states so that we can potentially implement the fancy algorithms described in #207?

favonia:
Or if we want to specialize this machine, then I would like to pass around the set of "locally protected" (bound) parameter variables (or "declarations").

src/redprl/new_machine.sml
```diff
@@ -0,0 +1,306 @@
...
+  fun stepView sign stability = 
+    fn `x <: (mrho, rho, psi) || stk =>
+       (Var.Ctx.lookup rho x || stk
+        handle Var.Ctx.Absent => raise Neutral (VAR x))
+     | meta $# (ps, ms) <: env || stk =>
+       let
+         val (mrho, rho, psi) = env
+         val (us, xs) \ (m <: (mrho', rho', psi')) = Metavar.Ctx.lookup mrho meta handle Metavar.Ctx.Absent => raise Neutral (METAVAR meta)
+         val rho'' = ListPair.foldl (fn (x, n, rho) => Var.Ctx.insert rho x (n <: env)) rho' (xs, ms)
+         val psi'' = ListPair.foldl (fn (u, (r, _), psi) => Sym.Ctx.insert psi u (readParam psi r)) psi' (us, ps)
```
favonia: So, I am worried that `rho'` and `mrho'` have to be forced because both of them can refer to `us` or `xs`.

favonia: PS: My other ideas, if working, will only need to force `ms`, which might be better because it seems all the `xs` or `ms` generated by the machine would be empty. That is, it seems the only non-empty `xs` (or `ms`) are given by the users.

jonsterling:
Re: locally bound variables, yes, I had a thought that we could have this be part of the state, and then it would be taken into account when determining if we are permitted to compare two dimensions for apartness.

src/redprl/new_machine.sml
jonsterling: @favonia Could you come up with an example of where this is a problem?

Suppose `rho'` mentions some `x` (in the sense that `rho'` takes some variable `y` to `N(x)`). Then if we encounter the variable `y`, it will be changed to `x` by the variable rule; then, by the variable rule again, `x` will be replaced again, because of the substitution `ms/xs`.

Is there a problem that I'm missing?

jonsterling: @favonia Ahh, I see what would happen. Yes, I think this is a problem.

jonsterling: @favonia I think what we would need to do to fix this is to add these further substitutions to the leaves of the environment tree. Maybe this could be made efficient through some laziness...

What do you think?

favonia:
@jonsterling Yes, please include that set. Otherwise the new machine could be more conservative under binders than the old machine.

favonia:
@jonsterling Could you cherry-pick 2c340bd1dffbb41c62d9c71209964a294bf45ec0 or other changes to the doc to the master branch?

jonsterling:
@favonia Oh, my mistake, I was pushing to the wrong branch... I'll fix it.

favonia:
I was wrong---the machine can actually generate non-empty `xs` (or `ms`) by itself. I don't know how to (nicely) avoid all the forcing then. :disappointed:

jonsterling:
@favonia Should I start updating this PR to match the rules in the PDF?

favonia:
@jonsterling I guess it's stable now, though there are probably bugs...

favonia:
@jonsterling FYI: I learned from @gebner that Lean3 is not doing any kind of lazy substitution but just caching and keeping upper bounds of de Bruijn indexes.

jonsterling:
@favonia I would prefer not to merge this PR yet for the following two reasons:

1. It doesn't implement the strategy described in the PDF / note
2. It's not yet done.

So, I wanted to check with you—do you think that we are ready to begin implementing the algorithm described in the note? If so, I can start doing that within the next few days.

favonia:
@jonsterling I have a feeling that maybe you thought one of my comment to #271 was to this PR? Anyways, I think the note is ready modulo bugs.

jonsterling:
@favonia I was responding to this comment: https://github.com/RedPRL/sml-redprl/pull/262#issuecomment-314350143

Anyway, thanks for the info! We can start implementing the new machine then.

jonsterling:
Tomorrow, I begin re-implementing along the lines described in the note. 🎊

jonsterling:
@favonia OK! Now (tomorrow morning) I will begin implementing the simpler machine, and consign the gnarly stuff to the ash-heap of history.

favonia:
I will miss the closures on fire. Maybe they will not burn unless we force them.

favonia:
@jonsterling I wonder if it makes sense to keep a set of `syms` at each frame so that you do not have to remove a symbol from the set when going out of a binder?

jonsterling:
@favonia That's an interesting idea, and I tried to do it, but I couldn't figure out how to structure it properly; maybe we could try it again after this pull request gets merged?

jonsterling:
I want to merge this now so that the machine doesn't block any other tasks. It seems to work!

This PR includes the replacement of the ABT library as well.

Issue #56, opened 2016-02-14
Integers, less than, natural numbers, and truncation levels


jonsterling:

favonia:
Should we implement natural numbers at the same time?

jonsterling:
@favonia That sounds reasonable! We should think about how we actually want to represent the integers—should it be primitive, or should it use the naturals?

And we should confer with Carlo to decide the specific form of the naturals... My preference would be *strict* naturals / naturals where the compositions compute to the cap or something like that, if possible.

favonia:
Indeed. Firstly, natural numbers and truncation levels should just be subtypes of integers. Secondly, I want both the binary and unary representations, but have to think more about what to do. The motivation is to extract efficient programs that will be appreciated by most computer scientists. By the way, I believe whether it is strict or not is irrelevant to the representation---the difference only lies in the dynamics of `hcom` and we can add `snat` just like `sbool`.

jonsterling:
@favonia thanks, I think what you say makes sense.

cangiuli:
By the way, on paper I was planning on renaming `bool` to `wbool` and `sbool` to `bool`, so that future types can be called `unit`, `nat`, etc. rather than `sunit` and `snat`.

jonsterling:
👍

favonia:
@jonsterling @cangiuli After some internal struggling I now prefer a new parameter sort for integers, along with a new 0-arity operator taking an integer directly from Standard ML's `Intinf` or `Z` from the math textbook. The `nat` PER relates all the non-negative integers. The induction principle for `int` should have three cases, and the principle for `nat` is just a specialized version omitting the negative branch. We could have different symbols for both induction principles, though. The successor `S(n)` is just `n + 1`, etc. Realizers using these induction principles would be terribly inefficient, but later on we can provide a primitive less-than that connects to "real' programming languages as in NuPRL, and one can prove theorems about realizers using the primitive less-than constructs.

How is the above plan?

@jonsterling When shall we rename `BOOL` to `WOOL`... I mean `W_BOOL`?

favonia:
(I changed the above comment. See the updated version on GitHub.)

favonia:
(@jonsterling I changed the above comment. See the updated version on GitHub.)

jonsterling:
@favonia I like that idea, it sounds good for programming(!). I can't remember what @cangiuli said about this—is it ok if we just bung this into redprl? 

[...]

PR #265, opened 2017-07-01
Add integers, stage 0.

favonia:
Fulfilling #56.

### DIFF
--- a/src/redprl/machine.sml
+++ b/src/redprl/machine.sml
@@ -121,6 +121,11 @@ struct
            @@ (O.MONO O.S_IF `$ [([],[]) \ S.HOLE, ([],[]) \ S.% t, ([],[]) \ S.% f], b)
            <: env
 
+     | O.MONO O.INT `$ _ <: _ => S.VAL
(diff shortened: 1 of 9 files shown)


favonia:
@jonsterling I think we can try to merge this first without the ability to prove anything interesting. So the dependency is that interesting operators must wait for the new machine, and the renaming of wool will be in git-conflict with integers. I suggest this PR -> machine -> wool -> more on integers or other stuff.

jonsterling:
@favonia I think this looks like a reasonable start.

Question: is there any reason to make numerals a parameter sort? It might be simpler to define the operator like this:

```
data mono_operator = ... | NUMBER of int
```

favonia:
@jonsterling Do you want to reason about integers abstractly? Then you probably want a number sort. See the attached testing cases. Otherwise yeah we do not need the `ParamTerm`.

favonia:
@jonsterling I was talking about this:
```
Thm Abstract{i:num} : [
  (number i) in int
] by [
  auto
].
```

jonsterling:
Ah, I see. Well, I'm fine with having it be the way you wrote it.

PR #266, opened 2017-07-01
renaming bool, wbool/sbool, if/s, etc.

jonsterling:
To avoid ambiguity, the recursor for wbool is now called wbool-rec, and
the recursor for strict bool is just if.


jonsterling:
We could also just change the parser so that it can recognize which `if` to use, depending on how many arguments it has.

favonia:
How about `wif`?

src/redprl/redprl.lex
```diff
@@ -76,12 +76,12 @@ whitespace = [\ \t];
 
 "?"                => (Tokens.QUESTION (posTuple (size yytext)));
 "fcom"             => (Tokens.FCOM (posTuple (size yytext)));
-"bool"             => (Tokens.BOOL (posTuple (size yytext)));
+"wbool"            => (Tokens.WBOOL (posTuple (size yytext)));
```
favonia: And
```diff
+"wool"             => (Tokens.WBOOL (posTuple (size yytext)));
```

jonsterling:
@favonia I've updated this PR to include `wif` and `wool` 😜

favonia:
@jonsterling Let's hope RedPRL gets popular and then some dedicated user writes a blog post about this. :stuck_out_tongue_closed_eyes:

PR #221, opened 2017-06-18
Avoid some forcing in the stack machine.

favonia:
- [x] Wait for #226.
- [ ] Wait for https://github.com/RedPRL/sml-typed-abts/issues/86.
- [ ] Avoid some forcing.

PR #267, opened 2017-07-02
Final Forcing Framework

favonia:
Related to #262 #230 #221.

### DIFF
--- a/doc/machine.tex
+++ b/doc/machine.tex
@@ -167,7 +167,7 @@ the stack.
     \Step{\mu}{
       \Cfg{a}{\mathcal{E}}{\mathcal{F}}{\pi}
     }{
-      \Cfg{M}{\mathcal{E}'}{(\mathcal{F}'[\mathcal{F}])}{\pi}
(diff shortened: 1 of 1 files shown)


jonsterling:
@favonia The structure you've started writing here looks interesting; I guess it has forcing in one place only, which (based on what you have said) sounds necessary. Do you think that this will work?

favonia:
@jonsterling I think it will work, but not as well as I hope it will do.

favonia:
@jonsterling By the way, do you mind if I create macros like `\EE` for `\mathcal{E}`?

jonsterling:
@favonia That's fine

favonia:
@jonsterling I'd like to do the merging in the end if you feel this pull request is okay.

jonsterling:
Great! I think I understand what you've done, but would you mind explaining how it works?

And am I right that the only instance of forcing is when you compose the two F substitutions?

jonsterling:
yes, whenever you are ready, feel free to merge

favonia:
@jonsterling There are also forcing in the form of `M[...]`.

PR #268, opened 2017-07-05
Sequel to FFFF.

favonia:
@jonsterling Is this what you are asking for?

By the way, there are design issues such as whether we should just keep a list in the end. I think that is a little bit too far unless you would force everything immediately.

### DIFF
--- a/doc/machine.tex
+++ b/doc/machine.tex
@@ -1066,6 +1066,16 @@ So, we have $\DoubleForce{M}{\EE}{\LL}$ everywhere. Here is the definition disgu
   }{
     \DoubleForce{r}{\EE}{\FF} = \Force{\EE^+(r)}{\LL}
   }
+  \and
+  \inferrule{
+  }{
+    \DoubleForce{
+      \perfectparens{
+        \vartheta[\vec{r}]\perfectparens{\overline{\vec{x};\vec{a}.M}}
+      }
+    }{\EE}{\LL} =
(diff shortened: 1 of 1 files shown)


favonia:
Wait, let me fix some stupid bug...

jonsterling:
Hi @favonia, thanks for this!

Would it be possible to add a case in the last section for an arbitrary operator `\theta`? That is,  how forcing works on a term `\theta({x...}[a...]. M ....)`.

jonsterling:
(probably it's straightforward, but it would be helpful to see)

favonia:
@jonsterling I am not sure how to typeset it. Should I use the syntax for meta variables?

jonsterling:
@favonia Yeah, I think so; an operator is (schematically) an operator head `\vartheta` together with a list of binders, which you could write `\vec{x};\vec{a}.M` (as in the 2nd-order substitutions).

The one thing to be cautious about is that operator heads vary over dimensions, so you will have to apply a dimension substitution to an operator head, as in `\vartheta\psi`.

favonia:
@jonsterling So, there are pros and cons for either keeping a list of "final substitution" in the end or collapsing them early. Do you have some intuition about how to figure out which is more efficient?

jonsterling:
@favonia It's hard for me to tell at this point...

favonia:
@jonsterling Anyways, does the note make more sense now? I might still miss something somewhere.

favonia:
@jonsterling Just added some words to give some intuition about the final list.

jonsterling:
@favonia I'll try to read through it again tonight (my brain is a little fried right now). thanks for writing this up!!

PR #270, opened 2017-07-05
Fix test example in readme.

scott-fleischman:


favonia:
@scott-fleischman Thanks!

PR #271, opened 2017-07-06
Add record types and reduce shift/reduce conflicts.

favonia:
Design decisions:
- [ ] Parser should detect duplicate labels.
- [ ] `Syn.in` turns the list into a dict.


jonsterling:
Cool! I think that I remember you mentioning an idea before about trying to make sigma types a special case of this. I think that'd be cool, but I'll leave it up to you as to when to do that!

Do we know what the dynamics of hcom, coe should be for record types? (Probably it should be fairly straightforward based on sigma types, but I wasn't sure.)

favonia:
@jonsterling Well, I intended the record types to be non-dependent. The only sharing part in my mind is the witnesses. The hcom and coe should be component-wise hcom and coe. @cangiuli

jonsterling:
@favonia Ah, I see... Is the reason for non-dependent records that it's easier to get done now, or some other reason?

I'm ok with non-dependent records for now, but ultimately we would want to subsume them under a dependent record type.

favonia:
@jonsterling Well, the fields in non-dependent ones can be reordered, but the ones in dependent ones cannot... once we decided the naming resolution convention then it is probably fine. Now my implementation is toward very extensional record types. We have to decide what to do with duplicate labels. By the way do you think the parser should detect duplicate labels?

jonsterling:
@favonia The fields in a dependent record (witness) can certainly be re-ordered; it's only the type of a dependent record whose fields cannot be re-ordered.

I think that we shouldn't bother with trying to make the parser detect duplicate labels for now.

favonia:
@jonsterling Maybe my question should be rephrased as "when should we report the error of duplicate fields?" Also I think a more reasonable way to do dependent records is to give up reordering altogether. It could be that a type mentioning x is equal to another type not mentioning x, and that our criterion would depend on which version you are using, which in my opinion is not so great.

I just tried Agda with duplicate labels. You can define such a type, but the projection and `record{ }` syntax is not really working...

jonsterling:
@favonia 

1. For duplicate field names, this could be part of the typing rule. I guess you could make it part of the parser, but it might be a little gnarly.

2. I don't understand what you're saying about reordering... There's no problem with having unordered maps as the program, and an ordered telescope as the type; this is because the dependency does not lie at all in the program, but in its type. (but maybe I am missing something important?)

favonia:
@jonsterling Sorry I misread your previous post. Then I think the only corner case is when there are duplicate labels and you have to decide which one if they are allowed.

Alright I got a new idea---for non-dependent cases, how about allowing duplicate fields but require them to be equal?

jonsterling:
@favonia Good question... We could allow duplicate fields in the way you suggest, but it seems like most cases of duplicated fields could be a mistake; do you have a use-case in mind? 

I think the rules would be cleaner if we didn't have duplicate fields (consider how we would compute projections, if duplicate fields were possible!). So, maybe I'm coming around to your original idea, which is to rule it out in the parser.

favonia:
@jonsterling With colons and another change in the syntax, there are no more parser conflicts. Now meta variables and custom operators share the same syntax.

favonia:
Now maybe a special syntax for tuples and records... something like this?
```
(tuple [a -> 1] [b -> 2]) in (record [b -> int] [a -> int])
(! a (tuple [a -> 42])) = 42 in int
```

jonsterling:
wow, amazing! Great job on fixing the parser. I'm so bad at parsers, as you no doubt can tell :)

Can you also look at the pretty printer and update anything that needs changing there? it is in pretty.sml I think. (Or, if you don't have time, I'm happy to do it!)

jonsterling:
That looks good! My suggestions are as follows:
1. in the tuples, can we get rid of the arrows?
2. in the records, can we replace the arrows with colons?

looks fine anyway though.

favonia:
Hmm. I am now wondering whether we should recycle the `< x , y >` notation. Actually, why do we want the middle comma at all? Maybe we could have `<x y z w>` or something like that.... like maybe the following would work?
```
<[a 1] [b 3]> in (* [b : int] [a : int])
(! a <[a 42]>) = 42 in int
```

jonsterling:
@favonia I kind of prefer `(tuple ...)` for now, but maybe we can experiment in the future. My reasoning is that in the term language, I prefer to avoid a proliferation of different kinds of brackets, etc. But if in practice it seems a lot better, we could switch to it.

favonia:
@jonsterling alright. there is still a question about why we want the comma at all, but I guess we can "fix" it in another PR.

jonsterling:
@favonia the `<t1,t2>` notation is not part of the term language, but part of the tactic language. The tactic language doesn't use this sexpr discipline at the moment, so the comma makes sense there for now.

One day we may try to unify these notationally, but for now they are different.

favonia:
@jonsterling It seems there was `LANGLE term COMMA term RANGLE` for `term`. Actually I am refactoring the grammar now and I can just remove the comma if we want...

jonsterling:
@favonia Oh! I didn't realize that. I think that whole production can just be removed. I would like the witness for sigma types to use something like `(pair m n)` or maybe a special case of your tuples.

favonia:
@jonsterling It turns out `,p` is the only term that cannot be viewed as a tactic.

src/redprl/redprl.grm
```diff
@@ -456,12 +479,8 @@ atomicRawTac
   | RULE_LEMMA LBRACKET OPNAME customOpParams RBRACKET (Ast.$$ (O.POLY (O.RULE_LEMMA (OPNAME, customOpParams)), []))
   | RULE_LEMMA OPNAME (Ast.$$ (O.POLY (O.RULE_LEMMA (OPNAME, [])), []))
 
-  | LPAREN LBRACKET OPNAME customOpParams RBRACKET bindings RPAREN (Ast.$$ (O.POLY (O.CUST (OPNAME, customOpParams, NONE)), bindings))
-  | LPAREN OPNAME bindings RPAREN (Ast.$$ (O.POLY (O.CUST (OPNAME, [], NONE)), bindings))
-
-  | operator (Ast.$$ (operator, []))
```
favonia: @jonsterling I realized I might have screwed up something. Did you intend to allow gradual type constructions? I can add this functionality back (if implemented in the interpreter).

jonsterling: Can you elaborate what you mean by `gradual type constructions`?


In general since the tests pass, my guess is that you've done fine 😄

favonia: Oops I meant gradual *term* construction. It's like if I type `pair` it would generate two goals or something.

jonsterling: @favonia Ah; no, we don't allow that.

Instead, we have special forms for tactics that do this sort of thing.

jonsterling: (what I mean is, a *tactic* expression like `if x then else` generates two subgoals, but this is because we parse the empty space as the identity tactic, and you can fill it in with a real tactic as you get further along.)

favonia:
@jonsterling Since the entire `term` can almost be embedded into the tactic language, how about change `,p` to `'p`?

jonsterling:
@favonia I'd rather leave `,p` as-is, since there's an intuition that is being captured in a notation like this code:


    `(pair ,x ,y)

favonia:
@jonsterling Oh I meant maybe we can somehow completely blend `exp` into `tac` so that it would be ```(pair `x `y)```.

jonsterling:
@favonia Ah, I see! I don't think I want to do at the moment that because terms and tactics are different sorts, and we would need to somehow insert implicit coercions somehow... For now, I think I prefer the distinction to be explicit, though maybe we will change our minds in the future 😄

favonia:
@jonsterling Alright. Are you fine with the parser blending almost every term construct into the tactic language, then? I found some seemingly outdated code (probably because some copy-and-paste) and I hope it's more consistent now. However, I don't know the reason for the parser to recognize some forms of `term` as `tactic`, and so I don't know whether pouring almost everything into `tactic` is a good idea.

jonsterling:
@favonia Generally, I'm fine merging things as far as you like within the parser so long as there is no ambiguity, since the sort checker takes care of the important stuff.

I think the reason things were separated the way they are was to avoid parsing conflicts/ambiguities, but since you've fixed that so beautifully, I see no problem with having these as merged as possible within the parser.

favonia:
@jonsterling I am now running into a potential conflict, and thinking about it, perhaps we should change
```
hcom(...) [ ] [ ]
```
to
```
hcom(... [ ] [ ])
```
? The reason is that ``tuple [a 1] [a 2] m`` will probably be ambiguous; `hcom` is fine because of the equal sign.

jonsterling:
@favonia Maybe we should change something, but I don't like the `hcom(... [ ] [ ])`; I'm trying to keep a coherent style of notation based on some extensions to sexprs...

What does `(tuple [a 1] [a 2] m)` mean? Doesn't every element need a label?

favonia:
@jonsterling I mean that the expression
```(a b c d tuple [e f] [g h] i j k)```
can mean both
```(a b c d (tuple [e f] [g h]) i j k)```
and
```(a b c d (tuple [e f]) ([g h] i) j k)```

**Comment by jonsterling** [comment]
@favonia That would not be correct syntax, at least according to my intentions for how RedPRL's term notation is supposed to work. As I said before, it is based on sexprs, which means that the parentheses really are mandatory. Under that discipline, it is not ambiguous.

So, we should definitely not parse `tuple ...` as a term; only `(tuple ...)` should parse.

**Comment by favonia** [comment]
@jonsterling The additional parentheses on the second and the third lines are just to illustrate the problem. The issue is that `(a tuple [b c] d)` can be parsed as
```
(
  a
  tuple [b c]
  d
)
```
or
```
(
  a
  tuple
  [b c] d
)
```

**Comment by jonsterling** [comment]
@favonia Maybe I'm confused, but why are you using the word `tuple` here? Tuples only begin with a parenthesis.

**Comment by jonsterling** [comment]
Furthermore, I don't see how the first parse that you give is meaningful / correct. Certainly it does not represent a form of term that I am familiar with... I think I'm misunderstanding your examples.

**Comment by favonia** [comment]
Hmm. But `tuple` is not taking any expression arguments. Maybe I should say that if you want `(tuple [b c])`, then you might want to reconsider `hcom(... [...] [...])` for consistency. This is actually why I brought up the issue about the `hcom` syntax.

**Comment by favonia** [comment]
Never mind I read the rules wrong.

**Comment by favonia** [comment]
So the `hcom` syntax is actually `(hcom ... [...] [...])` and somehow I failed to read the rules. It seems to me that you want to change it? Or were you just correcting the typo `hcom (...)` where I put `hcom` outside the parentheses?

**Comment by jonsterling** [comment]
@favonia I'm having trouble parsing some of the comments 😉, but let me summarize how I think the syntax is (meant to be):

    (hcom cap [tube1] [tube2] ...)

If it isn't like that now, then we should change it back to that; but if it is like that now, then that's fine 😄

**Comment by jonsterling** [comment]
Oops, I mean

```
    (hcom type cap [tube1] [tube2] ...)

```

But you get the idea

**Comment by favonia** [comment]
@jonsterling I'm stuck at writing out the eq rule for `(! label x)` and `(! label y)`. I want to a new kind of `MATCH` which takes the "`label`th" element which is not at a fixed position in ast.

**Comment by jonsterling** [comment]
@favonia Sorry for the delay in responding, I got sick unfortunately 😄 

Could you write in "pseudo-paper-notation" the rule that you would like to have? (or show where you get stuck?) then I can more easily help.

**Comment by favonia** [comment]
```
 M = N in record{\vec{lbl_i -> ty_i}}
-----------------------------------------
 proj{lbl_i}(M) = proj{lbl_i}(N) in ty_i
```
This is similar to `DProd.FstEq` and `DProd.SndEq`, but I cannot write something like
```sml
val (goalTyA, holeTyA) = makeMatch (O.MONO O.DPROD, 0, holeTy, [], [])
```
because I do not know the position of `lbl_i` in `\vec{lbl_i -> ty_i}`.

jonsterling:
@favonia Aha, I see what you're saying! 😄  It may be sensible to add a judgment for doing what you want, but another option would be to use the left/elim rule for record types.

If you want to add a judgment for this, analogous to `MATCH`, you are welcome to (if you think it will make the rules better behaved).

favonia:
@jonsterling Hmm. I'm not so positive about how negative records should be...

test/success/record.prl
```diff
@@ -0,0 +1,45 @@
...
+Thm RecordTest4 : [
+  (! a (tuple [a tt] [b ff])) = tt in bool
+] by [
+  auto;
+].
+
+Thm RecordTest5(#p) : [
+  #p in record
+  >>
+  #p = tuple in record
```
favonia: Yay!

jonsterling:
@favonia I thought over it again, and I think that using the elim rule is too bureaucratic, so I like your idea of using this auxiliary judgment.

test/success/record.prl
jonsterling: cool!

PR #273, opened 2017-07-12
Improve the Emacs highlighting and the pretty printer.

favonia:


jonsterling:
@favonia BTW, does the flycheck integration even work anymore? I haven't used the emacs mode for months.

favonia:
I only used C-c C-l. By the way, I also changed the pretty printer to print only one side if both sides are alpha-equivalent.

PR #274, opened 2017-07-12
Really add the records. Close #261.

### LINKED ISSUE #261: Labeled record types
A sort for labels, and rules and tactics for record types.

favonia:


favonia:
Sorry I wrongly clicked the review request. This PR is not ready yet.

favonia:
I have not fully figured out whether the parser should sort the labels, but at least it works. Currently the following piece would fail for various reasons, and sorting the labels can solve one of them.
```
Thm RecordTest7(#p) : [
  x : #p in (record [a : bool] [b c : record])
  >>
  #p = (tuple [c (! c #p)] [a (! a #p)] [b tuple]) in (record [c b : record] [a : bool])
] by [
  auto
].
```

src/redprl/judgment.fun
```diff
@@ -27,6 +27,7 @@ struct
            in
              (([],[]), tau)
            end
+       | MATCH_RECORD (lbl, _) => (([],[]), RedPrlSortData.EXP)
```
favonia: Is there a better way to get `EXP`?

favonia:
@jonsterling I have not done the `hcom` and `coe` for records.... but let me change the document instead since we are creating a new machine.

src/redprl/judgment.fun
jonsterling: I can never remember which structure aliases are defined where 😦 

You could just define `structure S = RedPrlSortData` locally in this module, or you could try and see if something like that is already in scope.

BTW [off-topic], one of the frustrating things about ML modules is that when you ascribe a signature, even transparently, you can't access any datatype constructors that are defined in the module externally—unless you extend the signature, which results in a doubling of code.

MatthewFluet: @jonsterling One trick, which doesn't always work, is to define a structure with just the datatype, use that to constrain the signature, and then build up the real structure constrained by the signature:
```
structure FooDT =
struct
  datatype t = A | B | C
end
signature FOO =
sig
  datatype t = datatype FooDT.t
  val toString: t -> string
end
structure Foo : FOO =
struct
  datatype t = datatype FooDT.t
  fun toString x =
    case x of A => "A" | B => "B" | C => "C"
end
```
Of course, if your datatype really depends upon other components that are hard to define outside the module, this doesn't always work.

You can attempt similar tricks that define the whole structure first, then define the signature with the appropriate datatype replications, and then rebind the structure constrained by the signature (shadowing the original):
```
structure Foo = struct ... end
signature FOO = sig datatype t = datatype Foo.t ... end
structure Foo : FOO = Foo
```
This works in MLton (with MLB), but not in SML/NJ (with CM, IIRC, because CM gets a little confused by two definitions of `Foo`.)  Technically SML doesn't allow for:
```
local
structure Foo = struct ... end
in
signature FOO = sig datatype t = datatype Foo.t ... end
structure Foo : FOO = Foo
end
```
because signatures are only allowed at the very top-level.  SML/NJ allows the above and one can achieve much of the same effect with MLBs, but, unfortunately, not with the exact same source code for the two systems.

favonia: @jonsterling I am fine with `RedPrlSortData.EXP` if there is no alias, but I was lost.

jonsterling: @MatthewFluet thanks very much for the tips!

PR #276, opened 2017-07-16
Restart the integer implementations.

favonia:
This is WIP.


favonia:
@jonsterling I feel a little bit uncomfortable with the new categorical judgment `LESS`. On the one hand, everything should reflect something at the judgmental level, and it seems a little bit odd if the `LESS` type depends on the exact implementation of `INT`. On the other hand, having both the judgmental and typal `LESS` sounds like a waste of time.

jonsterling:
@favonia To be honest, I think it is fine to just do this all in the course of defining the `LESS` *type*.

The idea that every type should internalize something at the judgmental level is a principle that applies primarily to *semantics*. Our proof theory should be pragmatic, and adopt that principle when it is helpful, but eschew it when it adds complexity.

With all that said, one possible issue with defining this as a type is that the `head expansion` rules that you defined for the judgment are not derivable, since we do not have a full theory of computational congruence.

favonia:
Okay. It is easy for me to make the change. We will face the same issue again when it comes to internalized equality and universes. I support dropping `EQ_TYPE` in favor of universes.

> With all that said, one possible issue with defining this as a type is that the head expansion rules that you defined for the judgment are not derivable, since we do not have a full theory of computational congruence.

I thought about it, and feel an eager `LESS` type should solve the problem? Are there other issues? Or maybe that a theory of congruence based on stable computation? BTW I want to make many things strict, not just `LESS`.

jonsterling:
1. Let's leave the discussion of `EQ_TYPE` for another day; universes are good, but we will almost certainly always need some special treatment of judgmental equality for types. Why? Because with universes you have to supply a level, but we never need to know the level in type functionality / auxiliary subgoals.

2. I don't think making things eager is necessarily going to help us in a lot of places (it brings its own complications in the presence of open terms), but it might help for `LESS`; it seems highly arbitrary though.

My concern with eagerness is that it is not purely a victory. For instance, imagine if we add eager pairs; then `fst(<x, y>)` does not step to `x` in the open operational semantics, because `<x,y>` is not a value (it is neutral!). Maybe this is fine, but it's an example of where eagerness brings its own complications that we have to think carefully about.

Re: theory of stable congruence, yes, we should have that; but it needs to be developed carefully in concert with Carlo before we consider it for implementation. I chatted with Carlo about this a few times earlier in the year, and we're all very interested in having it, but it may be fairly difficult to achieve.

jonsterling:
Re: internalized equality, that's one of the benefits of linking it directly to a form of judgment: then we can apply computation as we please. This would become less important of course once we develop computational congruence...

So, leaving `LESS` this as a judgment might be helpful for now for the same reason; but I'll leave it up to you.

I look forward to having universes so that we can *define* more of these things rather than writing them out as primitives!

favonia:
Let me take a step back---is there another way to keep `LESS` lazy but `(less (+ 3 2) 6)` provable without a full theory of stable computational congruence?

jonsterling:
Here's a couple ways:

1. Add primitive rules like the ones you had for the judgment which allow head expansion in the arguments of the type.

2. It would be derivable by asserting a judgmental equality between `(+ 3 2)` and `5` at type `nat` (or `int` or whatever), and then using the *unimplemented* Subst rule for equality (we need to implement this!).

The latter is ultimately the kind of thing that we want to be doing for a lot of stuff anyway; even with the Subst rule that we're going to implement, it's a little annoying to do, but that can be papered over using tactics.

favonia:
Alright. So I chose the approach 1, and will probably stay with it for now. For the approach 2, the semantics of less would be
```
tv in (less m n)
```
if and only if `m` evaluates to `num{m'}`, `n` evaluates to `num{n'}`, and `m' < n'` in standard math. The problem is that the above three subgoals appear to require even more judgment forms.

By the way, I might swap `TV` and `AX` in the future. `TV` for the trivial sort `TRIV`, and `AX` is the internalized one. With this RedPRL is more consistent with NuPRL I think, and `TV` seems to match `TRIV` better.

(By the way)^2, an earlier version of `LESS` it even takes a type describing how the LHS is less than the RHS. Then I feel it is too much. :stuck_out_tongue: Another minor issue is that should I use the name `LESS_THAN` instead of `LESS`?

jonsterling:
1. I take it that `TV` is an abbreviation for `trivial` or something? If so, I agree that it makes sense to switch it with `AX`.
2. Either `LESS` or `LESS_THAN` is fine with me, whatever you prefer.

favonia:
> I take it that TV is an abbreviation for trivial or something?

Yes.

favonia:
@jonsterling What does `Wf` in `FromWfJdg` stand for?

jonsterling:
@favonia `Wf` stands for `wellformedness`

favonia:
@jonsterling Is the new `SYNTH.FromWfHyp` following the spirit then? By the way, I don't fully understand why in `lcf_model` the `Computation.EqHeadExpansion` is called... As a user I might expect computation is happening as much as possible, not just one side of a particular kind of judgments.

src/redprl/syntax.sml
```diff
@@ -16,14 +16,18 @@ struct
      VAR of variable * sort
    (* the trivial realizer for equality, which is called 'axiom' in NuPRL. *)
    | AX
+   (* the trivial realizer for interalized equality, less than, etc. *)
```
jonsterling: It might be better to distinguish these two 'trivial' realizers by saying that one targets forms of judgment which have no computational content, and the other targets types whose elements have no computational content.

jonsterling: Also, can we switch these two so that `AX` is for programs and `TV` is of sort `TRIV`?

src/redprl/refiner.fun
```diff
@@ -91,6 +91,21 @@ struct
...
+        val Syn.INT = Syn.out tm
+        val CJ.LESS (a, b) = lookupHyp H z
+      in
+        if Abt.eq (a, tm) orelse Abt.eq (b, tm) then
+          T.empty #> (I, H, Syn.into Syn.INT)
+        else
+          raise Fail "Did not match"
+      end
+
+    fun FromWfHyp z = FromEQ z orelse_ FromLESS z
```
jonsterling: Maybe this should be called `FromHyp`, now that it involves things other than well-formedness hypotheses.

src/redprl/syntax.sml
favonia: I'll fix these soon. In fact, I wanted to fix them.

jonsterling:
Re: head expansion, maybe the user-level `head-expansion` tactic should do as you say. Ultimately, we're going to replace this tactic with some combinators that allow us to specify exactly what computation we want (e.g. compute where for how many steps, etc.).

src/redprl/refiner.fun
favonia: Oh, I thought you meant the well-formedness of having EQ, not that EQ itself being the definition of well-formedness. Can this be viewed as the well-formedness of LESS? I am thinking whether we could figure out more if not restricted to well-formedness.

jonsterling: I am confused by your comment so I will just say what is meant by "well-formedness goal" in the Nuprl tradition: it means a goal to establish that something is a member of some type. So, the term 'wellformedness' here made sense when the hypothesis was an equality judgment (it was a 'wellformedness hypothesis'), but doesn't make sense when the hypothesis is a `less-than` judgment or whatever.

favonia:
By the way, I have not figured out the best way to state `ind` to prove all the things I want. I have a strong feeling that we need internalized equality. Maybe we can discuss this tomorrow.

jonsterling:
@favonia Yes; Bob and I had a chat about this a couple months ago and came to the conclusion that either we need a higher order logical framework, or we need internalized equality. There are important rules that we can't write down otherwise...

At this point, my vote is for internalized equality; the improved LF would help in other ways too, but I don't want to blow up the core of RedPRL for the sake of something we're really going to want to have anyway. It will be cool to experiment with a spectrum of concepts, like pretypes, coe-types, etc.

favonia:
@jonsterling Forget about the "nullary" (I might say "\inf-ary"?) numbers in Nuprl and I'm finishing up a proper definition of integers.

src/redprl/operator.sml
```diff
@@ -108,16 +100,20 @@ struct
...
    * POLY: we have to compare the operators manually. *)
   datatype mono_operator =
-   (* the trivial realizer for equality, which is called 'axiom' in NuPRL. *)
-     AX
+   (* the trivial realizer of sort TRIV for judgments lacking interesting
+    * computational content. *)
+     TV
+   (* the trivial realizer of sort EXP for types lacking interesting
+    * computational content. This is the "ax(iom)" in Nuprl. *)
+   | AX
```
favonia: Are the above lines accurate and clear?

jonsterling: looks great!

favonia:
@jonsterling I need help with the tactic language. I was testing the `nat-rec` and ended up with
```
num.prl:24.3-24.4 [Info]:
  hole.
  Goal 21.
    @0 : nat
    @1 : nat
    a : nat
    >> nat
```
I wanted to say something like `hyp @1` but failed. Not sure how to proceed.

jonsterling:
@favonia Names that begin with `@` are generated by the system and cannot be written by the user. The reason for this is that we want any name mentioned in the tactic script to be provided by the user, to avoid annoying coincidences.

There is a form called `fresh x, y, ... in tac` which allows the user to push their own names into a tactic, so that *these* get used rather than system-generated names. See https://github.com/RedPRL/sml-redprl/blob/52102df25b3000b3e03a6f7d14b0ad1ad0fef7f7/test/success/primitive-sequencing.prl and https://github.com/RedPRL/sml-redprl/blob/20bd3ccc64358d0501112aa0b1a03350012642b5/test/success/logical-investigations.prl#L44 for examples.

favonia:
@jonsterling I see. Next: the following code
```
Thm Pred : [
  (-> nat nat)
] by [
  lam a.
  fresh x,y <- elim a;
  [ `zero ]; auto;
  [ hyp x ]; auto;
  ?
].
```
results in
```
num.prl:26.3-26.4 [Info]:
  hole.
  
num.prl:19.5-19.9 [Warning]:
  1 Remaining Obligations
```
I wonder how it is possible to have a remaining obligation while the question mark shows nothing. Do you have any suggestion before I try brute force crazy debugging?

jonsterling:
@favonia What you have is an issue of scope; recall that `lam` scopes around *everything* to its right. To find the remaining subgoal change your script to:

```
Thm Pred : [
  (-> nat nat)
] by [
  {lam a.
  fresh x,y <- elim a;
  [ `zero ]; auto;
  [ hyp x ]; auto};
  ?
].
```

favonia:
@jonsterling Oh, right, I still have to prove `nat` is a type. LOL

favonia:
By the way I believe the script can be rewritten in a better style, but maybe only you can do that right now. I would put in some ugly tests for now.

jonsterling:
@favonia I think eventually we will just have the development tactics prove things like that automatically for obvious goals, since it's currently a little annoying to have to put in the brackets and then use 'auto'.

favonia:
@jonsterling So we essentially run into the same problem as in `StrictBool.ElimEq` that the user needs to provide a motive.

jonsterling:
@favonia I don't see the problem; can you give an example in terms of `StrictBool.ElimEq`?

favonia:
@jonsterling Suppose I want to type-check
```
if(M, N0, N1) in C
```
The issue is that `C` should equal to `C'[M/x]` for some `C'`, and in general there is no obvious way to figure out what `C'` is even if we know `D = C[TT/x]` and `E = C[FF/x]`. For strict bools, however, we can cheat by picking `C'` as `if(M, D, E)`.

However, the `Nat.ElimEq` seems very difficult, because for the successor case we have to invent in some sense a "fixed point". I think it might be the time to have the user put in something; indeed, can I use `JDG_TERM` to do that?

favonia:
@jonsterling By the way, the current implementation of `StrictBool.ElimEq` only handles the restricted cases where `C' = C` not depending on `x`.

jonsterling:
Ah, i see. the reason I was confused is that there didn't seem to be any problem in the book rule, but after taking a closer look, this is because it does not even attempt to figure out the real motive: it just does the non dependent case. 

To fix both, it is usually possible to guess the correct motive of induction (for both bool and nat) by dividing into two cases:

1. the if statement is a redex: then you should reduce it rather than applying the structural equality rule

2. the if statement is neutral: then is is usually easy to infer the motive.

for now, don't worry about inferring; you could ask the user to type a motive in, but make sure you get binding right (might be tricky if you don't know how). if you can't figure it out, please wait until i have time to do it.

structural equality rules are a lower priority than refinement rules anyway, for which the identity of the motive is always trivial/immediate.


best,
jon

favonia:
@jonsterling I think I probably know how (without thinking carefully about the bindings), but I think it might makes more sense to generate a subgoal for it? I have a feeling that JDG_TERM will do the job but I want to double check with you.

jonsterling:
yes. JDG_TERM is fine for now - jon

favonia:
@jonsterling Let's merge this first, which contains several fixes and improvements along the way.

test/success/num.prl
```diff
@@ -15,3 +15,67 @@ Thm NatOne : [
...
+    [a' ind] (lam [x] (succ (app ind x)))))
+].
+
+Thm Plus2Wf : [
+  Plus2 in (-> nat nat nat)
+] by [
+  fresh a, b, b', ind <- auto;
+  [ `(-> nat nat) ];
+  auto
+].
```
favonia: So, not only we can refine stuff to create the addition function, but also we can prove the addition function is well-typed.

jonsterling: this is awesome!

PR #279, opened 2017-07-21
Clean up all the names at once.

### LINKED ISSUE #278: Syntax for the function application?
I suppose we can have `(lam [x] a)` and `($ a b)` just like `(abs {i} a)` and `(@ a i)`? (The current pretty printer does not generate valid code.)

PS: Other minor things that I will update after #262 is done:
- `S1_REC`
- `PATH_APP`
- `APP`
- `S_IF`, `IF`, etc

favonia:
Close #278.


src/redprl/categorical_judgment.sml
```diff
@@ -10,6 +10,8 @@ sig
   val MEM : 'a * 'a -> 'a redprl_jdg
   val TYPE : 'a -> 'a redprl_jdg
 
+  val fromAst : RedPrlAst.ast -> RedPrlAst.ast redprl_jdg
```
favonia: I added `fromAst` so that `sortOfJudgment` is just a composition of `synthesis` and `fromAst` and this should prevent future oversights. I did not change the signature because I am not sure if this is a Good Idea.

PR #281, opened 2017-07-21
Left rule for path types

### LINKED ISSUE #191: refinement rule for path-elimination
Right now, I don't think we have a refinement rule for eliminating paths. It should just take a dimension directly as a parameter. Otherwise, it would be analogous to the Pi left rule.

Then, I can stop doing that dumb thing where I insert a witness in a proof, like `'(,p @ r)` or some crap.

- [ ] Implement primitive refinement rule ~, expose in refiner's interface (we can't put this in the `Elim` megarule, since the dimension would need to be a parameter)~
- [ ] Add development-calculus operator `let b = a@r.`, analogous to path application but where `x` is a hyp symbol and `r` is a `dim` parameter.
- [ ] add stuff for parsing, etc.

jonsterling:
To resolve #191.

### DIFF
--- a/src/redprl/categorical_judgment.sig
+++ b/src/redprl/categorical_judgment.sig
@@ -1,25 +1,26 @@
 signature CATEGORICAL_JUDGMENT =
 sig
-  type 'a jdg
+  type ('sym, 'a) jdg
 
-  val map : ('a -> 'b) -> 'a jdg -> 'b jdg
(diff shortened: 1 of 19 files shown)


jonsterling:
This is pretty sweet, as the rule doesn't actually take any extra parameters (even the dimension "parameter" is constructed through refinement). As a result, it can participate directly in the general `elim` infrastructure.

The only kind of weird thing is that to "return" a dimension in a tactic, you have to use the `&r` form. (Suggestions for better syntax are of course welcome!)

PR #284, opened 2017-07-22
Polytypes, stage 0.

favonia:
Goal: seamless integration of strictly discrete Kan types, Kan types, coe types, hcom types and pretypes.


favonia:
Needs help about the concrete syntax.
```redprl
A distype
A kantype (alias: A type)
A hcomtype
A coetype
A pretype

M = N din A
M = N kin A (alias: M = N in A)
M = N hin A
M = N cin A
M = N pin A

M dsynth
M ksynth
M hsynth
M csynth
M psynth

A dtrue
A ktrue
A htrue
A ctrue
A ptrue
```

Or maybe `in` can be an alias of `pin`?

jonsterling:
Re: concrete syntax, I would really prefer that instead of having a bunch of things like `dtrue, ktrue, etc.`, we have just a parameter like `true{kan}` or whatever.

Ultimately, it doesn't matter too much, so feel free to do what you prefer...

src/redprl/operator.sml
```diff
@@ -88,10 +88,86 @@ structure RedPrlParameterTerm = AbtParameterTerm (RedPrlParameter)
...
+     | PRE => "pre"
+
+  local
+    structure Internal :
+    (* this could be the new meet semi-lattice *)
+    sig
+      type t = kind
+      val top : t
+      val leq : t * t -> bool
+      val meet : t * t -> t
```
favonia: I realized the current meet semi-lattice could be improved but I do not want to do the propagation again till I am sure what we really want.

favonia:
How about this?
```
a = b in M with discrete
a = b in M with kan
a = b in M with hcom
a = b in M with coe
a = b in M with pre
a = b in M
```
`a = b in M` becomes an alias of `a = b in M with pre`.

jonsterling:
I like that

src/redprl/operator.sml
jonsterling: Yeah, I would recommend not trying to perfect it until we determine just what we need.

favonia: Usually we (or I) can change the signatures and structures at the same time, but now changing the signature is too expensive.

jonsterling: @favonia Yeah; in the future, one option that I favor would be to define new signatures inside the RedPRL project until they are perfected, and after that, move them into whatever library makes sense. that can lower the cost.

But in this case, do whatever you feel is most convenient for you.

favonia:
Hmm I am not sure about the left elimination rules of various types... how do we know that `A -> B` is Kan implies that `B` is Kan in the current code? It could be that `B` is not Kan but somehow magically `A -> B` is Kan, right? I do not have a courterexample right away. @cangiuli

favonia:
For now I will put use the top element (i.e., pretypes) in the context for those "unknown" types.

favonia:
@jonsterling How did we get away without checking the goal type in `DFun.AppEq`? Is it a clever trick or a bug? https://github.com/RedPRL/sml-redprl/blob/ea12b29b5d905a6ee34ed9748812369db281d8b0/src/redprl/refiner_types.fun#L623-L638

jonsterling:
@favonia The goal type is known to be a type because it follows from the `SYNTH` goal.

favonia:
@jonsterling But how do we even know it's synthesizing the same type?

jonsterling:
@favonia What do you mean by synthesizing the same type? The semantics of synthesis are meant to be, if `SYNTH m ~~> a` then `m in a` and thence `a pretype`.

jonsterling:
@favonia Ahhh!! I see. Yes, there is a bug, we need one more subgoal!

src/redprl/categorical_judgment.sml
```diff
@@ -1,42 +1,70 @@
...
+     *   There exists some `m` of sort `tau`.
+     *   The realizer is such an `m` of sort `tau`.
+     *)
    | TERM of RedPrlSort.t
+    (* `DIM_SUBST ((p, r), m)`:
+     *   `p` should eventually be `DIM_REF p'` for some dimension term `p'`,
+     *   `r` is a dimension variable and `m` is an expression of sort `EXP`.
+     *   The realizer is the result of the substitution of `r` by `p'` in `m`.
+     *)
    | DIM_SUBST of ('a * 'sym) * 'a
```
favonia: @jonsterling I wish that you could do an early review on the meaning of all the judgments.

jonsterling: Sure, I'll do so today.

favonia:
> @favonia Ahhh!! I see. Yes, there is a bug, we need one more subgoal!

Actually, the rule is so wrong that it needs more goals. Anyway I will fix it along the way.

src/redprl/categorical_judgment.sml
```diff
@@ -1,42 +1,70 @@
 structure RedPrlCategoricalJudgment :
 sig
   datatype ('sym, 'a) redprl_jdg =
-     EQ of ('a * 'a) * 'a
-   | TRUE of 'a
-   | EQ_TYPE of 'a * 'a
-   | SYNTH of 'a
+    (* `EQ_TYPE ((a, b), k)`:
+     *   `a` and `b` are equal types, taking into account the additional structures specified by `k`.
+     *   (That is, `EQ_TYPE ((a, b), KAN)` means they are equally Kan, not just having the same PER.)
```
jonsterling: We should be a bit more cautious in this sentence, since the meaning of type equality in CHiTT is never that the types have the same PER (CHiTT has an intensional type equality). Maybe we should say, "...means that they are equally Kan, in addition to being equal pretypes".

src/redprl/categorical_judgment.sml
```diff
@@ -1,42 +1,70 @@
...
+   | EQ of ('a * 'a) * ('a * RedPrlKind.t)
+    (* `TRUE (a, k)`:
+     *   `EQ_TYPE ((a, a), k)` and there exists a term `m` such that
+     *   `EQ ((m, m), (a, k))` is provable.
+     *   The realizer is such an `m` of sort `EXP`.
+     *)
+   | TRUE of ('a * RedPrlKind.t)
+    (* `SYNTH (m, k)`:
+     *   There exists a term `a` such that `EQ ((m, m), (a, k))` is provable.
+     *   Usually there is a canonical choice of `a`, but the canonicity is not required.
```
jonsterling: Can we remove the sentence about canonical choices? I think that sentence makes sense, but it might be confusing for people looking at the documentation.

src/redprl/categorical_judgment.sml
```diff
@@ -1,42 +1,70 @@
...
+   | SYNTH of ('a * RedPrlKind.t)
+    (* `TERM tau`:
+     *   There exists some `m` of sort `tau`.
+     *   The realizer is such an `m` of sort `tau`.
+     *)
    | TERM of RedPrlSort.t
+    (* `DIM_SUBST ((p, r), m)`:
+     *   `p` should eventually be `DIM_REF p'` for some dimension term `p'`,
+     *   `r` is a dimension variable and `m` is an expression of sort `EXP`.
+     *   The realizer is the result of the substitution of `r` by `p'` in `m`.
```
jonsterling: very good. nice job figuring out what I did, even though I didn't really document anything properly.

favonia:
@jonsterling Thanks. All fixed. Now I am more confident.

favonia:
Progress report: `refiner_types.fun` at least compiles.

favonia:
Let me just re-restart everything.

favonia:
There's a bug in the current `Nat.Elim`... I will fix it along the way.

favonia:
Changes in the queue:

1. `TUPLE` syntax will take a list instead of a splay tree.
2. `CUBICAL` instead of `PRE`.
3. Many helper functions for `Hyps`.
4. `Nat.Elim` will be weakened for safety (currently broken).

jonsterling:
@favonia 👍 sounds very good!

favonia:
@jonsterling This is not fully working yet. How to get rid of the auto here?
```
Thm BoolTest : [
  (-> wbool wbool)
] by [
  [lam x. if x then `tt else `ff]; auto
].
```
The remaining goal without `auto` is `x : wbool type with cubical >> x : wbool type with kan`. This can be proved by `auto`, and I thought it can then be discharged automatically?

jonsterling:
@favonia Huh, I'm surprised that it's not discharged automatically w/o the extra `auto`; I will investigate it in the morning and get back to you.

favonia:
Alright let me change the concrete syntax and optimize something else...

src/redprl/syntax.sml
```diff
@@ -10,7 +10,20 @@ struct
...
+  struct
+    val empty = []
+
+    exception Absent
+    fun lookup (lbl : label) fields =
+      case List.find (fn x => #1 x = lbl) fields of
+        SOME (_, tm) => tm
+      | NONE => raise Absent
+
+    fun remove lbl = List.filter (fn f => #1 f <> lbl)
```
jonsterling: At least in SML/NJ, this doesn't typecheck. Replacing this line with the following works:

```
    fun remove lbl = List.filter (fn (lbl', _) => lbl' <> lbl)
```

(Or one could add a type annotation; the problem is "unresolved flex record".)

jonsterling:
@favonia I looked at the issue where the typehood goal wasn't getting resolved automatically, and here's the issue.

The "development tactics" (like `if x then t1 else t2`) expect the subgoals generated by the elim rule that they call to be of a certain shape: that is, they expect a certain number of subgoals, and for those subgoals to be of a specific kind. But with the very clever things you have done to the refiner (such as testing for equality, and also things like `makeTypeIfLess`), this expectation has been disrupted, which makes it extremely difficult to write good tactics that call rules.

There are a couple of ways to resolve this problem:

1. We could make rules *simple* and **predictable** in the sense that they don't try to do anything clever when generating their subgoals. This has the benefit of making it possible to write naïve tactics that call rules and resolve some of their subgoals, without requiring the tactic to replicate the logic of the rule (for instance, determining whether a subgoal was generated by checking if something is "known" to be a type, or if it is alpha equivalent to something, etc.). My opinion has always been that *rules* should be simple and predictable. But this approach has the drawback that it will result in more subgoals to prove than are necessary in some cases.

2. We could instead change the "development" tactics to not make any assumptions about the shape of the auxiliary subgoals—that is, they would only make assumptions about the initial prefix of the subgoals telescope which includes the *principal* goals, but not the well-formedness goals. I am not sure how this would work, but maybe it is possible to design something workable.

3. An intermediate step between these ideas is to have a uniform shape of subgoals generated by rules; that means, rather than generate only `2` subgoals in one cases and `3` subgoals in another case, always generate `3` subgoals; in the case where one of those subgoals is trivial, rather than omitting it, instead generate a subgoal for a "trivial"/"unit" judgment which you could add.

I lean toward #3 as the most well-behaved and modular approach to this. And I think it would not involve a very invasive code change, since all those conditional subgoal-making-helper-functions could be made to just return this trivial goal in case the subgoal was redundant (such as in the case of the `...IfLess` and `...IfDifferent` things).

What do you think?

favonia:
@jonsterling Ah, I see. For the second approach, how about returning two sequences of goals? My problem with the third approach is how to discharge a trivial goal if a rule is applied to a trivial goal... should it generate the same number of trivial subgoals, pretending the trivial goal is of the right form. By the way, could you point to the code of this particular development tactic so that I have a better idea about it in practice?

src/redprl/refiner_types.fun
```diff
@@ -150,68 +171,93 @@ struct
...
-        val CJ.TRUE ty = lookupHyp H z
+        val (I, H) >> CJ.TRUE (cz, k) = jdg
+        (* if(FCOM) steps to COM *)
+        val k = K.meet (k, K.COM)
+        (* for now we ignore the kind in the context *)
+        val CJ.TRUE (ty, _) = Hyps.lookup z H
         val Syn.WBOOL = Syn.out ty
+        
+        (* kind because of FCOM *)
+        val goalKind = makeTypeIfLess (I, H) (cz, k) K.top
```
favonia: If the failure is due to this rule, the goal `goalKind` is actually unconditional. So I guess the tactic assumes this rule only generates two goals? Now this rule will generates three, always.

jonsterling: @favonia Ah! I see! Then we can easily fix this case in the tactic elaborator.

jonsterling:
@favonia 

1. I think should not return two sequences of goals; this begins to break the notion of "refinement rule" in a way that may make sense from a pragmatic POV, but which doesn't seem to have a theoretical basis.

2. Good question about trivial subgoals; I think that the question of how to discharge a trivial subgoal when a rule is applied to it will not come up. My thought is that if we generate some subgoal like `TRIVIAL`, when `Auto` encounters it, that will cause it to be discharged; I think it is no problem if we do not allow clients to make any assumptions about the specific content of the auxiliary subgoals, but that we guarantee their number.

jonsterling:
To see what kinds of assumptions are made about the subgoals generated by rules, check out `tactic_elaborator.fun`.

src/redprl/refiner_types.fun
favonia: Let me study these further to make sure there are no other surprises.

src/redprl/refiner.fun
```diff
@@ -468,20 +528,6 @@ struct
       |>: goal1 >: goal2 #> (I, H, substVar (hole1, z) hole2)
     end
 
-  fun makeNamePopper alpha =
```
favonia: I don't know what this is for and it's not used, so I removed it.

src/redprl/refiner.fun
```diff
@@ -708,32 +754,57 @@ struct
...
+             | CJ.TRUE (a1, k1) =>
+                 isUnary andalso Abt.eq (a0, a1)
+                 andalso K.greatestMeetRight' (SOME k0, SOME k1) <> SOME k0
+             | _ => false
+        in
+          case Hyps.search H isUseful of
+            SOME (lbl, _) =>
+              ( TypeEquality.FromEqTypeHyp lbl orelse_
+                TypeEquality.FromWfHyp lbl orelse_
+                TypeEquality.FromTrueHyp lbl)
```
favonia: @jonsterling I believe some of these have been implemented somewhere, but I could not find them... so here are my version.

src/redprl/refiner.fun
```diff
@@ -708,32 +754,57 @@ struct
...
+          case Hyps.search H isUseful of
+            SOME (lbl, _) => Equality.FromWfHyp lbl alpha jdg
+          | NONE => raise E.error [Fpp.text "Could not find suitable hypothesis"]
+        end
     in
       fun AutoStep sign alpha jdg = 
         StepJdg sign alpha jdg
           handle exn => 
-            FindHyp alpha jdg
+            (EqTypeFromHyp orelse_ EqFromWfHyp) alpha jdg
```
favonia: @jonsterling The actual change is that the above complicated search replaced `Hyp.Project` in `AutoStep`.

src/redprl/operator.sml
```diff
@@ -94,11 +94,110 @@ structure RedPrlParameterTerm = AbtParameterTerm (RedPrlParameter)
...
+         | (DISCRETE, _) => DISCRETE
+         | (_, KAN) => top
+         | (KAN, HCOM) => COE
+         | (KAN, COE) => HCOM
+         | (KAN, _) => KAN
+         | (HCOM, HCOM) => top
+         | (HCOM, _) => HCOM
+         | (COE, COE) => top
+         | (COE, _) => COE
+         | (CUBICAL, CUBICAL) => top
```
favonia: @jonsterling Now, everything reduces to `top`, `meet` and `greatestMeetRight`.

favonia:
Two more examples to fix and now I'm looking at `tactic_elaborator`. Other necessary "fixes" to the examples are just beautiful beyond words.

favonia:
@jonsterling Fixing `DEV_S1_ELIM` is trivial, but `DEV_BOOL_ELIM` is a little bit annoying because only `WBool.Elim` checks the kind of the motive, not `Bool.Elim`, and thus one has to branch on the type to know the number of subgoals. I am not sure about the *Immortal* way to fix this.

jonsterling:
Good point :) I can also think about how to improve it... For now, just do the easiest thing that can possibly work, and we can improve it over time :)

favonia:
@jonsterling Seriously, one possible solution is to forget about weak bools. It's practically useless except for being a prototype of something that is actually useful.

favonia:
(I know we should keep it until RedPRL and CHTT become mature (and popular)...)

jonsterling:
> Seriously, one possible solution is to forget about weak bools. It's practically useless except for being a prototype of something that is actually useful

@favonia That's a great point! We put them in because it was in the paper, but they are probably the most useless example of a "HIT" that I've ever seen 😆

favonia:
@jonsterling In addition to `Lcf.allSeq`, could we have `Lcf.tailSeqAfter 2` or something like that? Then we say we want to apply `auto` after a certain point. This is the second approach.

jonsterling:
@favonia That's a really great idea! I can add that to the library tomorrow.

jonsterling:
That would make our tactics a lot more modular.

test/success/num.prl
```diff
@@ -22,7 +22,7 @@ Thm Pred : [
   lam a.
   fresh a', ind -> elim a;
   [ `zero ];
-  [ use a' ]
+  [ `,a' ]
```
favonia: The question is, should we strengthen the tactic `use` to accept mismatch kinds?

jonsterling: @favonia Good question; I think it would be cool if `use` could "just work" in cases where the kinds are not the same, but compatible in the appropriate way. But you know better than I how feasible that is.

favonia:
I suggest merging this while leaving the following stuff to fix:

1. Enhance `use` to handle kind mismatch better.
2. Rename rules and tactics.
3. Add `Lcf.allSeqAfter` or something like that, and then remove those extra `auto`.
4. Remove rules which (obviously) require user inputs without clear contexts from `auto`.
5. Optimize the kind-checking further.
6. Understand the mysterious behavior of the custom tactic in `num.prl`.
7. Highlighting.
8. Update failing test cases.
9. For some reasons `lam` does not take underscores anymore?

jonsterling:
@favonia Ok, I agree! Merge whenever you want, and we can undertake the remaining tasks later.

jonsterling:
I'm restoring the branch so that I can look into that weird issue we talked about last night.

PR #290, opened 2017-07-24
Primitive rule registry

jonsterling:
This allows proofs to call primitive rules by name, rather than trying to steer `auto` and `auto-step` in the right direction. It's kind of a hack, but I think it is more economical (engineering-wise) than trying to do something nicer.


jonsterling:
The development calculus tactics should be updated to use this instead of `auto-step`, I think. That will make dev calculus programs a bit less permissive, which is ultimately a good thing and will prevent bugs and mistakes in proof scripts.

Issue #288, opened 2017-07-23
Make 'auto' have "deep" alternative search


jonsterling:
The title of this ticket is not so clear, but let me explain.

Currently, when checking equations like `hcom = m` there is a choice between using cap-equality and tube-equality. Right now, cap-equality is favored, but this is not always the correct choice: sometimes both are immediately possible, choosing cap equality will ultimately result in impossible goals.

One way to resolve this is to have `auto` generate a kind of search tree, so that in places where there are multiple possible choices, we should actually try all of them! Then, `auto` would terminate when *one* of its search branches succeeds.

To do this, we should expose a primitive `Auto` tactic from the refiner, instead of the current implementation which actually *reduces* the tactic `auto` to an LCF recursive tactic.

jonsterling:
To be honest, what would be cooler is to adjust the tactic superstructure to support something like `t1 ANDELSE t2`, which has a different semantics from `t1 ORELSE t2`. In particular, `(t1 ANDELSE t2) THEN t3` is the same as `(t1 THEN t3) ANDELSE (t2 THEN t3)`.

But that would be a difficult and invasive change; I will think about it.

favonia:
For the current theory (I believe) there is a very good ordering of rules to avoid backtracking, but of course complexity can grow over time.

jonsterling:
I already have very basic examples where hcom checks fail because of the ordering.

favonia:
@jonsterling Can you show it? I would like to know.

jonsterling:
Yes, I'll share when I get home :)

jonsterling:
@favonia Here's an example you can play with:

```

Thm NatSymm : [
  (->
   [m n : nat]
   (path {_} nat m n)
   (path {_} nat n m))
] by [
  lam a. lam b. lam pab.
  // I'm using the primitive interface rather than "<i> ..." to avoid 'auto' getting called;
  // this way, we can expose and understand the problem.
  fresh i:dim <- auto-step;
  // for our first subgoal, provide the hcom explicitly.
  [`(hcom{0 ~> 1} nat ,a 
     [i=0 {j} (@ ,pab j)]
     [i=1 {_} ,a])];
  [?oh-no, // this is true by the tube rule, but that won't get called (I think the cap rule is getting called instead)
   ?fine; auto // this subgoal is fine, I think the cap rule gets called
   ]
].
```

jonsterling:
[...]

PR #291, opened 2017-07-24
improve ordering of rules called in auto, make hcom sym examples work

jonsterling:
Per the discussion in #288, I have experimented here with favoring the equational `hcom` rules over head expansion in the `auto-step` decision procedure. This had the result of making that naïve version of that `NatSymm` proof work fine with `auto`.

I also found that @favonia had already put in a failing test called `test/failure/bad-hcom-step.prl`, which identified precisely this problem already. @favonia's Thought is very powerful!

Since that example now works, I have deleted the test (since we have an analogous test for `nat` under `test/success/num.prl`).

What I've learned from this is that head-expansion is not always best—I had hoped that we would be able to use it always to simplify composition typing problems, but it turns out that this is not really a good idea. Instead, we can use head expansion to simplify composition problems that we don't know how to solve using the equational rules.


test/failure/bad-hcom-step.prl
```diff
@@ -1,10 +0,0 @@
-Thm BoolSym : [
-  (-> 
-   [a b : bool]
-   (path {_} bool a b)
-   (path {_} bool b a))
-] by [
-  { lam a. lam b. lam p. <i> 
-     `(hcom{0 ~> 1} bool ,a [i=0 {j} (@ ,p j)] [i=1 {_} ,a])};
-  [auto-step; head-expand]
```
jonsterling: Previously, this test should have failed for two reasons:

1. The overly conservative `safeStep` stuff would have thrown an error when trying to reduce the `hcom`.
2. After the machine was fixed, this test continued to fail for a different reason: head expansion succeeded, leaving behind a still-unsolved goal. This describes the period of time before the `<i>` and `lam` development tactics automatically called `auto` on their auxiliary subgoals.
3. Finally, after I changed the development tactics to call `auto`, this test would have failed because an impossible subgoal was generated by `auto`, which is now fixed in this PR.

A very complicated story!

favonia: This is probably because I forgot to change bool to wool... Glad that it's useful!

favonia:
Another possibility: can we have "equality reflection" for discrete types?

jonsterling:
@favonia Do you mean "equality reflection" in the sense of being able to turn a *path* into an equality? Sounds interesting. We can ask @cangiuli.

That kind of reflection principle would also have rendered the "bad" goals provable, as an alternative.

test/failure/bad-hcom-step.prl
favonia: By the way I would suggest change it to wbool and keep it.

jonsterling: That's a good idea! I'll add it back as you say.

favonia:
> @favonia Do you mean "equality reflection" in the sense of being able to turn a path into an equality? Sounds interesting. We can ask @cangiuli.

Yes. This way, maybe we can still reduce everything first. This would be a difficult meta-theorem to prove, and it might not be the "optimal" order even if it is okay...

jonsterling:
@favonia Yeah; though I think that the reflection principle should be a last resort, since it will not be easy at all to automate its application. We should include it if it is true, but it is not a replacement for "optimal" ordering of rules, if that is even possible.

favonia:
Alright, but the current change probably does not address this sufficiently---it is possible that M steps to M' and then evals to M'', and we actually have to stop at M' to apply other rules.

jonsterling:
yeah. I'm worried that there is no general way to do this; ultimately, people may need to drop down to calling primitive rules sometimes. :(

cangiuli:
> "equality reflection" for discrete types

This should be easily provable from any correct definition of discrete. Suppose we have a path; then it is equal to each of its endpoints by discreteness, and by transitivity, those endpoints are equal to each other.

PR #292, opened 2017-07-24
Fix broken app eq rule

jonsterling:


jonsterling:
what a relief that all the tests still work! 😆

PR #293, opened 2017-07-24
Dependent records

jonsterling:


jonsterling:
I think this is basically good to go.

Once I get a good elimination rule in for records, as well as appropriate dev calculus syntax, we can consider just deleting sigma types entirely and adding some notation.

I want to extend the dev tactics with something like the following:

```
lbl-pat ::= lbl | lbl = name
tactic ::= let {lbl-pat,...} = hyp. tactic
```

The behavior of this would be to call the elimination rule, using the appropriate names; crucially, not all names need to be given, and they can be given in any order. So this will take a little bit of programming on my part to make it work, but ultimately I think it would be quite nice.

I'm not sure I'll do it in this PR though.

jonsterling:
Heh; I think with the current way that things are set up for the LCF apparatus, it's not actually clear how to implement the semantics of the dev calculus form that I want. It has to do with the fact that I don't know what the goal is whilst "computing" a tactic form, and that I can't make recursive calls to the tactic interpreter from the rule interpreter. (It's ok if this doesn't make sense to anyone but me at the moment).

I think I would like to un-abstract some of the stuff that is currently in a library, and inline it into RedPRL to avoid this kind of problem. I'll think about doing that over the next few days.

src/redprl/redprl.grm
```diff
@@ -86,16 +87,44 @@ struct
...
+
+
+  fun checkNoDuplicates pos lbls = 
     let
-      val (lbls, tys) = ListPair.unzip (flatten tbs)
+      fun insert (lbl, set) = 
+        if LabelSet.member set lbl then
+          raise RedPrlError.annotate pos (RedPrlError.error [Fpp.text "Duplicate label", Fpp.text lbl])
+        else
+          LabelSet.insert set lbl
```
favonia: This can be done by `operate'` (or `operate`?) in one step, but maybe you don't want to use clever tricks...

jonsterling: Thanks for the reminder! I will change it, I think it makes sense to use whatever operations are available.

jonsterling: @favonia Aha, it turns out that `operate` is part of `DICT` but not of `SET`.

jonsterling:
At this point the only thing that remains is the dev-calculus forms for dependent records; currently what is available are the *low-level* interfaces: you can use the refiner to introduce and eliminate dependent records, and you can also write codes manually and "typecheck" them.

I would like to merge this now, since to get the dev-calculus forms working in the way that I want them to work, I will need to address #294.

src/redprl/refiner_types.fun
```diff
@@ -807,35 +813,35 @@ struct
...
+                 val m1 = Syn.LabelDict.lookup map1 lbl
+                 val env' = Var.Ctx.insert env var m0
+                 val goals' = goals >: makeEq (I, H) ((m0, m1), ty')
+                 val hyps' = hyps @> (var, CJ.TRUE ty)
+                 val famGoals' = if isFirst then famGoals else famGoals >: makeType (I, hyps) ty
+               in
+                 {goals = goals', famGoals = famGoals', env = env', hyps = hyps', isFirst = false}
+               end)
+            {goals = Lcf.Tl.empty, famGoals = Lcf.Tl.empty, env = Var.Ctx.empty, hyps = H, isFirst = true}
+            fields
```
favonia: Did you check whether there are "unused" fields?

favonia: I guess it could work either way, but I guess you (or we?) do not want "extra" fields.

jonsterling: @favonia If I understand your question correctly, the answer is that I am moving toward an extensional treatment of records, under which we require that the specified fields exist, but do not penalize the existence of random other garbage in the record.

Record values under the new regime are un-ordered, and should be viewed in an extensional sense as finite maps that only need to obey the specification given by the record telescope.

PR #297, opened 2017-07-27
make function and path intro tactics n-ary

jonsterling:


test/success/num.prl
```diff
@@ -86,7 +86,7 @@ Thm NatSymm : [
    (path {_} nat a b)
    (path {_} nat b a))
 ] by [
-  lam a. lam b. lam pab.
+  lam a b pab.
```
favonia: Can we have something called "intro(s)" which basically does the same thing but with a warning that this is RedPRL not Coq/...?

jonsterling:
Maybe we could do something like that :) But I will hold off on it.

PR #302, opened 2017-08-01
Scope bugs

jonsterling:
I introduced some scoping bugs with #298. I'm now adding in dynamic assertions to catch them, and will fix them in this PR.


jonsterling:
Ah, I believe this is due to a bug in the ABT library, fixed by https://github.com/RedPRL/sml-typed-abts/pull/93.

PR #306, opened 2017-08-02
higher order unification...

jonsterling:
I'm adding higher order pattern unification to the abt library, and i'm testing it out in RedPRL. Currently I have a tactical to do second-order pattern matching on a term.

I've also added a tactical `query gl <- goal. tac` which projects the conclusion of the current sequent, which in combination with the matching tactic enables something like `match goal`. Of course, this is not as powerful as Coq's goal matcher, because we have no way to inspect the hypotheses just yet. I would like to add that modularly through queries that target the left side of the current sequent.

### DIFF
--- a/src/redprl.cm
+++ b/src/redprl.cm
@@ -17,6 +17,7 @@ is
   $libs/sml-cats/cats.cm
   $libs/sml-telescopes/telescopes.cm
   $libs/sml-typed-abts/abt.cm
+  $libs/sml-typed-abts/abt-unify.cm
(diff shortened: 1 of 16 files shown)


favonia:
You probably forgot to update `sml-dependent-lcf`.

jonsterling:
@favonia Indeed I did. By the way, it's not necessary to immediately mechanically update every single library that depends on something once you make any commit to this; it's only important if you introduce a breaking change. Ultimately, it will be noticed as soon as you want to make a change to `sml-dependent-lcf`, so it's not a huge deal.

PR #307, opened 2017-08-04
begin sketching out some ideas for a REAL metalanguage

jonsterling:
This will be a long-running branch as I experiment with some ideas for a ML-style metalanguage for RedPRL to replace our shitty tactic language.

I'm taking a lot of inspiration from Pierre-Marie Pedrot's work on Ltac2, as well as Bauer/Stone/etc.'s work on Andromeda.


src/redprl/metalanguage/dynamics.fun
```diff
@@ -0,0 +1,163 @@
...
+   | SND of hole
+   | EACH of {tactics: mlterm list, done: Lcf.jdg Lcf.Tl.telescope, metaenv: Tm.metaenv}
+   | SPLICE of {tactics: mlterm list, var: Lcf.L.var, done: Lcf.jdg Lcf.Tl.telescope, remaining: Lcf.jdg Lcf.Tl.telescope, metaenv: Tm.metaenv}
+
+  type stack = (continuation, mlterm) closure list
+
+  type names = int -> Tm.symbol
+  datatype state = LOCAL of Lcf.jdg | GLOBAL of Lcf.jdg Lcf.state
+  datatype control = ## of (state * names) * (mlterm, mlterm) closure
+  datatype machine = |> of control * stack | <| of control * stack
```
jonsterling: Currently the machine regards the proof state as "state", and executes scripts and programs against it. This seems the most obvious thing to do, but it would be better if we could somehow disentangle things so that the machine is actually producing in the end a (semantic) tactic. This would make it easier to program the behavior of builtin tacticals in SML, without having to be able to encode their functionality in RedML.

I'm not certain if I can do what I described above, but I'm going to think about it.

jonsterling:
I'm going to have another try at this where I do a "denotational" interpreter into a tactic monad, `jdg -> state * 'a`.

jonsterling:
I think that now I can begin to envision how the high-level development tactics would be programmed in this ML.

I could have an internalized ML type of "object-scopes" `x. t` (in which an object-symbol is bound), and of "symbol references" `&x` which are terms that refer to an object-symbol. Then, 

```sml
fun lam (tac : unit object_scope) : unit = 
  push x in
    refine 'dfun/intro';
    fork [tac // &x, auto]
  end
```

Then, a proof that uses this could look like `lam x. lam y. tac`.

favonia:
Is there a reason to commit files (seemingly) generated by cmlex/cmyacc?

jonsterling:
> Is there a reason to commit files (seemingly) generated by cmlex/cmyacc?

I'm not sure yet, but the reason I have been committing these is that `cmlex`/`cmyacc` are not provided by default in SML distributions, and they can't be easily installed with `apt-get`, etc. (and of course, `smackage` is very buggy and a nightmare to get working.) So, it might make sense to allow people to build RedPRL without installing these—but that they would need to install them if they want to change the lexer or parser spec.

Longterm, I have been thinking too that it may make sense as well to keep these files checked in, because even if those tools become lost to history, RedPRL's code could still be built.

jonsterling:
But I haven't decided anything, and this branch is so experimental that I would recommend ignoring it for a while yet 😉

jonsterling:
Elaborating ML-style pattern matches to the focusing calculus may be too difficult for now. I may try a different approach.

cangiuli:
Please don't delete this branch.

PR #312, opened 2017-08-08
Update the travis script.

favonia:


favonia:
LOL... the MLton compiler probably needs more memory than the container mode provides.

favonia:
A remaining issue is to have travis deploy only once as we have two builds now...

jonsterling:
@favonia We should probably just get rid of the deploy stuff, since that doesn't really do the right thing at this point, I think.

PR #313, opened 2017-08-08
Update failing test cases.

favonia:
They need some love. ❤️


test/failure/lexical-error.prl
```diff
@@ -1,3 +1,3 @@
-Thm BoolTest : [ (-> bool bool) true ] by [
-  { lam _. `_tt }; auto
+Thm LexicalError : [ (-> bool bool) true ] by [
+  { lam x. `_tt }; auto
```
favonia: For some reasons `lam` do not take underscores anymore. Could it be that we do not have the "don't-care" pattern yet?

jonsterling: I probably broke that... I'll check into it.

Issue #311, opened 2017-08-08
Correct Nat.Elim without higher-order LF


favonia:
@cangiuli @jonsterling We should try to figure out the correct version of `Nat.Elim` without higher-order LF. The task is to eliminate `x:nat` from
```
Gamma0, x:nat, Gamma1 >> c true.
```
When `Gamma1` is empty, I believe it's safe to refine this into two subgoals:
```
Gamma0 >> c[0/x] true.
Gamma0, y:nat, z:c[y/x] true >> c[succ y/x] true.
```
However, I am not sure what to do with non-empty `Gamma1`.

jonsterling:
Why does it matter whether `Gamma1` is empty? Also, why not just follow the rules given in Nuprl which do not require any higher order framework? See the integer elimination rule from Nuprl:

![screen shot 2017-08-07 at 8 26 36 pm](https://user-images.githubusercontent.com/55057/29051195-d52ad6e8-7bae-11e7-96fb-eeec9e9dba55.png)

jonsterling:
Now I can't remember why I thought the left rule required a higher-order LF—Bob convinced me at one point, but I can't remember why. @cangiuli do you remember?

favonia:
@jonsterling I think this is maybe the "weaker" form because it is not doing induction on the entire sequent, and maybe this is what we can achieve at best. This also connects to my previous question whether we want to link `z` and `x`. In the above rule, the answer is no.

jonsterling:
@favonia That's a good point (re: it not doing induction on the full sequent); I think we should start with this rule, and then we can discover whether we actually need more than this. I have a feeling that the stronger cases could be encoded in some cases using function types, especially once we have more of your stuff in place. It's telling that Nuprl doesn't need anything more than this, and they have done a lot of gnarly stuff with integers and nats...

Re: whether `z` and `x` are related, that's another good observation, thanks for pointing it out; I think that we ultimately need them to be related for things like function types, but for this we should leave them not equated.

PR #315, opened 2017-08-08
Polytypes, stage 1.

favonia:
A continuation of #284.

### DIFF
--- a/src/redprl/refiner_types.fun
+++ b/src/redprl/refiner_types.fun
@@ -18,8 +18,8 @@ struct
   (* Rules in a type should be ordered as follows:
    *
    * EqType: the type formation rule.
-   * EqX: the introduction rule for the constructor X in a positive type.
-   * True: the introduction rule in a negative type.
+   * EqX: the introduction rule for the constructor X of a positive type.
+   * True: the introduction rule of a negative type.
    * Eta: the eta rule, if any.
    * Elim/FullElim: the elimination rule (see the paragraph below).
    * EqElim/EqX: structural equality for eliminators.
--- a/src/redprl/refiner_types.fun
+++ b/src/redprl/refiner_types.fun
@@ -142,7 +142,7 @@ struct
         (* ff branch *)
         val ff = Syn.into Syn.FF
         val Hff = Hyps.substAfter (z, ff) H
-        val (goalF, holeF) = makeGoal @@ (I, Htt) >> CJ.map_ (substVar (tt, z)) catjdg
+        val (goalF, holeF) = makeGoal @@ (I, Hff) >> CJ.map_ (substVar (ff, z)) catjdg
 
         val evidence =
           case catjdg of
--- a/src/redprl/refiner_types.fun
+++ b/src/redprl/refiner_types.fun
@@ -126,15 +126,10 @@ struct
         T.empty #> (I, H, trivial)
       end
 
-    (* This is an induction on the full sequent, justified by
-     * the semantics of open judgments.
-     *
-     * We mimicked Nuprl to keep the variable z in the subgoals.
-     *)
-    fun Elim z _ jdg =
+    fun FullElim z _ jdg =
       let
-        val _ = RedPrlLog.trace "Bool.Elim"
-        val (I, H) >> CJ.TRUE (cz, k) = jdg
+        val _ = RedPrlLog.trace "Bool.FullElim"
+        val (I, H) >> catjdg = jdg
         (* for now we ignore the kind in the context *)
         val CJ.TRUE (ty, _) = Hyps.lookup z H
         val Syn.BOOL = Syn.out ty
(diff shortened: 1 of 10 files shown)


favonia:
@jonsterling `refiner_kit.fun` is accumulating helper functions and I feel some of them could go into the upstream libs...

jonsterling:
@favonia If you feel any helper functions are mature enough to be moved into libraries, feel free to do so.

jonsterling:
@favonia Can you explain what is meant precisely by "induction on the full sequent" in your comments?

favonia:
> @favonia Can you explain what is meant precisely by "induction on the full sequent" in your comments?

It's the same issue we discussed in #311 and the wording can be improved. There are two kinds of Elim given the sequent
```
\Gamma, x : A, \Delta >> J
```
The "full elim" applies to arbitrary judgments. It enumerates all elements `a0, a1...` in A and consider
```
\Gamma, x : A, \Delta[ai/x] >> J[ai/x]
```
The "weak elim" only applies to the true goal. It looks at the branches in the elim expression and generates this kind of subgoals:
```
\Gamma, x : A, \Delta, (arguments to the constructor b0) >> C[b0 args/x]
```
For strict bools, we can do the "full elim" by consider tt and ff. For the circle, integers, natural numbers, weak bools, etc, I think it's hard or impossible to do full elim without some fancy LF.

favonia:
- Positive types that are easy to do "full elim":
bool, records, union, equality, void, 
- Positive types that seem difficult to do "full elim":
weak bool, integers, natural numbers, list, the circle, general inductively types
- Negative types operate differently and are more "uniform".

jonsterling:
@favonia Ok, thanks! I see what you're saying. I don't really understand the wording of "induction on the full sequent", but we can improve the wording.

1. It seems like there's actually two different things happening: one is the question of whether we can put our hypotheses before `\Delta` and then do a substitution on `\Delta`, or if we have to put the new hypotheses after `\Delta` and only substitute in the conclusion. The second issue is whether a rule can apply to an arbitrary judgment or just to `A true` goals.  These seemed orthogonal to me, and the difficulties in giving a 'full elim' for some types seem to have to do with one or the other of these issues, but not always both.

2. What would be extremely helpful is if you could write down a couple of these 'imaginary' full elim rules which we would like, but can't express in the current framework (maybe in a new issue, or an existing one?). That way, I could keep that in mind as we continue to evolve RedPRL and its logical framework.

favonia:
@jonsterling Sure. For this PR, how about naming these "full elim" as "FullElim" and then I put a long paragraph in the beginning of `refiner_types.fun`? Are you okay with "full/local" terminology?

I could not think of an example where one can do full elim in `A true` goal but not another. These full elims and the substitution on `\Delta` are justified directly by the semantics of open judgments, and should work on arbitrary open judgments as long as there's a sensible realizer for the conclusion. On the other hand, "local elims" are justified by other principles and are thus more limited.

jonsterling:
@favonia For a single type, we would only implement one of those two categories of elim; so I don't know if it makes sense to actually rename them, though some documentation that describes the distinction would be welcome.

For instance, in Nuprl the elim rule for integers is just called "integer elimination", and they don't make any terminological distinction between this and the elim rule for booleans; the general practice seems just to supply whatever the strongest expressible elimination rule is, and then call it "elimination".

I haven't yet fully understood the "full/local" terminology, but I'm fine with it if you add your proposed documentation paragraph.

And if you really want to do the renames, feel free to do so—I don't feel really strongly about what I've said, so I trust your judgment if you decide that you need to rename things.

favonia:
Yes, the renaming is mostly for documentation purpose---I was thinking that maybe it's good to remind the tactic writers of the nature of those rules are, but I will think about it.

favonia:
Some documentation is added. @cangiuli I believe you would be interested. @jonsterling do not merge this yet because I still have to rename rules according to the new thought and merge two elim's in Bool into one (I stupidly thought they have to be separated).

favonia:
@jonsterling Now it's done.

src/redprl/refiner_types.fun
```diff
@@ -949,49 +968,88 @@ struct
...
-        val a1u = substSymbol (P.ret u, v) a1v
+        val w = alpha 0
+        val a0w = substSymbol (P.ret w, u) a0u
+        val a1w = substSymbol (P.ret w, v) a1v
+        val tyGoal = makeEqType (I @ [(w,P.DIM)], H) ((a0w, a1w), ka)
 
         val a00 = substSymbol (P.APP P.DIM0, u) a0u
         val a01 = substSymbol (P.APP P.DIM1, u) a0u
-
-        val tyGoal = makeEqType (I, H) ((a0u, a1u), ka)
```
favonia: This goal was wrong and was fixed by introducing `w`.

src/redprl/refiner_types.fun
```diff
@@ -15,6 +15,76 @@ struct
...
+   *
+   * Another example is the natural number type `nat` which has elements `zero`,
+   * `(succ zero)`, ...; this also has infinitely many cases. It is imaginable
+   * that we can generate two "higher-order subgoals" in some higher-order LF:
+   *
+   * 1. `\Gamma, x:nat, \Delta[zero/x] >> J[zero/x]`
+   * 2. if `\Gamma, x:nat, \Delta[#n/x] >> J[#n/x]` is provable for some
+   *    numeral `n`, then `\Gamma, x:nat, \Delta[(succ #n)/x] >> J[(succ #n)/x]`
+   *    is provable.
+   *
```
jonsterling: I don't understand why we would want this very non-standard rule... or any of these things with "infinitely many cases". We're doing proof theory!

We should not present the completely *standard* left rules for things like naturals as if they are deficient in some way. I would suggest that to avoid confusion, we don't say these things in this comment/documentation.

jonsterling: When Bob and I originally discussed the question of whether we need a higher order judgment for the nat-elim rule, I think that this wasn't what we had in mind; instead what happened is that we had gotten confused, and forgot what the standard rule looked like.

favonia: Ah, sorry I just merged this PR. Let me open another one.

jonsterling: @favonia That's ok! I was hoping you'd merge it, since I didn't want my quibbles to prevent you from getting your excellent work merged in. 😄

favonia: I think this might be the only thing that could make sense if it ever made sense, but I support the idea of removing them from the comment unless we know concretely what kind of "higher-order LF" we want/have.

jonsterling: > I think this might be the only thing that could make sense if it ever made sense

I agree 😄 I think that Bob and I confused each other so badly that what we had in mind made zero sense. Your idea is cool, and is probably the only way to read "we need higher order judgments for nat-elim" so that it means something real.

PR #316, opened 2017-08-09
Tweaks to the elimination rules.

favonia:
Continuing the work in #315.
- [x] Fix the document.
- [x] Double-check whether we are following the Nuprl convention when it does not hurt.

### DIFF
--- a/src/redprl/refiner_types.fun
+++ b/src/redprl/refiner_types.fun
@@ -53,25 +53,14 @@ struct
    * for example: it has elements `base`, `loop` and `fcom`; there are infinitely
    * many different shapes of `fcom` and thus infinitely many cases (unless our LF
    * can deal with them uniformly), and there are (infinitely) many equalities
-   * between these elements due to cubicality.
+   * between these elements due to cubicality. Another example is the natural
+   * number type `nat` which has elements `zero`, `(succ zero)`, ...; this also
+   * leads to infinitely many cases.
    *
-   * Another example is the natural number type `nat` which has elements `zero`,
-   * `(succ zero)`, ...; this also has infinitely many cases. It is imaginable
-   * that we can generate two "higher-order subgoals" in some higher-order LF:
-   *
-   * 1. `\Gamma, x:nat, \Delta[zero/x] >> J[zero/x]`
-   * 2. if `\Gamma, x:nat, \Delta[#n/x] >> J[#n/x]` is provable for some
-   *    numeral `n`, then `\Gamma, x:nat, \Delta[(succ #n)/x] >> J[(succ #n)/x]`
-   *    is provable.
-   *
-   * One also has to think carefully about how to construct proper realizers.
-   *
-   *
-   * Before we have a more powerful LF, however, we can still have a "weaker" or
-   * "local" elimination principle, `Elim`, that is just trying to generate a
-   * term that would satisfy the structural equality of the eliminator (EqElim).
-   * For the natural number type, the weaker elimination principle generates
-   * the following two subgoals:
+   * Before we have a more powerful LF, however, we can still have the second
+   * category of elimination rules, `Elim`, which is just trying to generate
+   * a term that would satisfy the structural equality (EqElim). For the natural
+   * number type, the elimination rule of this category generates these two subgoals
    *
    * 1. `\Gamma, x:nat, \Delta >> C[zero/x] true`
    * 2. `\Gamma, x:nat, \Delta, y:nat, z:C[y/x] >> C[(succ y)/x] true`
@@ -79,10 +68,10 @@ struct
    * and combine the realizers from the two subgoals into something of the form
    * `(nat-rec x ...)`.
    *
-   * There is no reason to have a weaker elimination principle if the stronger
-   * one is already there. Therefore, there is `Bool.FullElim` but not `Bool.Elim`.
-   * In Nuprl, everything is named "elimination" but I (favonia) think maybe
-   * it is useful to remind tactic writers of what is really going on.
+   * There is no reason to have an elimination principle of the second category
+   * if there exists one of the first category. Therefore, there is `Bool.FullElim`
+   * but no `Bool.Elim`. In Nuprl, everything is named "elimination" but I (favonia)
+   * think it might be useful to remind tactic writers of its category.
    *)
 
   structure Bool =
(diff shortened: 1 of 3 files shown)


src/redprl/refiner_types.fun
```diff
@@ -53,36 +53,25 @@ struct
...
    * `(nat-rec x ...)`.
    *
-   * There is no reason to have a weaker elimination principle if the stronger
-   * one is already there. Therefore, there is `Bool.FullElim` but not `Bool.Elim`.
-   * In Nuprl, everything is named "elimination" but I (favonia) think maybe
-   * it is useful to remind tactic writers of what is really going on.
+   * There is no reason to have an elimination principle of the second category
+   * if there exists one of the first category. Therefore, there is `Bool.FullElim`
+   * but no `Bool.Elim`. In Nuprl, everything is named "elimination" but I (favonia)
+   * think it might be useful to remind tactic writers of its category.
```
jonsterling: This is better, but I honestly don't think there is any need to even make a distinction between rules which have a real premise for every "case", and rules which introduce hypotheses for the inductive cases. I think it results in muddying of the waters, since even without such an explanation, nobody would be surprised to see that RedPRL does not have the omega rule 😆...

All rules of the former class are 'degenerately' members of the second class, and in a PRL, there is no reason to try and talk about the latter. Moreover, a higher order logical framework would not enable us to account for the latter kind of rule anyway, which is really something like the `omega` rule. To account for that, you would need a computational logical framework, which itself only pushes the problem up one "meta" level.

So, my recommendation is that we remove all discussion of these so-called "full" elim rules, and just call everything 'elim'.

If we want to make a distinction between different kinds of elimination rules, it should be between those which can target only `TRUE` and those which can target other forms of judgment. But the issue of whether you have a premise for every case seems neither here nor there.

favonia: Okay. I think you mean whether we can target something other than `TRUE` is orthogonal to whether there are potentially infinitely many subgoals, but I do not see how to deal with other kinds of goals by introducing hypotheses, and perhaps they are closely related issues? Other than this, if you meant keeping other parts of this long comment and just renaming FullElim back to Elim, I can easily do so.

> Moreover, a higher order logical framework would not enable us to account for the latter kind of rule anyway, which is really something like the omega rule.
Did you mean the _former_ kind?

jonsterling: > ... but I do not see how to deal with other kinds of goals by introducing hypotheses, and perhaps they are closely related issues?

Could you elaborate? You had brought up questions about doing things like matching on things like fcoms, but I think that's a separate matter and IMO we shouldn't think about it for the time being. My preference is if we never add facilities for doing that, though maybe some day it will become necessary. In any case, since my default position is that we wouldn't do this, I don't want to have a lot of comments and documentation that talk about it, since it may make things confusing for people as they try to learn from the codebase.

> Did you mean the former kind?

Oops, yes!

favonia: A better example is `nat`. How to refine these goals
```
Gamma, x:nat, Delta >> m = n in a
Gamma, x:nat, Delta >> a = b type
Gamma, x:nat, Delta >> synth m
```
by using some elimination rule of the "weaker" class (without internalized equality or the universe)? My point is that the restriction to true goals is because we do not know how to deal with other kinds of goals by introducing hypotheses.

favonia: For the documentation, is it clear to you that `S1.Elim` and `WBool.Elim` must be of the weaker class? They were implemented (unjustifiedly) as ones of the stronger class and so I had to weaken them without seeing a proof. I would feel safer to remove the entire comment if it's indeed clear that we can only have `WBool.Elim`, `S1.Elim` and `Nat.Elim` of the weaker class.

jonsterling: Thanks for your further comments! I thought that by "I do not see how to deal with other kinds of goals..." you were talking about an issue that had to do with higher-order stuff or whatever, but it seems like what you really mean is just that we don't have a way to eliminate (e.g.) weak bool into something that's not a type.

You're probably already there, but I just want to make sure that we are being cautious to distinguish between a real issue (different elim rules can eliminate into different classes of judgments) and a non-issue (we can't phrase some higher order judgment).

> For the documentation, is it clear to you that S1.Elim and WBool.Elim must be of the weaker class? They were implemented (unjustifiedly) as ones of the stronger class

Is this actually true? At least when I wrote these rules months ago, I made sure that they could only eliminate into `TRUE`. Or at least I thought I did. Can you point to a commit where this was not correct, and then fixed?

favonia: @jonsterling It indeed could only eliminate into `TRUE`, but that is not the only difference between strong/weak classes. Before #315 it also substitutes `base` or `loop` for `z` in `Delta` for the goal
```
Gamma, z:S1, Delta >> cz true
```
which IMO is not obviously justified. (It could be true but then I want to see a proof.)

jonsterling: @favonia Ah, I see what you mean. Right, that sounds like a plausible principle for `S1` but it would require some proof. That's a good catch.

My recommendation is that we just call all of these "elimination" and not bother with delineating it terminologically, with the expectation that each elim rule be the strongest thing that we have justified (or consider to be 'obviously' true).

It's fine to have a comment that warns about the issue of which parts of the sequent you are allowed to substitute into, but I still don't think we should be trying to categorize these terminologically at this point.

favonia: In fact, with internalized equality and binary sums I have a counterexample to the original `WBool.Elim`---just let the motive be `(+ (= z tt) (= z ff))`. For `S1.Elim` I need more time to prove or disprove it.

favonia: Sorry I take it back. The above motive would not pass the kind-checking. I need more tricks to form a valid counterexample.

src/redprl/refiner_types.fun
```diff
@@ -21,70 +21,14 @@ struct
    * EqX: the introduction rule for the constructor X of a positive type.
    * True: the introduction rule of a negative type.
    * Eta: the eta rule, if any.
-   * Elim/FullElim: the elimination rule (see the long comment below).
+   * Elim: the elimination rule. This should be in the strongest form we
+   *   consider to be obviously true within the limit of RedPRL.
```
favonia: Like this?

jonsterling: ❤️ !!

favonia:
Thanks. Now let's knock down the equality types.

PR #317, opened 2017-08-09
Polytypes, stage 2.

favonia:
Continuation of #284 and #315.

- [ ] Enhance the tactic `use` to handle kind mismatch better.
- [ ] Wait for `Lcf.allSeqAfter` or something like that, and then remove those extra `auto`.

The above list is left for stage 3.


favonia:
Since a serious bug was found and fixed in this PR, we need to merge this first.

PR #318, opened 2017-08-09
Add equality types, stage 0.

favonia:
What a long detour.

### DIFF
--- a/src/redprl/redprl.lex
+++ b/src/redprl/redprl.lex
@@ -75,6 +75,7 @@ whitespace = [\ \t];
 "_"                => (Tokens.UNDER (posTuple (size yytext)));
 "?"                => (Tokens.QUESTION (posTuple (size yytext)));
 
+"ax"               => (Tokens.AX (posTuple (size yytext)));
 "fcom"             => (Tokens.FCOM (posTuple (size yytext)));
 "bool"             => (Tokens.BOOL (posTuple (size yytext)));
 "tt"               => (Tokens.TT (posTuple (size yytext)));
@@ -100,7 +101,7 @@ whitespace = [\ \t];
 "lam"              => (Tokens.LAMBDA (posTuple (size yytext)));
 "record"           => (Tokens.RECORD (posTuple (size yytext)));
 "tuple"            => (Tokens.TUPLE (posTuple (size yytext)));
-"path"             => (Tokens.PATHS (posTuple (size yytext)));
+"path"             => (Tokens.PATH (posTuple (size yytext)));
 "abs"              => (Tokens.ABS (posTuple (size yytext)));
 "hcom"             => (Tokens.HCOM (posTuple (size yytext)));
 "coe"              => (Tokens.COE (posTuple (size yytext)));
--- a/src/redprl/refiner.fun
+++ b/src/redprl/refiner.fun
@@ -389,6 +389,7 @@ struct
         val _ = RedPrlLog.trace "Equality.RewriteTrue"
         val (I, H) >> CJ.TRUE (mainGoal, k) = jdg
         val CJ.EQ ((m, n), (ty, k')) = Hyps.lookup z H
+
         val x = alpha 0
         val Hx = H @> (x, CJ.TRUE (ty, k'))
         val (motiveGoal, motiveHole) = makeTerm (I, Hx) O.EXP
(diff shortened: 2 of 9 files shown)


jonsterling:
this is so exciting!!

src/redprl/refiner_types.fun
```diff
@@ -1055,4 +1055,111 @@ struct
...
+        val ax = Syn.into Syn.AX
+        val (goal, hole) =
+          makeGoal
+            @@ (I, Hyps.interposeThenSubstAfter (z, |@> (u, CJ.EQ ((m, n), (a, K.top))), ax) H)
+            >> CJ.map_ (substVar (ax, z)) catjdg
+      in
+        |>: goal #> (I, H, VarKit.subst (trivial, u) hole)
+      end
+
+    fun RewriteTrue z alpha jdg =
```
favonia: This is essentially a duplicate of `Equality.RewriteTrue`. Not sure if we want this at all.

jonsterling: We could factor this through an elimination rule for eq types that replaces `x : (eq A M N) true >> J[x]` with `x : M = N in A >> J[Ax]`.

src/redprl/redprl.lex
```diff
@@ -100,7 +101,7 @@ whitespace = [\ \t];
 "lam"              => (Tokens.LAMBDA (posTuple (size yytext)));
 "record"           => (Tokens.RECORD (posTuple (size yytext)));
 "tuple"            => (Tokens.TUPLE (posTuple (size yytext)));
-"path"             => (Tokens.PATHS (posTuple (size yytext)));
+"path"             => (Tokens.PATH (posTuple (size yytext)));
```
jonsterling: thank you for changing that! 👍

src/redprl/refiner_types.fun
favonia: It's slightly different, because in Nuprl style the original hypothesis is (almost?) always kept.

test/success/equality.prl
```diff
@@ -0,0 +1,27 @@
...
+  b : #A
+  >> (= #A a b) hcom type
+] by [ auto ].
+
+Thm EqualityKind1(#A) : [
+  m : #A discrete type,
+  a : #A,
+  b : #A
+  >> (= #A a b) kan type
+] by [ auto ].
```
jonsterling: VERY COOL. 😎

src/redprl/refiner_types.fun
jonsterling: @favonia That's a good point... I'm fine with keeping the hypothesis!

favonia: Sorry I thought you were talking about the elim rule. Yes, and that is why `Elim` is designed like that. I am leaning toward removing this today.

jonsterling: ok!

Great work.

src/redprl/refiner.fun
```diff
@@ -807,24 +815,18 @@ struct
...
          | Syn.RECORD _ => Record.Elim
          | Syn.PATH_TY _ => Path.Elim
+         | Syn.EQUALITY _ => InternalizedEquality.Elim
          | _ => raise E.error [Fpp.text "Could not find suitable elimination rule for", TermPrinter.ppTerm ty]
 
-      fun StepEq ty =
-        case Syn.out ty of
-           Syn.BOOL => Bool.Elim
-         | _ => raise E.error [Fpp.text "Could not find suitable elimination rule for", TermPrinter.ppTerm ty]
+      val FromEq = Equality.RewriteTrue (* todo: rewrite other kinds of goals *)
```
favonia: Maybe it makes more sense to have a separate `subst` or `rewrite` tactic?

jonsterling: @favonia Yeah... I think that rewriting is not really elimination, so it makes sense to give it its own tactic.

favonia:
I want to prove
```
x:nat >> (= int x x) true
```
but then this seems we have to allow non-members in equality types, following Nuprl's approach. The issue is that kinds interact with membership in a strange (?) way.

jonsterling:
I think we can do it without adopting nuprl's semantics for equality types... I'll expand on this more in a bit.

jonsterling:
@favonia OK, so the thing here is that to show `x : nat >> (= int x x) true` in the PRL means to simultaneously establish the truth of the consequent and its typehood; so, this would proceed as follows:

1. First, apply the equality intro rule to the goal `x : nat >> (= int x x) true`
2. Now we have to show this: `x : nat >> x = x in int`
3. But that should hold by the subtyping (not sure if we actually have such a rule, but we should at some point).

The reason that they had an alternative semantics for equality types in Nuprl (where `M,N` need not be elements of `A` for `Eq(A; M; N)` to be a type) was so that you could negate statements of membership. Without these semantics, it isn't possible to express something like `Eq(M; M; A) -> Void`, since for the implication to be a type, `M` must be an element of `A`.

That's an interesting and useful thing, but I think we don't need it yet.

favonia:
I was indeed thinking about proving that particular subtying rule is admissible, and then came to that open judgment. I thought alternative equality types, if ever working, are better than manually putting all the subtyping & subkinding rules into the core.

jonsterling:
> I thought alternative equality types, if ever working, are better than manually putting all the subtyping & subkinding rules into the core.

That may be so, but I am not sure you can get away with not having subtyping rules... In Nuprl, I think that they have some subtyping rules, but these are very general since `nat` is *formed* in Nuprl using a set comprehension. But I don't remember the specifics.

In any case, if we eventually feel that we need the Nuprl-style equality types, we could try and find out if they work in our setting. But I think that to have a useful proof theory for them, we might need the `base` type first (even if you don't necessarily need `base` to formulate the semantics of those equality types).

favonia:
Alright. Another possibility is to have general subtypes...

favonia:
Again there's still something we can tweak, but maybe in the next stage:

- [ ] recognize paths in discrete types as equality
- [ ] a separate tactic for rewriting

PR #320, opened 2017-08-10
Add integers, stage 1.

favonia:
We took a long detour through kinds and equality types, and now we are back!

### DIFF
--- a/src/redprl/variable_kit.sml
+++ b/src/redprl/variable_kit.sml
@@ -1,0 +1,17 @@
+(* variable kit *)
+
+structure VarKit =
+struct
+  fun ctxFromList l =
+    List.foldl
+      (fn ((tm, x), dict) => Var.Ctx.insert dict x tm)
+      Var.Ctx.empty l
+
+  fun toExp x = Syntax.into (Syntax.VAR (x, RedPrlSortData.EXP))
+
+  val renameMany = RedPrlAbt.renameVars o ctxFromList
+  fun rename r = renameMany [r]
+
+  val substMany = RedPrlAbt.substVarenv o ctxFromList
+  fun subst s = substMany [s]
+end
--- a/test/success/num.prl
+++ b/test/success/num.prl
@@ -43,7 +43,7 @@ Tac SimpleNatRecEq = [
   }
 ].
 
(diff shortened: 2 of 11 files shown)


favonia:
Hmm... can't prove the things I want (easily) without subtyping.

test/success/num.prl
```diff
@@ -96,4 +96,45 @@ Thm NatSymm : [
...
+  ($ IntPlus 4 3) = 7 in int
+] by [ auto ].
+
+Thm Int-6Plus10 : [
+  ($ IntPlus -6 10) = 4 in int
+] by [ auto ].
+
+Thm Int-1Plus-9 : [
+  ($ IntPlus -1 -9) = -10 in int
+] by [ auto ].
```
favonia: By the way, `($ IntPlus 3000 4000) = 7000 in int` seems to be an interesting input for profiling.

test/success/num.prl
```diff
@@ -96,4 +96,45 @@ Thm NatSymm : [
...
 
-Print NatSymm.
\ No newline at end of file
+Print NatSymm.
+
+Thm IntPred : [
+  (-> int int)
+] by [
+  lam a. fresh a', junk -> elim a;
+  [ `-1, fresh n, junk -> elim a'; [`0, `(succ ,n)]
```
favonia: So, if we have subtyping, then `fresh n, junk -> elim a'; [exact 0, exact (succ ,n)]` can be replaced by `use a'`.

favonia:
The proof is ugly but cannot really be improved without (significantly) changing the core of the refiner. I added `Synth.Witness`.

src/redprl/variable_kit.sml
```diff
@@ -0,0 +1,17 @@
+(* variable kit *)
+
+structure VarKit =
+struct
+  fun ctxFromList l =
+    List.foldl
+      (fn ((tm, x), dict) => Var.Ctx.insert dict x tm)
+      Var.Ctx.empty l
+
+  fun toExp x = Syntax.into (Syntax.VAR (x, RedPrlSortData.EXP))
```
favonia: This function is probably the most useful one in this structure.

Issue #319, opened 2017-08-10
Consider requiring hypotheses to be `A true`


jonsterling:
Now that @favonia has implemented this cool kind lattice thing, it is possible for us to begin revisiting the design of RedPRL's proof theory with an aim toward simplification. 

One thing that might make sense in the near term is to require the hypotheses in a sequent to all be of the form `A true with k`; to be precise, each hypothesis would just be a pair of a type and a kind. This would simplify and clarify the meaning of RedPRL sequents, especially as far as functionality is concerned.

We would still want the conclusion to be an arbitrary categorical judgment.

In order for this to work, we need:

- [x] Equality types
- [x] Universes

There's also some dark bits of RedPRL that I (JS) would have to change, but that I could certainly do that if we felt it was a good idea.

favonia:
With equality types, can we ban equality judgment now? To be more precise, only `true` and `eqtype` will be allowed in the context. After we have the universe the `eqtype` can be banned as well.

jonsterling:
@favonia It would be weird to prevent only some non-`true` judgments from appearing in the context. We should wait until we can do this systematically.

favonia:
@jonsterling Hmm... we can also try to work out a cubical universe first, but I already have some ugly proof involving equality now.

favonia:
Could we make it `A true with k at i`? A complete description of a universe involves (1) the kind and (2) the level (the size), so we can demand that everything in the context must be a triple `(type, kind, level)`.

jonsterling:
@favonia Yeah, I was thinking we might want something like that... I think the only difficult thing is that since RedPRL's refinement stuff doesn't support "multiple outputs" primitively, I guess we would need to have a sort for these triples and then need to actually have some judgments for projecting things out of it. All this can be papered over using helper functions, though.

favonia:
> actually have some judgments for projecting things out of it

Are you suggesting three new categorical judgments `PROJ_EXP` `PROJ_KIND` and `PROJ_LEVEL` which somehow gets these from something of sort `STRATIFIED BLAH`? Actually, due to these complexities I am giving up the idea of returning the strongest kind.

favonia:
[...]

PR #325, opened 2017-08-12
Universes, stage 0.

favonia:
Only when #319 is also done the universes can be considered done.

### DIFF
--- a/src/redprl.cm
+++ b/src/redprl.cm
@@ -36,6 +36,7 @@ is
   redprl/operator.sml
   redprl/abt.sml
   redprl/syntax.sml
+  redprl/univ_level.sml
   redprl/variable_kit.sml
 
   redprl/pretty.sml
--- a/src/redprl.cm
+++ b/src/redprl.cm
@@ -36,6 +36,7 @@ is
   redprl/operator.sml
   redprl/abt.sml
   redprl/syntax.sml
+  redprl/univ_level.sig
   redprl/univ_level.sml
   redprl/variable_kit.sml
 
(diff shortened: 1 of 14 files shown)


favonia:
@jonsterling Ah your comment is squashed out...

favonia:
@jonsterling What is your favorite *concrete* notation for "a Kan universe at level (max i j)" in our S#-expression style?

jonsterling:
Maybe (U kan (lmax i j))?  But I don't care too much, do it as you like :)

favonia:
Actually I do not know what to do with `(lmax i j)` in S#-expressions. They are not expressions.

The outer parentheses must be `{}`.

jonsterling:
There's not a hard rule; sometimes the param vs term distinction is an implementation detail. no need to get the notation perfect, since we can easily change it.

favonia:
Before reverting the mode of the universe level argument and figuring out the right way to have multiple outputs, we can still type-check everything by explicitly providing the universe levels. And that's what I'll do now.

jonsterling:
ok, agreed!

favonia:
Because there are no elim rules for universe levels, I am switching to inf-nary representations. Examples:
```
{lmax {labove a 3} {labove b} c {labove {lmax c d} 1} 2}
{lvl 100}
```

jonsterling:
@favonia This looks very interesting! Your ideas for universe levels are a lot more advanced than anything I have considered, and it seems like we will be able to do a lot of cool stuff with it! Nice.

favonia:
I would like to merge this PR first because it also refactors other parts such as the `RedPrlError` module.

favonia:
Oh, wait, I would like to remove `STRATIFIED` as well.

favonia:
@jonsterling Now it's really done (for this PR).

PR #327, opened 2017-08-14
Universes, stage1.

favonia:
Continuation of #325. (Will rebase once #325 is merged.)

### DIFF
--- a/src/redprl/machine.fun
+++ b/src/redprl/machine.fun
@@ -518,6 +518,11 @@ struct
            end
          | _ => raise Fail "Impossible record type")
 
+     | O.POLY (O.UNIVERSE _) $ _ || (_, []) => raise Final
+     | O.POLY (O.UNIVERSE _) $ _ || (syms, HCOM (dir, HOLE, cap, tubes) :: stk) =>
+         E.raiseError (E.UNIMPLEMENTED (Fpp.text "hcom operations of universes"))
+     | O.POLY (O.UNIVERSE _) $ _ || (syms, COE (_, (u, _), coercee) :: stk) => CRITICAL @@ coercee || (SymSet.remove syms u, stk)
+
      | _ => raise Stuck
 
   fun step sign stability unfolding (tm || stk) =
--- a/src/redprl/refiner.fun
+++ b/src/redprl/refiner.fun
@@ -398,7 +398,6 @@ struct
         val motivem = substVar (m, x) motiveHole
 
         val (rewrittenGoal, rewrittenHole) = makeTrue (I, H) (motiven, l, k)
-        (* favonia: this should be enough... is it? *)
         val motiveMatchesMainGoal = makeEqTypeIfDifferent (I, H) ((motivem, mainGoal), l, k)
       in
         |>: motiveGoal >: rewrittenGoal >: motiveWfGoal >:? motiveMatchesMainGoal #> (I, H, rewrittenHole)
--- a/src/redprl/refiner.fun
+++ b/src/redprl/refiner.fun
@@ -96,8 +96,8 @@ struct
     fun Symmetry _ jdg =
       let
         val _ = RedPrlLog.trace "Equality.Symmetry"
-        val (I, H) >> CJ.EQ_TYPE ((ty1, ty2), k) = jdg
-        val goal = makeEqType (I, H) ((ty2, ty1), k)
+        val (I, H) >> CJ.EQ_TYPE ((ty1, ty2), l, k) = jdg
+        val goal = makeEqType (I, H) ((ty2, ty1), l, k)
       in
         |>: goal #> (I, H, trivial)
       end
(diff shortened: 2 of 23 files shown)


favonia:
@jonsterling I encountered an annoying issue and am not sure what to do: I want to use normalized levels in categorical judgments, and so I wrote the conversion functions between the normalized levels and `Sym.t P.term`. This works great except that parsing something into `Signature.src_catjdg` needs the conversion functions from `string P.term` to (normalized) levels. It is annoying to support both `Sym.t` and `string` at the same time because a normalized level consist of a dictionary indexed by `Sym.t` and `string`, respectively, and thus a different key type means a different structure for comparing keys. I wonder if you have any suggestion about this? My first idea is to check whether `Signature.src_catjdg` really needs to be `(string, ast) redprl_catjdg`---if not, then we can also remove the ugly hack `fromAst` in `RedPrlCategoricalJudgment`.

jonsterling:
@favonia sadly the source version needs to be ast and not abt, because we do name resolution as a second phase *after* parsing.

favonia:
@jonsterling If I turn `RedPrlLevel` into a functor which takes a `Key` structure, that is, the `ast` version uses `RedPrlLevel (StringOrder)` or something like that and the `abt` version uses `RedPrlLevel (Sym.Ord)`. Do I have to worry about the conversion from the levels using strings to the levels using variables?

jonsterling:
@favonia I think you will have to convert these, since the functors are generative. But it's possible I'm misunderstanding.

favonia:
@jonsterling Hmm. Do you know where `(string, ast) RedPrlCategoricalJudgment.jdg` will be converted into `(Sym.t, abt) RedPrlCategoricalJudgment.jdg`? I would like to take a look before rewriting everything.

jonsterling:
I'm curious, what is the omega level for, and what does a judgment mean (in terms of the semantics) that uses it?

I know that Agda has an omega level, which I think is used in order to give a type to `(i : Level) -> Set i`.

favonia:
> I'm curious, what is the omega level for, and what does a judgment mean (in terms of the semantics) that uses it?

It indicates the level of the type theory we are actually working in. It means the type was defined at some (finite) level but we don't care where that happened.

> I know that Agda has an omega level, which I think is used in order to give a type to (i : Level) -> Set i.

It does not have such things yet unless @cangiuli goes wild...

jonsterling:
> It indicates the level of the type theory we are actually working in. It means the type was defined at some (finite) level but we don't care where that happened.

Ah, this is a great idea! It will enable us to simplify RedPRL's proof theory quite a bit, I think. For the benefit of @cangiuli, I will now clarify that in the context of Favonia's implementation, `Uω` is a "façon de parler".

favonia:
By the way, eventually, I think we still need the user to input the (finite) level somewhere till we have multiple outputs (for level inference).

jonsterling:
> By the way, eventually, I think we still need the user to input the (finite) level somewhere till we have multiple outputs (for level inference).

Yes, I agree; feel free to do the simplest possible thing (even if that means requiring the user to enter this stuff in), and we can improve it incrementally through a couple of different-but-complementary avenues:

1. Supporting multiple outputs, to enable level synthesis *in the proof theory*
2. Developing heuristic tactics to fill in unknown levels

But each of these can be pursued after we have the basis universe theory set up.

favonia:
Again this PR accumulated several important changes and fixes that should be merged first.

src/redprl/refiner.fun
```diff
@@ -380,39 +369,39 @@ struct
...
         val motiven = substVar (n, x) motiveHole
         val motivem = substVar (m, x) motiveHole
 
-        val (rewrittenGoal, rewrittenHole) = makeTrue (I, H) (motiven, K.top)
-
-        val motiveWfGoal = makeType (I, Hx) (motiveHole, K.top)
-        val motiveMatchesMainGoal = makeEqType (I, H) ((motivem, mainGoal), k)
+        val (rewrittenGoal, rewrittenHole) = makeTrue (I, H) (motiven, l, k)
+        (* favonia: this should be enough... is it? *)
+        val motiveMatchesMainGoal = makeEqTypeIfDifferent (I, H) ((motivem, mainGoal), l, k)
```
favonia: Is this enough?

src/redprl/signature.sml
```diff
@@ -241,14 +241,18 @@ struct
...
-      end
+      AstToAbt.convertOpen (metactx, metactxToNameEnv metactx) (env, env) (ast, Sort.EXP)
+
+    (* these should be included in the AstToAbt module *)
+    fun elabSym env (name : string) : Sym.t = NameEnv.lookup env name handle _ => Sym.named name
+
+    (* these should be included in the AstToAbt module *)
+    fun elabParam env = RedPrlParameterTerm.map (elabSym env)
+
+    fun elabLevel env = Option.map (RedPrlLevel.out o elabParam env o RedPrlAstLevel.into)
```
favonia: Not completely sure about `elabLevel` and `elabParam`... `symctx` and `varctx` are not used at all.

src/redprl/refiner.fun
jonsterling: I believe that this suffices, as a consequence of `motiveWfGoal`.

src/redprl/machine.fun
```diff
@@ -518,6 +518,11 @@ struct
            end
          | _ => raise Fail "Impossible record type")
 
+     | O.POLY (O.UNIVERSE _) $ _ || (_, []) => raise Final
+     | O.POLY (O.UNIVERSE _) $ _ || (syms, HCOM (dir, HOLE, cap, tubes) :: stk) =>
+         E.raiseError (E.UNIMPLEMENTED (Fpp.text "hcom operations of universes"))
```
jonsterling: Nice! Love the new error stuff.

favonia:
Thanks. I'll merge this tonight.

src/redprl/refiner.fun
favonia: The thing is that I'm not 300% sure if levels and kindhood all (eventually at the omega level) respect equality. They should, and if they don't, I think it's a bug. @cangiuli

cangiuli: I believe this is true but I have not proven it.

PR #328, opened 2017-08-16
Refactor categorical judgments and sequents.

favonia:
@jonsterling I would like to refactor categorical judgments first before refactoring levels in the same way. The idea is that everything should be (1) fully polymorphic, or (2) specialized to abt with resolved names, or (3) specialized to ast with strings. Nothing in the middle. See https://github.com/RedPRL/sml-redprl/blob/6b38bba850b74bc75dde831db443632fe1b8f353/src/redprl/categorical_judgment.sig for the style I am going for.

### DIFF
--- a/src/redprl/signature.sml
+++ b/src/redprl/signature.sml
@@ -406,7 +406,7 @@ struct
                       val taux = CJ.synthesis jdgx
                     in
                       (ps,
-                       Tm.Sym.Ctx.insert ctx x taux,
+                       Tm.Sym.Ctx.insert ctx x RedPrlSortData.HYP,
                        NameEnv.insert env (Sym.toString x) x)
                     end)
                   (params', symctx, env)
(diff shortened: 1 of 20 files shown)


jonsterling:
sounds good!

favonia:
@jonsterling Would you mind if I specialize (de-functorize) `SEQUENT` too? Basically replacing `CJ.Tm` with `RedPrlAbt`.

jonsterling:
@favonia Yes, feel free to do whatever you think is best!

favonia:
Sorry for the delay.

src/redprl/signature.sml
```diff
@@ -406,7 +406,7 @@ struct
                       val taux = CJ.synthesis jdgx
                     in
                       (ps,
-                       Tm.Sym.Ctx.insert ctx x RedPrlSortData.HYP,
+                       Tm.Sym.Ctx.insert ctx x taux,
```
favonia: I am not 100% sure about this...

jonsterling: This change doesn't seem right; but this is a tricky part of the code.

favonia: Yeah... it does not even compile. However, what's the purpose of `taux`?

jonsterling:
I'll come back to this tomorrow—unfortunately I'm busy all day today.

favonia:
> I'll come back to this tomorrow—unfortunately I'm busy all day today.

That's fine. I believe this is independent of what I'm going to do next.

Issue #330, opened 2017-08-27
Kan operations of composition and V types


favonia:
0. ~Fix various stuff to make universes really work (except their Kan operations).~
1. ~`MEET_WITH_COMPLEMENT` and unify kinds and levels more.~
2. `fcom`
    a. ~hcom for universes.~
    b. ~`Restriction` combined with `makeEq`, and conditional goal making in cap/tube coherence subgoals.~
    c. ~Heterogeneous formal composition delegate.~
    d. hcom and coe operators for `fcom`.
3. ~Only true judgments can be hypotheses. #319~

jonsterling:
This all sounds fairly reasonable, but why are the following part of the tasks for building universes?

> Turning kinds, levels and dimensions...

> Only true judgments can be hypotheses...

Aren't these orthogonal?

favonia:
I guess I don't want to implement Universe.Eliminate and then remove it. The expression stuff is again to avoid working hard with param_exp and then dropping it. Let me change the first item to "wait" though.

jonsterling:
@favonia OK! By the way, feel free to do things in whatever order you feel is best, and if you feel it will be better to make all those changes in this order, then please do so. I trust your judgment.

favonia:
@jonsterling Actually, you were right. I should not wait for #301 as there are still some design issues left.

jonsterling:
Are the operational semantics for `hcom` and `coe` on `V` types written down in a note somewhere?

I was going to try out an example like:

```

Thm NotV/test2 : [
  (= bool (coe 0~>1 [i] (NotV i) tt) ff)
] by [
  ?
].
```

favonia:
@jonsterling You should ask @cangiuli about the latest paper version. I did not implement them because we sort of agreed to finish other parts first, and that "other parts" in my opinion are not done yet. Otherwise I can finish the implementation in no time.

jonsterling:
@favonia OK! There's no need to rush, I was just curious.

cangiuli:
Feel free to implement them. I think all of us (and also Anders!) would like to see them in action.

PR #331, opened 2017-08-28
Universes, stage 2.

favonia:
Follow the path of #330.


favonia:
@jonsterling @cangiuli I could not really decide in `FormalComposition.True` what we should do. For the equation `0=1`, should we...

1. just put `AX` as the term, or
2. ask the user to input some garbage and skip the type checking?

Then, for the equation `x=1`, should we

1. ask the user to fill in something of type `B<1/x>` with access to `x`, or
2. ask the user to fill in something of type `B<1/x>` without?

jonsterling:
@favonia I am not sure that we should even have a `FormalComposition.True`; users should be constructing hcoms, which will turn out semantically to be formal compositions. But I think `fcom` is too low a level of abstraction for the PRL.

In general, you should feel free to put in your own garbage rather than asking the user to put in some garbage. I'm not sure about the `x=1` question, though.

It is probably fine for now, btw, to just require the user to write the term whole term rather than have this rule. Ultimately I would prefer to have a refinement-based solution, but if that seems too difficult, I wouldn't stress about it.

favonia:
@jonsterling The `FormalComposition.True` rule is constructing a `BOX` term, which is probably not derivable from `hocm`? Anyways, it's easier to put in `AX` for `0=1`. I think I'll ban the access to `x`...

jonsterling:
@favonia Ah, I guess I wasn't up to date with what that rule should do. Feel free to do whatever you think is best, and we can try it out in practice.

favonia:
@jonsterling @cangiuli Again I'm not sure of which should be _the_ left rule of `fcom`. Guess I'll skip it for the moment.

jonsterling:
@favonia I would not worry about being exhaustive with the rules. I think the best thing is to write the minimal sensible thing, and we can add more rules in response to practical needs. You're doing great!

src/redprl/machine.fun
```diff
@@ -154,10 +148,23 @@ struct
...
+                   P.VAR _ => false
+                 | P.APP P.DIM0 => true
+                 | P.APP P.DIM1 => true
+                 | P.APP _ => E.raiseError (E.INVALID_DIMENSION (TermPrinter.ppParam r))
+            in
+              if isBound syms r1 orelse isBound syms r2 orelse (isConstant r1 andalso isConstant r2) then
+                false
+              else
+                raise Unstable
+            end
```
favonia: The previous code allowed `x=1` where `x` is not bound.

jonsterling: @favonia Great catch!!

favonia:
Hmm. The following example fails because of some design flaws in the head expansion rules. I'm wrapping up this PR...
```
Thm Fcom/Reduce2 : [
  tt in (fcom{0~>1} bool [0=0 {j} bool])
] by [
  auto
].
```

jonsterling:
OK! I'll review this today so we can get it merged. Great work!

When you get a chance, maybe you can tell me a bit about that head expansion issue; no rush.

src/redprl/refiner_types.fun
```diff
@@ -1268,6 +1270,176 @@ struct
       end
   end
 
+  structure FormalComposition =
+  struct
+    val kindConstraintOnCapAndTubes =
+      fn K.DISCRETE => (K.DISCRETE, K.DISCRETE)
+       | K.KAN => (K.KAN, K.KAN) (* XXX more research needed *)
```
jonsterling: Does `more research needed` mean that these conditions are sufficient but maybe not necessary? Or the other way around?

src/redprl/refiner_types.fun
```diff
@@ -1321,5 +1493,24 @@ struct
       in
         |>: goalTy >: goalTy' #> (I, H, Syn.into Syn.AX)
       end
+
+    (* This rule will be removed once every hypothesis
+     * is required to be `A true`. *)
```
jonsterling: Good.

src/redprl/refiner_types.fun
favonia: I hope it's the former, but in some unfortunate cases it could be the latter... the fcom types are so complex that I am not sure about anything. `K.KAN => (K.KAN, K.KAN)` should be correct though.

jonsterling: OK! We can improve and correct if needed in the future as we begin to understand these things in a more precise way.

PR #334, opened 2017-08-31
Improve head expansion

favonia:
@jonsterling 

This PR fixes these:

0. We forgot to head-expand the type in `EQ`.
1. We forgot to unfold when the RHS is `CUST`.

It also addresses these:

0. Prevent random side-switching during head-expansion.
1. The tactic `head-expand` really tries all positions.

There is still a fundamental flaw in the refiner:
```sml
        case (Syn.out m, blocker1, Syn.out n, blocker2) of
...
         | (Syn.IF _, Machine.VAR z, _, _) => Bool.Elim z
         | (_, _, Syn.IF _, Machine.VAR z) => Bool.Elim z
```
The machine should tell us the innermost frame (if any) and then the refiner checks whether it is the `IF` frame. Right now we are looking at the outermost operator, which is not accurate.

### DIFF
--- a/src/redprl/refiner.fun
+++ b/src/redprl/refiner.fun
@@ -651,18 +651,20 @@ struct
   structure Computation =
   struct
     open Computation
-    fun TryEqHeadExpansionTy sign alpha = Lcf.try (EqHeadExpansionTy sign alpha)
-    fun TryEqHeadExpansionL sign alpha = Lcf.try (EqHeadExpansionL sign alpha)
-    fun TryEqHeadExpansionR sign alpha = Lcf.try (EqHeadExpansionR sign alpha)
+    fun TryEqHeadExpansionTy sign alpha = Lcf.try @@ EqHeadExpansionTy sign alpha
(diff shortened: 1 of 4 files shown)


jonsterling:
@favonia Nice work!

And you're right about that flaw, which looks at the wrong frame... We can fix that either here or in a separate PR, whatever you like.

jonsterling:
Re: that heuristic that is checking the wrong frame, maybe it would be simpler to not check any frame, and instead just `elim` that variable regardless of its type. Or if there are some cases where that's not a good idea, we could filter those out by looking at the type.

Just an idea! Not sure if it's a good one. In any case, once Travis passes, feel free to merge this.

favonia:
@jonsterling Fixing that flaw must change the machine as well... What's your opinion on exposing the frame? `VAR of RedPrlAbt.variable * Neutral frame option`?

favonia:
Oh, right, we can look up the type of the variable in the context and try to do something...

PR #335, opened 2017-08-31
Improve Elim in various ways.

favonia:


favonia:
Hmm got another idea.

favonia:
@jonsterling @cangiuli Some difficult questions: what should we check or do when `(hcom{r~>r}...)`, `(hcom{..} ... [r=r foo] ...)`, `(coe{r~>r} ...)` or `(@ x 1)` appears as `ty` in `EQ ((m,n), ty)`? What rules are provable and what's the Immortal order to try them? My current naive idea is to...

(1) try to evaluate `ty` first, and then
(2) try to apply "safe and useful" elims (`bool`, `void`, `record` and `=`) to the blockers of `ty` and restart the process, or
(3) try to apply rules to solve  `(hcom{r~>r}...)`, `(hcom{..} ... [r=r foo] ...)` or `(coe{r~>r} ...)` and restart this process, or
(4) try to solve `(@ x 1)`, or
(5) use the current heuristic for `EQ`.

I am not sure how important (3) is because in "normal" usage users should already specify the type (which should be one of the universe) and the step (1) will already work. However, (4) seems unavoidable, unless we want the user to manually prove some equality and then rewrite.

favonia:
@jonsterling How about a `simplify` tactic which will try to apply the following rules to things on the path to the blocker?
1. the cap rule to simplify `(hcom{r~>r}...)`
2. the tube rule to simplify `(hcom{..} ... [r=r foo] ...)`
3. the cap rule to simplify `(coe{r~>r} ...)`
4. the app-const rule to simplify `(@ x 1)`

favonia:
This PR does not address the above difficult question (moved to #336) but already greatly increased the power of `Elim` rules.

jonsterling:
Excellent! I'll finish reviewing this shortly.

jonsterling:
> How about a simplify tactic which will try to apply the following rules to things on the path to the blocker?

That sounds cool!

PR #337, opened 2017-09-01
More aggressively skip useless subgoals of the *COM family.

favonia:
Use `Restriction.make*IfDifferent` in more settings.

### DIFF
--- a/src/redprl/listutil.sml
+++ b/src/redprl/listutil.sml
@@ -39,3 +39,14 @@ struct
     fun init l = init' [] l
   end
 end
+
+structure ListPairUtil =
+struct
+  fun mapPartialEq f =
+    ListPair.foldrEq
+      (fn (x1, x2, ys) =>
+        case f (x1, x2) of
+          NONE => ys
+        | SOME y => y :: ys)
+      []
+end
--- a/src/redprl/refiner_composition_kit.fun
+++ b/src/redprl/refiner_composition_kit.fun
@@ -102,34 +102,46 @@ struct
            N_i = P_j in A [Psi, y | r_i = r_i', r_j = r_j']
      *)
     fun alphaRenameTubes w = List.map (fn (eq, (u, tube)) => (eq, substSymbol (P.ret w, u) tube))
-    fun genInterTubeGoals (I, H) w ((tubes0, tubes1), (ty, l, k)) =
+    fun enumInterExceptDiag f =
       let
-        val tubes0 = alphaRenameTubes w tubes0
-        val tubes1 = alphaRenameTubes w tubes1
-
-        fun interTube (eq0, tube0) (eq1, tube1) =
-          Restriction.makeEq [eq0, eq1] (I @ [(w,P.DIM)], H) ((tube0, tube1), (ty, l, k))
-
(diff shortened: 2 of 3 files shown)


favonia:
I am thinking that maybe in the future a rule can take an argument indicating whether even the number of auxiliary goals should be the same under arbitrary substitution. It's relatively easy to revert the optimization.

jonsterling:
> I am thinking that maybe in the future a rule can take an argument indicating whether even the number of auxiliary goals should be the same under arbitrary substitutions.

That's an interesting idea, but I don't think it's what I want to do... It continues to make things more complicated. I would like the rules to be very simple and not do sophisticated things; it would be better to produce *all* the subgoals, and then discharge these automatically using tactics. There is some performance overhead, but I have a feeling it won't be prohibitive.

But I'm not trying to make that change now; we can look at it later, after the important stuff is done.

favonia:
Alright, but such a tactic will benefit from a slightly more powerful LF I think---the ability to search across different subgoals and find redundancy. For example, if there are two identical subgoals, we only have to prove one of them.

jonsterling:
> For example, if there are two identical subgoals, we only have to prove one of them.

That's a great idea! I think we can support something like that if we think carefully. The only question will be whether it is in practice actually less expensive to do this searching than it would be to prove the thing `n` times (if you have 300 subgoals, which happens easily, it might be very expensive to hunt for duplicates).

favonia:
Is it easy to support searching within the subgoals generated by a particular rule?

jonsterling:
@favonia Subgoals don't carry any information about what rule they came from, so that's not easy. Ultimately, I think it is best (as a general principle, not necessarily in each particular case) if automation tactics only look at the shape of goals, and not try to do fancy things depending on what rule was used, etc.

favonia:
I meant a "rule modifier" in a tactic that takes the subgoals of a rule before merging the telescopes into one.

jonsterling:
@favonia Ah! Do you mean just like "apply this rule/tactic, and then let me do something with the subgoals generated by it"?

If so, then yes, that's what "multitactics" do. We only have a few multitactics at this time: one applies another tactic to all the subgoals, another applies a list of tactics pointwise to the subgoals, etc.

favonia:
Yes, and I think we can have a very sophisticated multitactic to mimic most of the `IfDifferent` and `IfLess` optimizations. It would probably be more expensive, though, as you predicted. Also, some invariants (like if a function type `(-> A B)` is in universe `i` then `A` is also in universe `i`) are probably not provable without another rule.

jonsterling:
My thought is that at some point (not now, I don't want to distract from the important stuff), we can try turning off the optimizations. If performance is still not horrible, there's no need to immediately work on such a multitactic; but if performance becomes really bad, we could think about the design of such a multitactic.

favonia:
As a record I think this PR reduces ~10% of the goals according to the experiments on `connection.prl`.

jonsterling:
That's quite impressive! Feel free to merge, btw---the convo about rule optimizations in general is separate from the day-to-day issues, on which I follow your judgment completely.

jonsterling:
Agreed.

PR #338, opened 2017-09-01
Univalence, stage 0.

favonia:


favonia:
@cangiuli @jonsterling It is time to propose your favorite definition of equivalences (as a labeled record).

favonia:
@jonsterling If you already proved some interesting equivalence somewhere please show me! This PR is done if I can verify it with some interesting examples, but writing them will take some time...

favonia:
Hmm `path` is a keyword so the above rule is almost never applicable. Oops.

src/redprl/machine.fun
```diff
@@ -140,6 +142,22 @@ struct
...
 
+  local
+    fun assertVariable stability syms u =
+      case stability of
+        NOMINAL => ()
+      | CUBICAL =>
+          if SymSet.member syms u then ()
+          else raise Unstable
+  in
+    fun branchOnDim stability syms r f0 f1 fu =
```
jonsterling: nice combinator

favonia:
Apparently I do not know how to comment on one commit. I meant the commit 097dac7a75a743ff11523c8af26b87da0039cc4e.

test/success/ua.prl
```diff
@@ -0,0 +1,49 @@
...
+        lam c, c'. <i>
+	  `(tuple
+	    (hcom{1~>0} #A ,a [i=0 {j} (@ (! proj2 ,c) j)] [i=1 {j} (@ (! proj2 ,c') j)])
+	    (abs {j} (hcom{1~>j} #A ,a [i=0 {j} (@ (! proj2 ,c) j)] [i=1 {j} (@ (! proj2 ,c') j)])))
+      , auto
+      ]
+    ]
+  , auto
+  ]
+].
```
jonsterling: This is pretty cool!!

By the way, a higher-level/programming-style version of this script could be the following:

```

Thm IdIsEquiv (#A) : [
  #A hcom type
  >>
  (IsEquiv #A #A Id)
] by [
  unfold IsEquiv; unfold IsContr; unfold Fiber; unfold HasAllPaths;
  lam a.
  < proj1 =
     <proj1 = use a,
      proj2 = <_> use a>
  , proj2 =
     lam <proj2 = c>, <proj2 = c'>. <i>
       < proj1 =
          `(hcom{1~>0} #A ,a
            [i=0 {j} (@ ,c j)]
            [i=1 {j} (@ ,c' j)])
       , proj2 =
           <j> `(hcom{1~>j} #A ,a
                 [i=0 {j} (@ ,c j)]
                 [i=1 {j} (@ ,c' j)])
       >
  >
].
```

favonia: Can we make `unfold` take multiple arguments, by the way?

jonsterling: @favonia I was just grimacing over that myself 😄 I think we should definitely do that.

favonia:
Hmm something went wrong after rebasing. Now
```
Thm IdEquiv (#A) : [
  #A hcom type
  >>
  (Equiv #A #A)
] by [
  `(tuple Id (IdIsEquiv #A));
  unfold IdIsEquiv;
  ?
].
```
gives
```
Fail: Impossible failure during safeUnfold: nonexhaustive binding failure
```
but I recall it was fine, and travis seems to agree with me: https://travis-ci.org/RedPRL/sml-redprl/jobs/271302400.

jonsterling:
@favonia I wonder if I introduced this in my recent bugfixes to operator unfolding...

favonia:
@jonsterling Should I leave this to you? I can switch to the category example.

jonsterling:
@favonia Yes, I can have a look at the operator unfolding issue; it may take me a day or two though, I have a lot on my plate.

jonsterling:
@favonia Ah, I see what happened; there's a change in behavior which causes this kind of definition (where you put a sequent in the goal) not to behave in the way it used to. There's a bug in my fix, but fixing the bug won't fix this code.

I'll try and explain in more detail, and fix everything.

favonia:
To be honest, I was actually surprised that it worked with a suspicious hypothesis hanging around like that. In some way I am glad it failed, if that is the reason it failed.

jonsterling:
@favonia I fixed that bug, and committed it directly to this branch (hope you don't mind).

Going forward, I think we should avoid using sequents at the top level, since the resulting theorems cannot be easily re-used. We should try to use function types and universes wherever possible.

jonsterling:
(and equality types!)

favonia:
Maybe it is obvious, but should I merge this PR now?

jonsterling:
@favonia Yes please! Great work, this is a very exciting moment!

PR #349, opened 2017-09-03
in examples, work internally using universes rather than sequents

jonsterling:


test/examples.prl
```diff
@@ -192,31 +192,32 @@ Thm Not : [(-> [_ : wbool] wbool)] by [
...
-Tac FunExtTac(#A; #B; #F; #G) = [
-  let p : [(-> [x : #A] (path {_} wbool ($ #F x) ($ #G x)))] = id.
-  <i> lam a. use p [use a, @i]
+Tac FunExtTac{l:lvl} = [
+  query gl <- goal.
+  match [gl:jdg] {
+    [a, b, f, g | #jdg{(path {_} (-> %a %b) %f %g) true} =>
+      use {FunExt l}
+        [ `%a, `%b, `%f, `%g, id ]]
+  }
```
jonsterling: Here's an example of something that would be cool to generalize and have as a built-in tactic, rather than having to write it manually in each case.

We're using unification to fill in the arguments to the lemma.

favonia: Do you mean something like `use {FunExt l} [_, _, _, _, id]`? That is, unification everywhere?

jonsterling: Right. I think that's possible to implement, but the code will be a little tricky.

Issue #357, opened 2017-09-04
Suppress unused variables when pretty printing quantifiers


favonia:
So `(-> [_ : nat] nat)` would be printed as `(-> nat nat)`.

favonia:
@jonsterling What is the most efficient way to check occurrence?

jonsterling:
@favonia I think that using `varctx` should be OK, since IIRC this is "preconstructed" when terms are built.

Issue #333, opened 2017-08-30
Avoid invoking rules that obviously require further user inputs in the `auto` tactic.


favonia:

favonia:
29547690a56f4ed48598e5ff1eba1ed287cee2dd is not good enough. In the cases `APP` and `PATH_APP`, only when we can easily synthesize the function types or the path types should `dfun/eq/app` or `path/eq/app` be used.

jonsterling:
@favonia This feels like a different kind of case though... The task of synthesizing the type of some program (`SYNTH R`) is well-defined even if out-of-context: so if `auto` gets stuck and the user has to provide an explicit type there, then that is maybe not so bad. Whereas for things like positive elim rules, the goal that is generated is just something like `TERM EXP` or whatever, and we need to ensure that `auto` never generates subgoals like this, since you really can't guess what to do with that.

Does what I'm saying make any sense? I'm sorry that I'm not phrasing it very clearly.

favonia:
Hmm maybe after #359 it would be more manageable. Right now it is often the case that I have to come up with some type for a gigantic term, and those terms heavily use `void/elim` which gives no type information at all.

jonsterling:
@favonia Ah, I see! Interesting.

In Nuprl, by the way, they actually attach type annotations internally to subterms and these help drive things like inference heuristics. I'd like to avoid this as much as possible, but something along those lines may become necessary ultimately. It's hard to tell.

jonsterling:
I agree that it seems like if we implement the granular unfolding rules that you suggested, we can avoid some of this problem... Usually if you have to prove some typing lemma about a very big term all at once, it means that things were unfolded too early.

favonia:
Hmm this is still not good enough unless we support very fancy fixed point unfolding like this
```
  ($ Plus (S n) m) ~~> (succ ($ Plus n m))
```
unlike the current unfolding
```
  ($ Plus (S n) m) ~~> ($ (........very long code......) n m)
```
where you might have to prove well-formed over and over again. Before we have this fancy unfolding, the reason to hold `fun/app` and/or `path/app` is that we can infer more info from `TRUE` or `EQ` judgments than that from `SYNTH` ones. It seems easier to write a tactic in the user land to discharge the non-dependent cases.

jonsterling:
@favonia Indeed; This is something that you would probably need to just prove as a lemma.

PR #358, opened 2017-09-04
Semi-simplicial types and everything that makes them possible, stage 0.

favonia:


favonia:
To be clear, `internalize` is derivable and probably should be removed later.

src/redprl/refiner.fun
```diff
@@ -498,80 +498,48 @@ struct
...
+        val _ = RedPrlLog.trace "Computation.SequentHeadExpansion"
+        val (I, H) >> catjdg = jdg
+        val (goal, hole) = makeGoal @@ Seq.map (eval sign) jdg
       in
-        HeadExpansionDelegate sign (I, H) maker b
+        |>: goal #> (I, H, hole)
       end
 
-    fun TrueHeadExpansion sign _ jdg =
+    fun MatchHeadExpansion sign _ jdg =
```
jonsterling: I think we should be cautious about having head expansion for the "match" judgments. I'm not completely sure of what there semantics would be if we have such a rule.

favonia: Hmm... are you worried about some computation turning something matched into something unmatched?

favonia: Anyways I removed it.

jonsterling: Yeah, something like that... It might be OK, but it's hard for me to say what the meaning of the judgment is if we have this rule—I envisioned it as a very syntactic thing.

Preferably, this computation would be done *before* the match subgoal was generated (somehow).

favonia:
I would like to improve RedPRL ( #333, #357, #359, etc)  further before proving the associativity lemma.

jonsterling:
sounds reasonable!

It's great that we can try out some real examples now, because those are telling us what parts of RedPRL most urgently need to be improved.

favonia:
I managed to have RedPRL generate ~60 subgoals after >1 second, and that sounds like a problem to me.

jonsterling:
@favonia Ah, indeed... Yeah, I kind of expected that ultimately something along those lines would happen fairly frequently—this is the essential contradiction of non-"formal type theory": you can prove hard things, but often the 'easy' things become very hard!

We will have to see how much we can mitigate this.

test/success/semi-simplicial.prl
```diff
@@ -61,20 +61,16 @@ Thm Pick/compose : [
     [ lam c, p1, p2. elim p2
     , lam c. fresh c', c'/ih -> elim c;
       [ lam p1. elim p1
-      , lam p1. unfold Pick in p1; reduce in p1;
-        let < head = p1/h, tail = p1/t > = p1. elim p1/h; reduce in p1/t;
-        [ lam p2. unfold Pick in p2; reduce in p2;
-          let < head = p2/h, tail = p2/t > = p2. elim p2/h; reduce in p2/t;
-          unfold Pick in goal; reduce in goal;
+      , lam p1. let < head = p1/h, tail = p1/t > = p1. elim p1/h;
+        [ lam p2. let < head = p2/h, tail = p2/t > = p2. elim p2/h;
```
favonia: Maybe it is even better to run the pattern matching without leaving the type of `p1` expanded?

favonia:
(Too many changes and new toys to summarize.)

test/success/semi-simplicial.prl
```diff
@@ -71,14 +108,12 @@ Thm Pick/compose : [
...
           < head = `ff
           , tail = `($ ,c'/ih ,p1/t ,p2)
           >
         ]
       ]
     ]
-  ];
-
-  // one line to solve all well-formedness goals
-  At0; <| <| EqNatRec; internalize |> || id |>; NatArg; EqNatRec; EqNatRec; internalize; auto
```
favonia: No need to prove well-formedness if you gave RedPRL enough hints.

test/success/semi-simplicial.prl
```diff
@@ -70,8 +70,7 @@ Thm Pick/succ : [
 ] by [
   lam n, m. auto;
 
-  [ NatArg; EqNatRec; EqNatRec ];
-  [ EqNatRec; NatArg; EqNatRec; EqNatRec ]
+  repeat { NatArg || EqNatRec }
```
favonia: We still need to prove well-formedness of the definition of `Pick` because we have to unfold `Pick`. This is related to #333.

test/success/semi-simplicial.prl
```diff
@@ -0,0 +1,118 @@
...
+          ]
+        , lam p2.
+          (ReplaceGoal ($ Pick/succ ,a' ,c'));
+          < head = `ff
+          , tail = `($ ,c'/ih ,p1/t ,p2)
+          >
+        ]
+      ]
+    ]
+  ]
```
jonsterling: very cool that there are no further subgoals here. nicely done.

PR #372, opened 2017-09-15
Semi-simplicial types and friends, stage 1.

favonia:


test/success/semi-simplicial.prl
```diff
@@ -48,29 +50,30 @@ Thm Pick : [
...
      [head : bool]
      [tail : (if head ($ Pick n m) ($ Pick (succ n) m))])))
 ] by [
-  lam n, m. auto;
-
-  repeat { NatArg || EqNatRec }
+  lam n, m.
+    let aux0 : [($ Pick ,n ,m) type at 0] = auto.
+    let aux1 : [($ Pick (succ ,n) ,m) type at 0] = auto.
+    unfold Pick; reduce; auto
```
jonsterling: Nice. This seems like it would be a more efficient proof too.

src/redprl/redprl.grm
```diff
@@ -970,7 +970,7 @@ declArgument
 
 declArguments
   : declArgument ([declArgument])
-  | declArgument SEMI declArguments (declArgument :: declArguments)
+  | declArgument COMMA declArguments (declArgument :: declArguments)
```
jonsterling: good, I think this is less confusing anyway.

At some point I'd like to come up with a notation for declarations that looks more like what the "redprl sexprs" look like, but I haven't been able to come up with anything good yet.

favonia:
@jonsterling Is this conflicting with your new work? I was thinking about merging this first so that I can start other branches in a more systematic way.

jonsterling:
@favonia Let's get your stuff merged first, yes; then I can address whatever conflicts occur. 👍

PR #378, opened 2017-09-15
delete a dangerous / easily-misused function

jonsterling:


jonsterling:
My incorrect use of this function a while ago was the source of some very mysterious bugs that only surfaced when I fixed some other behavior! Those uses were fixed in a previous PR, but I want to delete this function so nobody makes my mistake again.

favonia:
It seems what `Extract` can achieve now is covered by `Print`. Is it correct?

jonsterling:
@favonia The only difference is that `Extract` just prints out the extract term, whereas `Print` prints out a full elaborated declaration (including the name of an operator, its arguments, etc.).

PR #386, opened 2017-09-25
Fix scoping issues

### LINKED ISSUE #264: Paramater sort checking
This should be rejected, because `i` is of the wrong sort. However, it seems that the current parameter sort checking is not taking a parameter context...?
```
Thm Cap{i : exn}(#A) : [ 
  a/type : #A type,
  x : #A
  >> (hcom{0 ~> 0} #A x [i=0 {_} x] [i=1 {_} x]) = x in #A
] by [
  auto
].
```

jonsterling:
Resolves #264


favonia:
Yay!!!

jonsterling:
@favonia Sorry it took me so long to get around to it! 😄

favonia:
It surprised me a little that we accumulated so many tests violating the scopes. Good that we catch them now!

jonsterling:
@favonia Indeed! It is hard to see them with your eyes, and often they pass for the wrong reason

PR #387, opened 2017-09-26
simplify & monomorphize representation of judgments

jonsterling:
In my last PR, I made it so that we don't need a separate "source representation" of sequents (we just use ASTs). This means that we can simplify and monomorphize the judgments modules.


jonsterling:
I also did some renaming, "categorical" => "atomic".

Issue #301, opened 2017-07-29
Wild idea: remove notion of symbols and parameters altogether?


jonsterling:
RedPRL's syntactic framework is two-level: there is a theory of parameters, and then second-order operator signatures are indexed over this theory. Then, the second-order theory is generated from such a signature.

This makes the most sense in the semantics (especially for languages like those defined in PFPL, where parameter binding is substructural), but in the implementation of RedPRL, it has admittedly led to a lot of difficulties and duplicated work without much in return. **Moreover, since RedPRL's dimension theory is cartesian, binding of dimensions and of object variables has exactly the same meaning.**

(Here it is important to avoid a misconception that was promulgated in the discussions of CHiTT, where we said that "at the syntactic level, dimensions are treated as nominal atoms"; this statement really meant to refer to the *operational level*, not the syntactic level. The precise way to say it is that the syntax is cartesian, but that the operational semantics as a transition system enjoys naturality only with respect to permutation maps.)

What we gain from the implemented style is that we can have operators which take a name as a parameter, and this name participates in the parameter-bindings. Currently the places where this matters in a crucial way have nothing at all to do with dimensions; it has only to do with hypotheses in the proof theory, where it is nice to have "elim" take an honest-to-god hypothesis name as a *parameter* rather than take an argument and force it to be a variable.

One other nice thing that we get is that the operational semantics is completely driven by the shape of the head-operator, and each step is "parametric" in the exact materials that lie in the subtree of an operator. **I felt that this was very important from the start, but I wonder if it really matters for implementation.**

But I think that the above is not enough to get in return for the amount of effort we have put in. I would like to consider regarding dimensions as terms of sort `DIM`. We would change the syntax for things like `hyp` and `elim` tactics to take a term argument rather than a "name".

+ *This would also allow us to get rid of the `,x` stuff in terms that you supply to the refiner, which has always felt like a hack.*
+ We can also get rid of the `PARAM_EXP sigma` sort
+ We can get rid of the `PARAM_SUBST` "judgment" and just use ordinary substitution and metavariables.

Since RedPRL is the only client of the ABT library, we could remove support for symbols from it as well. This is a bit of a retreat, since that stuff would be useful for implementing stuff from PFPL using Bob's Thought, but I think it is unrealistic to expect that to happen at this point.

jonsterling:
This would also enable #300 to be accomplished more easily.

favonia:
[...]

PR #388, opened 2017-09-26
Level expressions

jonsterling:
Depends on #387. This leads to a lot of simplification; it also makes it possible to pattern match in tactic scripts on level expressions.

Related to #301

### DIFF
--- a/src/redprl/redprl.grm
+++ b/src/redprl/redprl.grm
@@ -286,6 +286,7 @@ end
  | RIGHT_ARROW | LEFT_ARROW
  | DOUBLE_RIGHT_ARROW | LONG_RIGHT_ARROW
  | TRIANGLE_RIGHT
+ | PLUS
 
  (* parameter sorts *)
  | DIM | HYP | META_NAME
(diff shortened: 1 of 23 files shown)


jonsterling:
This also fixes some bad bugs that I found along the way.

Still needed is to restore some of the fancier notation for level expressions.

jonsterling:
I want to do the same thing to dimensions, by the way.

jonsterling:
It will also be useful to do this to *kinds*, so that we can match on these in tactics.

favonia:
I wonder what the bad bugs are along the way?

src/redprl/redprl.grm
```diff
@@ -236,13 +236,21 @@ struct
 
   fun makeUseLemma (opid, params, subtermArgs) tacs = 
     let
-      val args = List.map (fn tac => ([],[]) \ tac) tacs
+      val args = subtermArgs @ List.map (fn tac => ([],[]) \ tac) tacs
```
jonsterling: @favonia This is one of the "bad bugs"

jonsterling: In fact, I can't remember, but this might be the only one.

PR #389, opened 2017-09-26
use {} rather than <> for record decomposition in tactics

jonsterling:
I didn't like the angle brackets; I felt they made things very hard to read. Open to better suggestions than this though.

### DIFF
--- a/src/redprl/redprl.grm
+++ b/src/redprl/redprl.grm
@@ -436,6 +436,7 @@ end
  | tactics of ast list
 
  | recordFieldTactics of (string * ast) list
+
  | devAppSpine of ast list
  | bracketedDevAppSpine of ast list
  | devDecompPattern of string O.dev_pattern
@@ -825,7 +826,7 @@ bracketedDevAppSpine
 
 devDecompPattern
   : VARNAME (O.PAT_VAR VARNAME)
-  | LANGLE tupleDecompPattern RANGLE (O.PAT_TUPLE tupleDecompPattern)
+  | LBRACKET tupleDecompPattern RBRACKET (O.PAT_TUPLE tupleDecompPattern)
 
 devDecompPatterns
   : devDecompPattern COMMA devDecompPatterns (devDecompPattern :: devDecompPatterns)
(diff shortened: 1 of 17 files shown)


favonia:
By the way, I have been wondering whether we could have abbreviation for sigma types. That is, some short notation involving `x` and `y` and `z` meaning `{proj1 = x, proj2 = y, proj3 = z}`. Will `{x, y, z}` make the grammar ambiguous?

jonsterling:
@favonia I wondered about that too; I would like to have a nice, positional notation for those too, but sadly `{x,y,z}` seems like it might make the grammar ambiguous. Now that I have started using braces for the named version, maybe we should use angle brackets for the anonymous version... So, like `<x,y,z>`, which I *think* would not be ambiguous.

jonsterling:
^ Ahh, this does make the grammar ambiguous (conflicts with dimension abstraction).

jonsterling:
Ok, I've pushed one  idea that is not ambiguous.

favonia:
I see. Ah, this is annoying. Perhaps we should spread the love of S-expressions to other parts sooner...

jonsterling:
> Perhaps we should spread the love of S-expressions to other parts sooner...

Ultimately we may want to do that; I think the "eclectic" notation for tactic scripts in RedPRL is kind of dumb, frankly. It's nice because it doesn't make you type in too many brackets and it is kind of pleasant to read, but it is really arbitrary in many places, and probably very difficult to learn.

Using the sexpr notation in the tactics may lead to one possible problem, is that it would become very difficult to distinguish the "program development notation" tactics from the programs themselves.

The other thing to consider is that when we eventually have the new metalanguage (which may *never* happen, or it may happen during this coming year, depending on how much time I have), probably it would be sensible to use an ML-style notation for that, and then the current tactic notation would disappear.

jonsterling:
The other nice thing about the idea of making RedPRL use our sexpr variant globally (even at the top level) is that it becomes really easy to provide excellent editor support to RedPRL (such as paredit). It's also basically impossible in practice to highly RedPRL's tactic language correctly, but it would be easy with sexprs.

PR #391, opened 2017-09-26
Treat kinds as expressions

jonsterling:
This lets us do unification and matching in tactics with kinds.

No support is yet available for "kind-polymorphic assertions": things just fail if a kind is a metavariable rather than a constant. That's ok for now, since that's a separate issue.

### DIFF
--- a/src/redprl/atomic_judgment.sml
+++ b/src/redprl/atomic_judgment.sml
@@ -84,17 +84,20 @@ struct
     structure O = RedPrlOpData
     infix $ $$ \
   in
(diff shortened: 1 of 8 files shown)


favonia:
The only case I can think of is when they are of the same variables, which is probably okay. I am reluctant to recognize any non-trivial derived relations between kind variables from the semantics. Not even that `K.top <= x`.

jonsterling:
@favonia Indeed; I favor leaving things as they are and not providing any fancy support for non-concrete kinds.

PR #392, opened 2017-10-03
Banish parameters

jonsterling:
This one's a nightmare of a PR... It's not done yet.

#301

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,9 +10,6 @@
 [submodule "lib/sml-dependent-lcf"]
 	path = lib/sml-dependent-lcf
 	url = https://github.com/JonPRL/sml-dependent-lcf.git
-[submodule "lib/sml-json"]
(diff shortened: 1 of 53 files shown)


jonsterling:
TODO: don't use symbols anymore to reference hypotheses. Instead use terms, and just ensure that these are actual variables at the use-site. It's a bit ugly, but it will simplify a lot of logic and fix a number of bugs.

jonsterling:
Wow! The merge conflict was minimal. I'm surprised!

jonsterling:
Now the main thing that remains is to restore the parsing for all these com-like gadgets.

jonsterling:
Now what remains is the following:

1. Go through and make sure the failure tests fail for the right reason
2. General cleanup of code

jonsterling:
I'm going to merge this now so that we can begin working on more stuff.

PR #397, opened 2017-10-05
Remove symbol binding

### LINKED ISSUE #301: Wild idea: remove notion of symbols and parameters altogether?

jonsterling:
Close #301

### DIFF
--- a/src/redprl/atomic_judgment.sml
+++ b/src/redprl/atomic_judgment.sml
@@ -82,41 +82,41 @@ struct
     infix $ $$ \
   in
     fun kconst k = 
-      O.MONO (O.KCONST k) $$ []
+      O.KCONST k $$ []
 
     val into : jdg -> abt =
-      fn EQ ((m, n), (a, SOME l, k)) => O.MONO (O.JDG_EQ true) $$ [([],[]) \ L.into l, ([],[]) \ kconst k, ([],[]) \ m, ([],[]) \ n, ([],[]) \ a]
-       | EQ ((m, n), (a, NONE, k)) => O.MONO (O.JDG_EQ false) $$ [([],[]) \ kconst k, ([],[]) \ m, ([],[]) \ n, ([],[]) \ a]
-       | TRUE (a, SOME l, k) => O.MONO (O.JDG_TRUE true) $$ [([],[]) \ L.into l, ([],[]) \ kconst k, ([],[]) \ a]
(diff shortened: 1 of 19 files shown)


jonsterling:
This also fixes some bugs I introduced recently.

PR #401, opened 2017-10-08
update dependencies

jonsterling:


jonsterling:
Jesus Christ, now SML/NJ build is broken for some reason. :(

jonsterling:
And it builds fine on my machine 😦

jonsterling:
OK, this branch fails locally on a fresh clone. good.

favonia:
The current `test-sml.sh` will still run the tests (against the last successful build) even if the building fails. Not sure how to fix this in a robust way. Is this what causes the problems here? The travis build is always trustworthy because the directory `bin` is newly created.

jonsterling:
@favonia What went wrong is that I am an idiot 😆  I screwed up the `.cm` file of one of the dependencies, now fixed. It's really hard to find the errors in the SML/NJ output!

PR #411, opened 2017-10-25
Sync with the upcoming paper.

favonia:
It doesn't really matter, but the current draft steps into `ax` instead of the cap. @cangiuli


cangiuli:
Good catch.

PR #412, opened 2017-10-25
Kan operations of fcom types, stage 0.

favonia:
#330.

### DIFF
--- a/src/redprl/machine.fun
+++ b/src/redprl/machine.fun
@@ -270,6 +270,7 @@ struct
               let
                 fun hcom x eps =
                   let
+                    (* maybe we can reuse (#1 tube)? *)
                     val y = Sym.named "y"
                   in
                     Syn.intoHcom
@@ -288,6 +289,7 @@ struct
                   end
                 val result =
                   let
+                    (* maybe we can reuse (#1 tube)? *)
                     val z = Sym.named "z"
                   in
                     Syn.intoHcom
@@ -319,6 +321,22 @@ struct
                tubes = List.map (fn ((r1, r2), (u, n)) => ((r1, r2), (u, coe (VarKit.toDim u) n))) (Syn.outTubes system)}
        in
          STEP @@ hcom || (syms, stk)
+       end
+     | O.GCOM $ [_ \ r, _ \ r', [u] \ ty, _ \ cap, _ \ system] || (syms, stk) =>
+       let
+         fun coe s m =
+           Syn.intoCoe
+             {dir = (s, r'),
+              ty = (u, ty),
+              coercee = m}
+          val ghcom =
+            Syn.intoGhcom
+              {dir = (r, r'),
+               ty = substVar (r', u) ty,
+               cap = coe r cap,
+               tubes = List.map (fn ((r1, r2), (u, n)) => ((r1, r2), (u, coe (VarKit.toDim u) n))) (Syn.outTubes system)}
+       in
+         STEP @@ ghcom || (syms, stk)
        end
 
      | O.FCOM $ [_ \ r1, _ \ r2, _ \ cap, _ \ tubes] || (syms, stk) => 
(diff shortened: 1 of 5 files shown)


src/redprl/machine.fun
```diff
@@ -234,7 +234,7 @@ struct
          (case stk of
              [] => raise Final
            | CAP _ :: stk =>
-               STEP @@ cap || (syms, stk)
+               CRITICAL @@ cap || (syms, stk)
```
favonia: @jonsterling I wonder if there's some way to catch this kind of bugs by tying as in the previous machine.

favonia:
@jonsterling @cangiuli Can we actually drop `ghcom` and implement `gcom` directly?

favonia:
Hmm, never mind. If `gcom` will step into `hcom` anyway, all the tubes will still have `coe`.

src/redprl/machine.fun
```diff
@@ -595,58 +653,58 @@ struct
...
                                             (w, nFromOne (#2 dir) (VarKit.toDim w)))
                                          , ((#1 dir, #2 dir),
                                             (w, Syn.into @@ Syn.VPROJ (#2 dir, coercee, Syn.intoFst @@ substVar (#2 dir, v) e)))
-                                         , ((#2 dir, Syn.intoDim0),
+                                         , ((#2 dir, Syn.intoDim 0),
                                             (w, Syn.into @@ Syn.PATH_APP (Syn.intoSnd frontFiber, VarKit.toDim w))) ]}
                                   end
                               in
-                                Syn.into @@ Syn.VPROJ (#2 dir, Syn.intoFst frontFiber, n)
+                                Syn.into @@ Syn.VIN (#2 dir, Syn.intoFst frontFiber, n)
```
favonia: Already found a bug lol.

favonia:
@jonsterling Should I worry about `substVar` not substituting meta-variables? Currently for dimensions meta-variables and variables are treated more or less in the same way, but I am worried that `substVar` will miss the meta-variables.

jonsterling:
@favonia I need to look at the code to see, but there have indeed been places that we needed to worry about this... It's unfortunate, but I don't yet know how to avoid it. What would be nice is if we never needed to do anything fishy with metavariables, but I don't know how to achieve that yet without complicating a lot of stuff.

After my exam today (which is what has been the blocker for me getting shit done on RedPRL, along with the NSF fellowship application), I'll try and do an audit of this code to determine if we need to be more careful.

jonsterling:
@favonia Do you ever substitute for the dimension that is written in an equation? I mean, if you have a tube like `[i=r ...]`, do you ever try to do `substVar (..., i)`?

favonia:
@jonsterling Yes. A lot.

jonsterling:
@favonia OK, in this case we have two choices:

1. Keep things as they are, and don't do fancy things with metavariables; only substitute dimension *variables*. In this case, theorems which take `#i:dim` as an argument won't really do what you naïvely expect, but perhaps this is semantically more sensible. It means it would be harder (usually impossible) to write a lemma about a term which is generic in a dimension `i`—every dimension would have to be introduced through path abstraction, or something analogous.

2. Use some custom `substDimVar` thing that deals with both variables and metavariables.

The latter seems like it would lead to less confusing behavior for users, but is semantically a bit weird. I kind of prefer the first approach (and it means there are some other things in RedPRL that I would want to remove), but I wonder what you think is best.

favonia:
I like the first approach along with some version of #418. Dimensions are no longer parameters. I do not want to recognize `(loog #i)` as values, and #418 will solve the practical issue.

jonsterling:
@favonia Great, I agree fully!

favonia:
Let's merge this to prevent potential merge conflicts?

jonsterling:
@Favonia Sounds good! I'll review this within today's morning.

PR #413, opened 2017-10-25
Kan operations of V types, stage 0.

favonia:
#330.

### DIFF
--- a/src/redprl/machine.fun
+++ b/src/redprl/machine.fun
@@ -533,14 +533,30 @@ struct
      | O.CAP $ [_ \ r1, _ \ r2, _ \ coercee, _ \ tubes] || (syms, stk) => 
        stepCap stability ({dir = (r1, r2), coercee = coercee, tubes = Syn.outTubes tubes} || (syms, stk))
 
-     | O.V $ [_ \ r, _ \ a, _ \ b, _] || (syms, stk) =>
+     | O.V $ [_ \ r, _ \ a, _ \ b, _ \ e] || (syms, stk) =>
          branchOnDim stability syms r
            (STEP @@ a || (syms, stk))
            (STEP @@ b || (syms, stk))
            (fn u =>
              case stk of
                [] => raise Final
-             | HCOM _ :: stk => E.raiseError (E.UNIMPLEMENTED (Fpp.text "hcom operations of V types"))
+             | HCOM (dir, HOLE, cap, tubes) :: stk =>
+                 let
+                   val v = Sym.named "v"
+                   val f = Syn.intoFst e
+                   fun vproj m = Syn.into @@ Syn.VPROJ (r, m, f)
+                   fun m' ty y = Syn.into @@ Syn.HCOM
+                     {dir = (#1 dir, y), ty = ty, cap = cap, tubes = tubes}
+                   val n = Syn.into @@ Syn.HCOM
+                     {dir = dir, ty = b,
+                      cap = vproj cap,
+                      tubes =
+                           ((VarKit.toDim u, Syn.intoDim0), (v, Syn.intoApp (f, m' a (VarKit.toDim v))))
+                        :: ((VarKit.toDim u, Syn.intoDim1), (v, m' b (VarKit.toDim v)))
+                        :: mapTubes_ vproj tubes}
+                 in
+                   CRITICAL @@ Syn.into (Syn.VIN (r, m' a (#2 dir), n)) || (syms, stk)
+                 end
              | COE _ :: stk => E.raiseError (E.UNIMPLEMENTED (Fpp.text "coe operations of V types"))
              | _ => raise Stuck)
      | O.VIN $ [_ \ r, _ \ m, _ \ n] || (syms, stk) =>
--- a/src/redprl/machine.fun
+++ b/src/redprl/machine.fun
@@ -48,8 +48,8 @@ struct
   local
     fun plug m = 
       fn APP (HOLE, n) => Syn.intoApp (m, n)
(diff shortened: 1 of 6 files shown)


favonia:
@jonsterling @cangiuli Wow, after some struggle it seems I finally found a way to further optimize the construction (in a non-trivial way)! To fully take the advantage of the new code, though, I would like to change the definition of contractibilty from
```
(* A (-> [a b : A] (path [_] A a b)))
```
to
```
(* [a : A] (-> [b : A] (path [_] A b a)))
```
To what degree I can break backward compatibility?

jonsterling:
@favonia The first definition that we had was nonstandard right? The standard definition is to give the base point, and then a family of paths to that base point, which looks like what you wanted to change it to. So I vote 👍

jonsterling:
And I'm excited to learn about the optimization you mention 😄

src/redprl/machine.fun
```diff
@@ -557,7 +557,103 @@ struct
...
+                                      {dir = (Syn.intoDim0, VarKit.toDim y),
+                                       ty = (y, substVar (Syn.intoDim0, v) a),
+                                       coercee = substVar (Syn.intoDim0, y) coercee}
+                                  in
+                                    Syn.intoAnonTuple
+                                      [ m
+                                      , Syn.into @@ Syn.PATH_ABS (z,
+                                          Syn.intoCom
+                                            {dir = (Syn.intoDim0, VarKit.toDim y), (* should I worry about name collision here? *)
+                                             ty = (y, substVar (Syn.intoDim0, v) b),
```
favonia: @jonsterling Should I worry about the `y` in `dir` colliding with the bound variable in `ty`?

src/redprl/operator.sml
```diff
@@ -253,6 +253,8 @@ struct
    | DEV_USE_LEMMA of opid * RedPrlArity.t option
    | DEV_APPLY_LEMMA of opid * RedPrlArity.t option * unit dev_pattern
 
+   (* where should this function go? *)
+   fun indexToLabel i = "proj" ^ Int.toString (i + 1)
```
favonia: Where should this function go?

cangiuli:
I don't mind changing the definition of equivalence. But it's a little awkward that it won't match the paper any more -- do you think this is a problem?

favonia:
@cangiuli I think we should update Part IV instead, and we can discuss this tomorrow.

src/redprl/machine.fun
jonsterling: It should be safe as-is.

src/redprl/operator.sml
jonsterling: I'm not sure... Here's probably as good as place as any (for now)

src/redprl/machine.fun
favonia: Do you mean that, after `Syn.into`, any foreseeable usage should see the `y` in `ty` as a refreshed (bound) variable?

jonsterling: Yeah, `Syn.into` will call `Abt.into`, which turns `y` into a de bruijn index; so, in the type, this bound `y` is no longer the same as the `y` that is mentioned in `dir`.

favonia:
@jonsterling @cangiuli Hmm... `(coe 0~>1 [x] (IdV x #l ty) a)` (correctly) reduces to `(coe 0~>1 [_] ty a)`, which may or may not be different from `a`. How much do people want this regularity? I have a feeling that this is actually provable with respect to the cubical type system we are working in.

test/success/V-types.prl
```diff
@@ -69,6 +69,16 @@ Thm IdV/Test1(#l:lvl) : [
...
 ].
 
+Thm IdV/Test2(#l:lvl) : [
+  (->
+   [ty : (U #l kan)]
+   [a : ty]
+   (= ty (coe 0~>1 [x] (IdV x #l ty) a)
+         (coe 0~>1 [_] ty a)))
+] by [
+  lam ty, a. refine eq/intro; reduce; auto
```
favonia: @jonsterling In this case I have to prevent the structural equality rule from running. Should we revisit #288? I am slightly leaning toward to something like `complete (structure + auto) || (reduce + ...)`, or avoiding treating ``coe`` specially, or something fancier...

jonsterling: @favonia I was thinking about it a few weeks ago and I think that it is inevitable that we will need to support backtracking, even if in most cases we can code something up without it... I think it would be worth figuring out how to add backtracking to Dependent LCF; then we could have a `append` tactical like in Isabelle (in Coq I think it is called `+`).

jonsterling: Honestly I think that if we don't implement backtracking, we really would need to consider limiting the `auto` tactic to only apply invertible rules—which limits its power by quite a bit.

jonsterling: It's very easy to understand how to add backtracking when you don't have validations (as in Isabelle), but I think the algorithm is harder when you do have validations. After this week, I can try to cook something up.

favonia:
@jonsterling @cangiuli Alright, what should we do now? We could merge this now and test, or I could spend more time staring at the code. I vote for the former so that I can work on composition types instead. Thought?

jonsterling:
@favonia I saw we merge this now... This kind of code is so difficult to write that I am not sure that staring at it more will help 😄  If there are bugs, we'll find them as we develop more examples.

favonia:
@jonsterling Alright I will do ugly branch-switching now. I'll merge this after your approval.

jonsterling:
@favonia I meant, we should merge now rather than wait for more examples, sorry for the confusion!

favonia:
@jonsterling Oh it is just that I thought you might want to take a final look.

test/success/V-types.prl
jonsterling: This issue is now resolved by my PR #419 which implements backtracking.

PR #420, opened 2017-10-31
Kan operations of fcom types, stage 1.

favonia:


favonia:
@jonsterling Lol. You have not seen nothing yet!

test/success/pi1s1.prl
```diff
@@ -71,3 +71,58 @@ Thm IntSuccEquiv : [
...
+// Needs rewriting on types
+//
+//Thm S1LoopToInt' : [
+//  (->
+//   [x : S1]
+//   [p : (path [_] S1 base x)]
+//   ($ S1UnivCover x))
+//] by [
+//  lam x, p. `(coe 0~>1 [x] ($ S1UnivCover (@ p x)) 0)
+//].
```
favonia: So, what's difficult here is to show that `0` is a member of `($ S1UnivCover (@ p 0))`. The most straightforward way is to replace `(@ p 0)` by `base`, but AFAIK we do not have such tactics yet. Internalization would not help unless we have relaxed equality types which do not demand `m in a` for `(= a m m)` to be a type.

It seems we need to improve tactics to easily handle such basic cases. On the other hand, I seemed to figure out a way to work around this for this part of the proof of pi1(S1).

jonsterling: Aha! Interesting. So, just to make sure I understand, we have `p` a path in the context whose `0`th endpoint is `base`, and we want to rewrite by this?

I see that you say we would need the Nuprl-style equality type to make it work using internalized equality, but I am having trouble understanding what goes wrong if you just use our equality type as-is. But I can try to understand it by stepping through the proof myself...

favonia: To clarify, I believe we can also handle this situation by having special tactics that rewrite the types.

test/success/pi1s1.prl
```diff
@@ -0,0 +1,128 @@
...
+Thm IntSuccPath : [
+  (path [_] (U 0 kan) int int)
+] by [
+  <x> `(V x int int IntSuccEquiv)
+].
+
+Thm S1UnivCover : [
+  (-> S1 (U 0 kan))
+] by [
+  lam x. exact (S1-rec [_] (U 0 kan) x int [x] (@ IntSuccPath x)); refine s1/eq/s1-rec; auto
```
favonia: For this lemma, the reason I did not use the auto tactic is because, currently, the auto tactic would try to prove the membership in a universe by establishing `EQ_TYPE`. However in this case it is more straightforward to use structural equality at the element level directly. @jonsterling Perhaps another good place to use backtracking?

jonsterling: @favonia Good idea!

test/success/pi1s1.prl
favonia: Hmm. My "workaround" is not good enough; I still need to replace `(@ p 0)` by `base` somewhere.

favonia:
The examples showed that it works for the easiest cases. I think we need to improve the infrastructure a little bit before moving on (see the above discussion), and I should go back to the paper draft for the moment. @jonsterling I suppose I can merge this now? Feel free to take over `pi1s1.prl` if you have time.

jonsterling:
sounds good, feel free to merge!

PR #421, opened 2017-10-31
[experimental, do not merge] integrate updated lcf lib

jonsterling:


jonsterling:
(The reason I haven't merged this is that it's a bit slower than the old one. I want to optimize it a bit first.)

jonsterling:
I should be using some kind of state monad here.

favonia:
Is this subsumed by #523?

jonsterling:
@favonia No; this is some experiments I was doing on a new design, but it's nowhere near ready to merge.

cangiuli:
Please don't delete this branch.

PR #433, opened 2017-11-11
Basic results on lower h-levels

mortberg:
I started proving some basic results for lower h-levels (IsContr, IsProp, IsSet), but I ran into a lot of problems. Can someone please take a look at IsPropIsProp and explain what is going on.


favonia:
Let me open up an issue for the highlighting bug.

test/success/hlevels.prl
```diff
@@ -0,0 +1,74 @@
...
+].
+
+Thm PropSet(#l:lvl) : [
+  (->
+   [tyA : (U #l kan)]
+   [h : (IsProp tyA)]
+   (IsSet tyA)
+  )
+] by [
+  // unfold IsSet, IsProp; // why can't I chain unfolds like I can for lam?
```
jonsterling: I think the notation for this is `unfold [x,y,z]`. I can't remember why...

test/success/hlevels.prl
```diff
@@ -0,0 +1,74 @@
...
+  unfold IsSet; unfold IsProp; // why do I need a ; and not .?
```
jonsterling: The `;` is a tactical that sequences tactics, like in Coq. The `.` doesn't mean anything on its own, but is part of the notation for certain tacticals, including `lam x. tac`.

jonsterling:
@mortberg I've have a look at your code and fixed it, but I can't figure out how to make a pull request to your fork of RedPRL. The main lesson is that you should try to use the (undocumented!!) development tactics rather than writing a term manually, since information is lost in the latter style (the woes of not having decidable typing).

In general, it is WAY easier if you just push your branches to the main RedPRL repository rather than using your own fork. That way we can collaborate on pull requests. But if you prefer not to do it this way, that's cool too I guess.

If you want to get the corrected code, pull `hlevels` from the main repository and then push it to your fork.

test/success/hlevels.prl
favonia: The reason was that commas were (are?) used for something else. Think about `[unfold A, B]`.

jonsterling: @favonia ahh yes, that's right!

mortberg: Ok, this is quite confusing for the user... How about only using `,` to repeat a tactic so that I could write `lam x, y, z` or `unfold x, y, z`?  If you implement a rewriting tactic `,` could be used for that as well so that I can write `rewrite H1, H2, H3`.

Alternatively there could be no separator so that I just write: `lam x y z`, `unfold x y z` and `rewrite H1 H2 H3`. Maybe the parser is not flexible enough?

jonsterling: @mortberg  The problem is that we have the tactical `[t1, t2, t3]` already; what do you propose to do instead?

test/success/hlevels.prl
mortberg: This is extremely confusing if you know Coq... Why not use `.` to separate tactics and `;` for sequencing them?

test/success/hlevels.prl
mortberg: Hmm, what does that do?

I kind of like the no separator syntax as it is even faster to type. :)

test/success/hlevels.prl
jonsterling: We don't have any 'top-level' interact like in Coq; what I mean is that in Coq, you're focused on the first subgoal, and `.` just applies a tactic to that subgoal. RedPRL is more in the style of LCF, where you must write a single script to cover your goals, rather than issue commands that change the proof state (as in coq). So, we use the period for a different purpose.

Since we don't have anything analogous to the `.`-structured proofs of Coq, I think there is no opportunity for confusion.

jonsterling: We might choose a different delimiter than a period, of course... In some cases, something like `=>` might work nicely; on the other hand, it wouldn't work for the `let` expressions.

anyway, I'm open to designing a better concrete syntax, since the current one is pretty crappy.

mortberg:
Ok, I'm used to working with local branches on my own fork, but I can work directly on the main repo if you want. I think you could have made a PR to the branch on my fork though and once I merge that it will show up here.

test/success/hlevels.prl
mortberg: Oh, I understand now! I'm not used to writing a single tactic to solve all subgoals... I can't wait for the RedPRL tutorial so that I can learn what is going on without having to ask a million stupid questions :)

jonsterling:
@mortberg OK, cool :) I tried to make a PR to the branch on your fork, but every time I did so, GitHub told me there was no difference between the branches. That didn't make sense to me, maybe it was a UI glitch.

Feel free to keep using your own fork if you're more comfortable with that; I've just found that having forks is a bit redundant and adds to the bureaucracy of the project. But I don't mind it.

mortberg:
Ok, strange! I think forks scale better if you have a lot of developers, but for small projects it is indeed quite bureaucratic.

test/success/hlevels.prl
```diff
@@ -0,0 +1,68 @@
...
+  unfold [IsSet, IsProp];
+  lam tyA, h, a, b, p, q.
+  <j i> `(hcom 0~>1 tyA a
+               [i=0 [k] (@ ($ h a a) k)]
+               [i=1 [k] (@ ($ h a b) k)]
+               [j=0 [k] (@ ($ h a (@ p i)) k)]
+               [j=1 [k] (@ ($ h a (@ q i)) k)])
+].
+
+// This proof should be easy, but it is not...
```
mortberg: So this is still quite complicated... Any ideas how to make it better or why it is so complicated? It is essentially just a direct application of  `PropSet`...

favonia: I just checked this. Ideally the proof should be done by
```
lam tyA, h1, h2. <i> lam a, b. `(@ ($ (PropSet #l) tyA h1 a b ($ h1 a b) ($ h2 a b)) i)
```
but the auto-unfolding in the matching judgments is not implemented yet. :cry: Note the differences between parameters to meta-variables and arguments to functions. @jonsterling Perhaps we can treat all the extra parameters as arguments, but at very least some error messages are needed.

jonsterling: As I described before, it's not likely that writing out the proof term like this is going to work... You really do have to interact with the refiner by using tactics, and that's a blessing rather than a curse: it lets you develop the proof interactively, rather than by having to come up with this proof term all by yourself. If you look at the proof that's written down, it's not really that much more complicated than the proof term you wanted; the only part about it which is more complicated is the fact that you have to unfold things here and there, which is a shame. We may be able to improve that...

Also, regarding auto-unfolding in those match judgments (a technical issue that @favonia will understand), if we want to do this, it should be done in the part of the `auto` tactic that deals with `MATCH` judgments, rather than in the rules for those things. Does that sound reasonable?

favonia: I am wondering if @mortberg refers to `IsPropIsProp`. If so, in that special, simple case, the proof term should work once we have auto-unfolding and maybe other minor improvements.

mortberg: Yes, I am referring to `IsPropIsProp`! It seems to me like RedPRL is going to be extremely difficult to use if just applying a simple lemma turns into 7 lines of code... Ideally the proof should be exactly like what @favonia wrote. So making the automation handle the unfolding better seems like something to prioritize.

jonsterling: I too am referring to `IsPropIsProp`. I think that auto-unfolding in the appropriate places will help, as I said before. But it is not at all the case that the ideal code is what @favonia wrote; we do *not* want users to be typing in terms manually (this is too low-level), they really should be using the interactive tactics. What we want instead is to get unfolding to work well-enough in practice that the tactics are not longer the manual codes.

Here's how it might look if unfolding worked better:

```
  lam tyA, h1, h2.
  <i> lam a, b. use (PropSet #l) [`tyA, `h1, `a, `b, use h1 [`a, `b], use h2 [`a, `b], `i]
```

The above is the ideal code. (Obviously, the concrete notation leaves something to be desired, but what I have written above has the structure of the proof that we want; just quoting a term and hoping that we can guess a typing derivation for it is not what we want.)

jonsterling: One thing I want to stress is that the current `auto-step` tactic is poor enough that there are a lot of low-hanging fruit that can improve the current situation (see issue #438 for @favonia's thoughts on it). But it is not a problem that we can solve in the general case, and it's likely that doing "easy" things in RedPRL will *nearly always* be harder than doing "easy" things in a system with decidable judgmental equality.

This is the trade-off, and we can negotiate it to the point of making many things nicer than they are, but it will always be an issue. We're hoping that the gain in expressiveness which we get will make it worth it, but that is by no means guaranteed. These kinds of difficulties are sadly part of the territory of implementing "extensional type theory" and/or "computational type theory", and we can never really overcome them in a real sense.

mortberg: I see! I guess I'm just too used to writing proof terms directly these days :)

How about merging this PR now and then I'll update it once the ideal code works?

jonsterling:
Sounds great!

Really nice job on your first redprl proofs! I hope we can make it easier and more intuitive in the future :)

mortberg:
Yeah, this is the problem with users... They ask a lot of questions and complain if things don't work the way they expect, but hopefully I'm a good guinea pig to get you ready for the POPL tutorial :)

jonsterling:
Haha true :) Please keep it coming! RedPRL is so mysterious and in many areas just unnecessarily difficult to use, that we really need all the questions and criticisms we can get. 


Maybe simultaneously with creating documentation, we can also try to reform the notations to be less weird and confusing too. Since you aren’t as blinded by history as I am, your input would help us come up with something more usable :) There is nothing which is set in stone.

PR #436, opened 2017-11-11
add more backtracking!

jonsterling:


src/redprl/refiner.fun
```diff
@@ -637,7 +637,7 @@ struct
          | (Syn.PROJ _, Syn.PROJ _) => Lcf.rule o Record.EqProj (* XXX should consult autoSynthesizableNeu *)
          | (Syn.DIM_APP (_, r1), Syn.DIM_APP (_, r2)) =>
            (case (Abt.out r1, Abt.out r2) of 
-               (`_, `_) => Lcf.rule o Path.EqApp orelse_ Lcf.rule o Line.EqApp
```
jonsterling: The `orelse_` is useless, since the first one will simply succeed; if it was a line, we don't know it yet, and we will be stuck. Naturally we would want to use backtracking here, but this blows up the search space sadly, and our examples no longer terminate in reasonable time.

jonsterling:
I'm finding that the architecture of a stateless `auto-step`-based system isn't really going to scale to our needs. I think instead we may need to have some kind of monolithic `auto` tactic which is recursive somehow and maintains some state; we could set it up to use fuel so that the user can control how deep the search goes.

PR #439, opened 2017-11-12
More verbose output from MLton to prevent timeout on travis.

favonia:


favonia:
Hmm this is not working because `test-*.sh` hides the output...

Issue #455, opened 2017-11-18
Tactic cheat sheets


favonia:
This is a spin-off thread from #454. Let's (find someone to) make something like this: http://pr-if.org/doc/play-if-card/play-if-card.pdf

With inspiring slogans, red stars and floating lambdas, of course.

jonsterling:
I love this idea! Great thinking, @favonia.

cangiuli:
Does this cheat sheet include the order of the subgoals generated?

jonsterling:
That'd be very helpful, though ideally this is not something users should need to know about (since using low-level rules should be as rare as possible.) Sadly today, it is helpful to know it.

Issue #404, opened 2017-10-09
Selectors inside judgments?


favonia:
Perhaps we could support `[x at left, x at type, y at term, z at type]` which specifies only parts of the judgments. These will be translated into indexes of the subterms in the tactics, depending on the judgment forms. These tags will be used by `rewrite`, `unfold` and `reduce`.

favonia:
@jonsterling Currently `AJ.map` will leave levels and kinds untouched. Is this intentional?

favonia:
@jonsterling I believe we can remove `AJ.map'`. (My previous question does not make sense.)

Issue #395, opened 2017-10-05
unify all of RedPRL under sexpr notation (?)


jonsterling:
Having a uniform notation for all of RedPRL's sublanguages would be really nice; RedPRL-style sexprs seem a natural choice, because they are easy to parse and easy to print, and (excitingly) it is well-known how to provide excellent editor support for sexprs (e.g. paredit, etc.). Furthermore, it is probably one of the only notations which can be completely accurately highlighted in most editors without compiler-support.

There's a few questions that would need to be addressed:

1. How to differentiated between raw program terms and analogous tactics? For instance, right now we have program terms like `(lam [x] M)` and tactic expression `lam x. t`. One way to deal with the issue is to use different names for the operator, but this may get out of hand.

2. How to avoid tactic scripts getting too deep w/ parentheses; sometimes it can be stressful to edit a very deep point in a large sexpr... Sequential scripts that involve `fresh x -> ...` may become fairly gnarly, for instance. Perhaps this doesn't matter.

I'm sure there are other questions too. I wonder if @favonia has any thoughts about this.

I'm still not completely certain that we should do this; for instance, when we work on the new metalanguage (if that ever gets done), will we adopt a traditional ML-style notation, or will we use sexprs there?

favonia:
For the first one, how about separately having `(intro [x y z] ...)` as the tactic?

jonsterling:
@favonia That's not too  bad, I'm open to it.

freebroccolo:
> How to differentiated between raw program terms and analogous tactics? For instance, right now we have program terms like `(lam [x] M)` and tactic expression `lam x. t`. One way to deal with the issue is to use different names for the operator, but this may get out of hand.

Couldn't you use some sort of special quasiquotation mechanism for this?

jonsterling:
@freebroccolo 

> Couldn't you use some sort of special quasiquotation mechanism for this?

The idea makes sense, but there is an asymmetry between between those tactic-terms and the real terms—the former follow essentially the syntax of sequent calculus, whereas the latter are more like natural deduction. So, for introduction forms they basically coincide, but elimination forms tend to differ.

jonsterling:
Here's some examples of what it might look like 

```lisp
(define (Identity [#l : lvl])
 (-> [ty : (U #l)] ty ty)
 (intros [ty x]
  (use x)))


(define (Monoid [#l : lvl])
 (U (++ #1))
  `(record
    [ob : (U #i)]
[...]

Issue #376, opened 2017-09-15
Allow `fresh` without a follow-up tactic?


favonia:
Currently, the grammar of `fresh` is `fresh blah -> foo; bar`. I feel we should also allow just `fresh blah -> foo`.

jonsterling:
@favonia If we add this, it should be expanded during parsing to `fresh blah -> foo; id`, I guess.

favonia:
That is certainly doable. In the beginning I thought we should ban this syntax, but then I realized the names  are probably visible in the `foo` tactic, and so they could be useful. Is my reasoning correct?

jonsterling:
@favonia Yes, I think so. The semantics of tactic sequencing in RedPRL is that you are pushing names onto the stack for use (only) in `foo`.

Later, btw, I have plans to actually change the semantics of this name stack in a much more sensible way, but I'll wait to discuss that when I have a bit more bandwidth.

PR #456, opened 2017-11-18
Proposal for improvements to concrete notation

### LINKED ISSUE #422: Use `abs x. tactic` instead of `<x> tac`.
I think it is more uniform to have
```
lam x.
let x = ....
abs x.
```
to introduce bindings. Thoughts?

jonsterling:
This doesn't get close to addressing the confusing-ness, but I think that this contains some improvements that will make things more consistent and less confusing.

Resolves #422


favonia:
Random idea: would `claim` match mathematical writings better?

jonsterling:
I kind of like that idea!

favonia:
Do you have ideas about the syntax in #404 in the new paradigm? And sorry for destroying the mergability.

jonsterling:
@favonia I don't have any thoughts yet about that, but I'm open to ideas. I think #404 is a good idea.

jonsterling:
And no worries about the merge conflicts 😄 

@cangiuli @favonia @mortberg Do you think that this PR is generally an improvement? If so, I can merge it and then update any other branches that are currently open. Also happy to hold off on merging this until another time.

favonia:
I think the most difficult part is to remember when to use `in` and when to use `=>`. The rule I figured out is that if it results into a binder in the realizer, then it's `=>`. Otherwise, it's `in`. Is that right? Following this rule, I might suggest
```
fresh x -> blah in blah
```
instead of
```
fresh x in blah; blah
```

jonsterling:
@favonia Yes, you're right that this is still confusing... At first I thought to use `=>` for everything! That would be quite consistent, but I was worried it might be a bit surprising.

I think the rule you figured out may hold, but it is not intentional (at least, I wasn't thinking of it when I came up with these notations). I was more thinking about how in languages (like OCaml) we use an arrow for function binders, but `in` for let bindings, and I tried to extrapolate from there.

favonia:
Let me list the notations for a clearer view, as a precursor of #455...
```
lam x y z =>
abs x y z =>
```
How about semicolons for all of these, then?
```
claim lemma : [ .... ] by [ .... ]; ...
let x = ...; ... (very Haskell-ish)
query x <- loc; ...
fresh x -> tac; ... (???)
```
Also I think
```
a, b | #jdg{%a = %b in bool}
```
should be replaced by
```
[[a b] #jdg{%a = %b in bool} => tac]
```
or something.

jonsterling:
@favonia The reason I don't want to use the semicolon for these is that the semicolon already has a meaning on its own: it is a tactical. Since these tacticals cannot really be construed as instances of the semicolon, it would lead to confusion I think.

jonsterling:
I like your idea about improving the match syntax!

jonsterling:
But maybe the semicolon is OK, perhaps users don't need to understand the difference. What do you think?

favonia:
Are you talking about `;` being the sequencing operation?

jonsterling:
Yeah

favonia:
I don't know. These notational issues never bother me too much. I haven't thought about the semantics carefully... perhaps you could elaborate more on the important differences in your mind? In the name of god of S-expressions, I think they can be "disguised" as sequencing if we really want to. Suppose we finally decide to use `(seq [x y z] tac1 [w] tac2)` as the ultimate proof construct:
```
[lemma] (claim [ .... ] [ .... ])
[x] (...?...) // let
[x] (query loc)
[x] (elim y)
```
Note that `(seq [x] tac1 [y] tac2)` takes the freedom to break the usual binding structure. It should be parsed into some ABT that makes sense. See #395.

jonsterling:
@favonia One important difference is that in `t1; t2`, `t2` is applied to all the subgoals generated by `t1`. But in `lam x; t`, `t` would be applied only to one of two subgoals generated by `lam x`. So, it disrupts the semantics of the semicolon and makes it a little harder to understand what's going on.

Maybe it doesn't really matter... I have a hard time figuring out which kinds of deviations are most likely to cause confusion among users.

jonsterling:
With that said, I'm leaning toward finding a way to escape from the `seq` (which is currently written using that `fresh` thing) being a binding operator... It is kind of nice for modularity of proofs, but it has also caused a lot of complication. I don't yet have a proposal for resolving that at the moment, so feel free to ignore this remark.

favonia:
I think the difference you mentioned is clear and I did not suggest semicolons for `lam` or `abs`. How about other tacticals?

jonsterling:
> I think the difference you mentioned is clear and I did not suggest semicolons for lam or abs. How about other tacticals?

Oh, then I misread your comment! Sorry. I think it makes sense the ones you actually suggested. 😄  Your proposal has the benefit of making it so that `in` is only ever used for one thing, which is to select targets like in `reduce` or `unfold`!

favonia:
> `[[a b] #jdg{%a = %b in bool} => tac]`

Actually now I'm not sure about `[a b]` which is very S-expression-with-RedPRL-flavor but we are not faithful S-expressionists yet.

jonsterling:
> Actually now I'm not sure about [a b] which is very S-expression-with-RedPRL-flavor. We are not faithful S-expressionists yet.


Maybe a compromise for now would be to at least get rid of the commas, since I have got rid of commas from the other things.

test/success/equality-elim.prl
```diff
@@ -1,16 +1,16 @@
 Tac GetHole(#c : [exp].exp, #t : [exp].tac) = [
-  query gl <- concl.
+  query gl <- concl;
   match gl {
     [hole | #jdg{(#c %hole)} => (#t %hole)]
   }
 ].
 
-// We can write a cool user-defined tactic for asserting and then rewriting along an equality.
+// We can write a cool user-defined tactic for claiming and then rewriting along an equality.
```
jonsterling: lol autocorrect 😎

jonsterling:
@favonia I think I've integrated your Thought now.

favonia:
Since we are changing the grammar, perhaps we can kill #376 as well.

jonsterling:
@favonia good idea! done.

test/success/hlevels.prl
```diff
@@ -38,19 +38,15 @@ Thm PropSet(#l:lvl) : [
...
-  (abs i => lam a b =>
-    let foo = (PropSet #l) [`tyA, `h1];
-    unfold IsSet IsProp;
-    use foo [`a, `b, use h1 [`a, `b], use h2 [`a, `b], `i]);
-  unfold PropSet; auto
+  abs i => lam a b =>
+    `(@ ($ (PropSet #l) tyA h1 a b ($ h1 a b) ($ h2 a b)) i)
 ].
 
 Thm IsPropIsProp'(#l:lvl) : [
```
favonia: Wait, did you see `IsPropIsProp'` below?

favonia: Perhaps we can delete `IsPropIsProp'` now.

jonsterling: oops, I mindlessly didn't see that

jonsterling: Well, I'll leave both in because the first one indicates that there are some kinks to be worked out in the tactics.

jonsterling:
I want to fix some parser issues (like #469 and #468) from *this* branch rather than from `master`, to avoid duplicating work.

@cangiuli To avoid causing trouble for you, there are a couple alternatives:

1. We could merge your partial `UA` work into master, and then I can rebase this onto that and update your code.

2. We could merge this first, but then I can quickly bring your branch up to date.

3. If you don't like this branch, we could abandon it or try again later.

cangiuli:
@jonsterling Feel free to merge my work into master. If you comment out `UARetSig` then the file should work. You can also squash my commits if you want; they're not useful and my rebase also caused some duplication.

jonsterling:
@cangiuli OK! Thanks. I'll take care of it!

Issue #464, opened 2017-11-19
Removal of kind tags to atomic judgments?


favonia:
Maybe level tags as well if they do not help synthesize the levels. I do not feel they are as useful as I thought.

jonsterling:
I would support removing these, if you think it can be done.

favonia:
Closed by #501.

PR #466, opened 2017-11-20
Destroy EQ_TYPE step by step.

### LINKED ISSUE #407: Remove EQ_TYPE judgment now?
@favonia Now that we have a top level as a façon de parler (not ontologically real), is it possible to remove the type judgments?

favonia:
Resolve #407 and #464.


src/redprl/refiner_types.fun
```diff
@@ -246,42 +244,42 @@ struct
...
         val Syn.WIF ((y, c1y), m1, (t1, f1)) = Syn.out if1
 
         (* motive *)
         val z = alpha 0
         val c0z = VarKit.rename (z, x) c0x
         val c1z = VarKit.rename (z, y) c1y
-        val Hz = H @> (z, AJ.TRUE (Syn.into Syn.WBOOL, inherentLevel, inherentKind))
-        val goalTy = makeEqType (Hz) ((c0z, c1z), NONE, K.top)
+        val Hz = H @> (z, AJ.TRUE (Syn.into Syn.WBOOL, inherentLevel))
+        val goalTy = makeEqType Hz ((c0z, c1z), l, k)
```
favonia: Reverted to safer checking. @cangiuli The question is, for `wool-rec` to be well-typed, maybe the motive `z.C` does not need to be Kan everywhere, but only at `C[m/x]` where `m` is the eliminated term. Did you prove or disprove this?

favonia:
Needs `SUB_TYPE` to handle universe mismatch.

src/redprl/refiner_misc.fun
```diff
@@ -192,13 +193,13 @@ struct
             Metavar.Ctx.empty
             (metas, args)
 
-
         val ty' = substMetaenv mrho ty
-        val _ = if Hyps.isEmpty H' then () else raise Fail "Synth.Custom only works with empty sequent"
+        val _ = if Hyps.isEmpty H' then () else
+          E.raiseError @@ E.IMPOSSIBLE (Fpp.text "Open judgments attached to custom operator.")
```
favonia: @cangiuli @jonsterling Any suggestion on the error message about somehow a sequent with non-empty `H` being associated with a custom operator? (It should be impossible, because the parser should prevent the user from doing so.)

@jonsterling Actually, is there a reason not to forbid `H` directly in typing?

jonsterling: @favonia I need to think about this a little bit. Don't worry about it for now, but I can hopefully make this a lot cleaner.

favonia:
Some bizarre things:
1. When a theorem fails to type check, the usage of that theorem results into arity mismatch.
2. When a rule fails because the binding is not exhaustive, the error message is terrifying.
3. Weird "Pos" raised from the Error module... shouldn't they be pretty-printed somewhere?

favonia:
@jonsterling I am not able to fix all the files and probably need your help.

jonsterling:
@favonia (1,2) are known, let me open a ticket for (1). 

And let me know what you need me to do, and I can help. Do you just want me to try to get the rest of the files to typecheck?

favonia:
@jonsterling How about (3)? I thought all the errors will be pretty-printed somehow. A `Pos` exception seems to indicate that there are some leaked exceptions. Is it related to (2).

Yes, I need help in making them typecheck. I modified the elaborator and something might be wrong there.

jonsterling:
@favonia I'm not yet sure about (3), but I think you are right about it.

I'm digging into the tests now and I'll try to figure out what's going wrong. At first glance, I already see some problems, which I'll see if I can fix.

jonsterling:
@favonia I'm feeling a little discouraged about this. Specifying these "maximum" (?) levels in the theorem statements seems like it is very brittle, and makes a lot of things a lot more difficult to prove. In particular, I'm getting `SYNTH` goals that I don't think I can prove.

favonia:
@jonsterling Would it be better if we implemented some level synthesizer for TRUE judgments?

favonia:
@jonsterling Are you talking about the impossible `SYNTH` in `univalence.prl` about the `IdIsEquiv`? I believe there's some bug, because the levels should match.

jonsterling:
Before continuing fixing the tests, I am going to spend a few hours trying to think if there is a better way to achieve the goals of this PR. After using it for a bit, I feel that I can say the following things:

1. It's not clear to a non-expert user what the meaning of these `at l` annotations is, and you just kind of try and guess what to put in them until things work. It really feels like an artifact of the design of the proof theory rather than something that the user should see.

2. I am worried that the brittleness of this approach will lead to a need for duplicated theorems, etc. I'm not sure of this, but it seems likely.

It feels a lot more brittle than Nuprl or pre-level-polymorphism Coq ever did, but maybe part of this is due to bugs rather than the design. It's a little hard for me to tell at the moment.

favonia:
In the meanwhile let me take over the branch...

jonsterling:
I looked into things a bit, and I don't think that issue with `IdEquiv` is caused a bug; I think that it is 'correct' behavior, but that the design of the proof theory has problems. Consider especially the `EqType` rule for `V` types; here, we are trying to show `(V ...) in (U l1) at l2`, and this decomposes to showing that something is an equivalence at level `l1`, but this leads us to some impossible goals. In particular, we have to synthesize something at `l1` which is only synthesizable at `l2.

favonia:
@jonsterling RedPRL seems to be happy when I supply the type ``(-> [ty : (U #l hcom)] (Equiv ty ty))` directly. I still believe it's related to the substitution you just fixed, and the bug is right in `Custom.Synth`...

favonia:
Another thing is that the auto-step tactic fails to solve some easy instances of `SUB_TYPE`.

favonia:
Level substitution bug fixed. Next one...

favonia:
I fixed the two bugs and seemed to introduce another one about levels...

@jonsterling I think `V.EqType` is fine. The problem is that the universe level of `IdIsEquiv` is unnecessary high due to `(U #l hcom)`. Then any usage of `IdIsEquiv` is locked at level `(++ #l)` unless it is unfold.

jonsterling:
@favonia Yes, that's kind of my point... I think we can't actually choose a sensible universe level beforehand, and the fact that we have to unfold it to make the thing provable suggests that our design isn't really workable.  Maybe I'm misunderstanding, though...

favonia:
@jonsterling Alright. I just wanted to clarify the issue here is that, for elimination, the universe level could suddenly drop. Here the function type and the domain type all have higher universe levels than the codomain type, and hence the problem...

favonia:
I believe we have to abandon this PR for the moment, but there are lots of useful things to pick up from this PR... in particular, are you able to debug the `num.prl`? Not sure how to expose more of its internal for me to trace down.

jonsterling:
@favonia I agree, this PR includes a number of things that we should salvage.

To clarify, do you want me to try and figure out what is currently going wrong with `num.prl` on this branch? I'd be happy to.

favonia:
@jonsterling Yes. That needs to be fixed.

src/redprl/refiner_types.fun
```diff
@@ -1840,149 +1750,96 @@ struct
...
-        val (motiveGoal, motiveHole) = makeTerm (truncatedHx) O.EXP
-        val motiveWfGoal = makeType (truncatedHx) (motiveHole, l, k)
+        val truncatedHx = truncatedH @> (x, AJ.TRUE (holeTy, l))
+        val (motiveGoal, motiveHole) = makeTerm truncatedHx O.EXP
+        val motiveWfGoal = makeType truncatedHx (motiveHole, l, K.top)
 
         val motiven = substVar (holeN, x) motiveHole
         val motivem = substVar (holeM, x) motiveHole
 
         val (H', catjdg') = Selector.map sel (fn _ => motiven) (H, catjdg)
```
jonsterling: Here is the problem in `num.prl`; this replaces every slot of an atomic judgment with `motiven`. I think the logic we want is to replace a `TRUE (a, l)` judgment with the judgment `TRUE (motiven, l)`.

jonsterling:
@favonia OK, I've fixed the problems that were affecting `num.prl`.

src/redprl/refiner.fun
```diff
@@ -245,7 +245,12 @@ struct
         val motiven = substVar (n, x) motiveHole
         val motivem = substVar (m, x) motiveHole
 
-        val (H', catjdg') = Selector.map sel (fn _ => motiven) (H, catjdg)
+        fun replace jdg = 
+          case jdg of 
+             AJ.TRUE (_, l) => AJ.TRUE (motiven, l)
+           | _ => jdg
+
+        val (H', catjdg') = Selector.map sel replace (H, catjdg)
```
favonia: Are you saying that the map now accidentally replaces the `l` due to the new fix about level substitution?

jonsterling: @favonia Yes; but it was probably not right before that even, because `Selector.map` was calling `AJ.map`; when the judgment in question is an equality `m = n in ty`, it would replace it with `motiven = motiven in motiven`.

favonia:
Thanks. Now everything passes except `V-types.prl`.

src/redprl/refiner.fun
favonia: Well, it can and should only rewrite a `TRUE` judgment.

favonia: The case analysis above on `Selector.lookup` rejects other forms of judgments.

jonsterling:
Ah! So it would have worked aside from the level issue.  Thanks for clarifying!

I feel that the change I made makes things a little more modular anyway though.

src/redprl/refiner.fun
```diff
@@ -519,126 +479,64 @@ struct
...
-          orelse_
-        StepEqTypeNeuByUnfold sign tys blockers
-
-      fun StepEqTypeNeuExpand sign ty =
-        fn Machine.VAR z => AutoElim sign z
-         | Machine.OPERATOR theta => Lcf.rule o Custom.Unfold sign [theta] [O.IN_CONCL]
-         | _ => fail @@ E.NOT_APPLICABLE (Fpp.text "StepEqTypeNeuExpand", TermPrinter.ppTerm ty)
-
-      fun StepEqType sign (ty1, ty2) =
-        (Lcf.rule o Computation.SequentReduce sign [O.IN_CONCL]) then_
```
favonia: I am a little confused by this line. Why do we need it if we are going to reduce it anyways?

jonsterling: I'm not sure :) I'm not sure which of us wrote this code, and I can't remember the details. I wonder if it affects anything?

favonia:
I want to clarify that expanding the definition is not needed, but synthesizing a type at some unknown level is the issue. The current mechanism leads to, for example, that domains and codomains live at the same level. Which can still handle the current case (as shown).

jonsterling:
One idea for resolving the contradiction of levels could be the following:

1. Don't annotate judgments with levels at all.
2. In a case where we need to prove that something is a type (but don't know at what level to try), just generate a new subgoal `TERM(lvl)`.
3. Every time we have something like `assertLevelLeq`, replace this with a subgoal of the judgment `l1 <= l2` or something.
4. Finally, once we have refined all our ordinary subgoals, we'll have a proof state which just has a bunch of subgoals like `TERM(lvl)` and `#0 <= #1`, etc. Then, we can write a solver that tries to find an assignment that satisfies the constraints; then, we write some multitactic `ResolveLevels` would call this solver and instantiate all the subgoals appropriately.

This multitactic `ResolveLevels` would get called at the end of a proof (perhaps automatically); if it succeeds in finding an instantiation, then the proof is done, otherwise the proof won't work. In almost every case, I expect that an instantiation could be found.

@favonia What do you think of this approach? I think part of the problem with the current proof theory is we are trying to fix level assignments too early & synchronously, whereas this may need to be asynchronous in the style described above.

favonia:
@jonsterling That might work. This is really the application of dependent LCF to Nuprl's approach on levels.

jonsterling:
@favonia OK, I think I will try it then! It seems reasonably lightweight.

favonia:
@jonsterling Cool. For the moment there are still two big things to be extracted from this PR: #464 and #481.

jonsterling:
@favonia OK, sounds good. After that is done, I'll get down to work. (or let me know if you need help in the meanwhile)

favonia:
Hmm #407 and a complete solution to #464 have to be done almost at the same time.

favonia:
I keep committing stuff to this PR just to see what's left better. The plan is that all the changes will eventually be covered by other PR, if they are worth merging. Let me close this to reduce confusion.

favonia:
Hmm Github refused to recalculate the diff?

favonia:
Maybe I have to open this to see the updated diff...

favonia:
Alright, I believe nothing major & interesting is left here.

PR #471, opened 2017-11-21
Proved equivalence induction!

cangiuli:


cangiuli:
@favonia @jonsterling RedPRL is slow on this file (~3.3s under SML/NJ). The slowest individual proofs are `LemPropF` and `PropSet`, each accounting for 20% of the time spent.

Here are some performance issues I've run into:

1. #472 forces me to turn application into ``` ` ``` once I finish a proof.
2. Unfolding definitions (that would get unfolded by `auto` anyway) can speed things up.
3. Having `claim`s whose definition can be unfolded would reduce some code duplication and unnecessary abstraction. I would certainly inline a couple proofs toward the end if this were possible.
4. Frequently one constructs pairs of a composite and its filler. If I understand correctly, all the adjacency subgoals are being generated and resolved twice in this case. I would like to simply check the filler and get the composite for free, but I don't see a way to do this right now. (If I say the filler is in the path type from the cap to the composite, path type formation will have to check the composite separately anyway. If I `claim` the filler in a line type, the definition is opaque which isn't good enough.)

PR #487, opened 2017-11-23
Apply substitution to levels and fix rewrite rules.

favonia:


test/success/num.prl
```diff
@@ -70,7 +70,7 @@ Thm Plus/zero/R : [
   lam n =>
     fresh n' ind -> elim n;
     [ `ax
-    , fresh x -> rewrite-hyp ind;
+    , fresh x -> rewrite ind;
```
jonsterling: I can't remember, what's the difference between `rewrite-hyp` and `rewrite`?

favonia: `rewrite-hyp` was the original one when `p` and `,p` were still different. Later on `rewrite` was added, if I remember it correctly. `rewrite-hyp p` can handle equality judgments but `rewrite` cannot. Without the comma, there's no need for `rewrite-hyp` to handle TRUE judgments because `rewrite-hyp p` can just be written as `rewrite p`. After internalizing the equality judgments, `rewrite-hyp` can be dropped completely.

jonsterling: @favonia OK thanks! That makes sense.

PR #491, opened 2017-11-24
More patches from #466.

favonia:
I am not sure about the refactoring of `StepEq` done in #466... I am worried that it contributes to the slowdown as well.


favonia:
@jonsterling Needs this to remove kind tags.

PR #494, opened 2017-11-25
Apply Jon Sterling Thought (?)

jonsterling:
This is (the beginning of) an attempt to remove kind and level annotations from judgments, and instead generate level constraints to be solved asynchronously.

This will be very broken for a while.


favonia:
Maybe it is more efficient for me to point out the rules of thumb:
1. the base type in `HCom` should be an `HCOM` type, and
2. the family in `Coe` should be a `COE` type, and
3. `EQ_TYPE` should check the kinds, as directed by those functions enumerating all possible kinds, and
4. elimination rules need subtyping for universes, and
5. tubes in the fcom types need to be at least `COE`.

jonsterling:
I will try again later.

favonia:
Anything left in this branch that the Overmind was not aware of?

jonsterling:
Let's just delete it 😄

PR #498, opened 2017-11-25
Update README to point to example/.

cangiuli:
The README currently points to a nonexistent file. I'm changing it to point to the univalence development.


favonia:
I feel it is better to point to example/README.prl, and only mention the univalence for citation?

cangiuli:
That's fine too, but I was a little confused about what README.prl is supposed to be.

favonia:
`example/README.prl` is intended to be the first file for users to look at. Maybe `example/basics.prl`? @jonsterling

PR #501, opened 2017-11-26
Remove most kind annotations.

favonia:
The level checking is suboptimal because this patch was extracted from PR #466, but most level checking will be removed anyways.

### DIFF
--- a/example/README.prl
+++ b/example/README.prl
@@ -192,7 +192,7 @@ Print FunExt.
 Tac FunExtTac(#l : lvl) = [
   query gl <- concl;
   match gl {
-    [a b f g k | #jdg{(path [_] (-> %a %b) %f %g) true with %[k:knd]} =>
+    [a b f g | #jdg{(path [_] (-> %a %b) %f %g) true} =>
       use (FunExt #l) [`%a, `%b, `%f, `%g, id]
     ]
   }
--- a/example/univalence.prl
+++ b/example/univalence.prl
@@ -15,7 +15,10 @@ Def Retract (#A,#f,#g) = [(-> [a : #A] (path [_] #A ($ #g ($ #f a)) a))].
 Tac ReflectUniv(#l:lvl, #k:knd, #t:tac) = [
   query goal <- concl;
   match goal {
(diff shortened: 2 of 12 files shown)


jonsterling:
@favonia This looks good! But it's going to make very bad merge conflicts for #494. I started working on #494 when I did because you said in a comment that kind and level annotations had to be removed *simultaneously*.

So, I'm fine with merging this, but it might mean that I don't have time to fix #494; so, if you want to take that over, then we can merge this now. Or we can wait until the full change is finished.

favonia:
@jonsterling I don't remember what exactly I said where, but I suppose I was talking about `EQ_TYPE` and kinds have to be removed at the same time, which you probably already know.

I did not remove levels yet because I wanted to first double-check that nothing interesting is left in #466.

favonia:
@jonsterling My plan is a less aggressive version of #494, which still has `SUB_KIND` or `SUB_TYPE` for level/kind mismatch. Most levels will be removed as well if my proposal in #407 is implemented. Does that sound okay to you?

jonsterling:
@favonia I am doubtful that this will work long-term... But if you feel it is the best course of action, I'm open to it.

Could you argue for it a bit, in comparison to the more aggressive approach pursued in #494? It would help to understand your long-term vision of these questions in relation to the proposals that we have discussed.

favonia:
@jonsterling My motivation is to finish the parts we agree (removal of various annotations) so that the miserable ending of #466 would not happen again. I do not have a strong argument against #494. My proposal suffers from code duplication, but on the other hand, #494 suffers from a potential issue that, for example, every time you want to establish membership in pushouts, you probably need 2 more universe levels. and the membership in fcom types could be much worse. I do not know which one would be better in the end.

favonia:
@jonsterling I was hoping that, once the level annotations are removed as well, you can focus on destroying EQ_TYPE with your Thought without worrying those nasty kind checking.

favonia:
About the miserable ending of #466: I want to make sure that useful things get committed even when the big proposed change fails.

favonia:
@jonsterling Thinking about it, perhaps I misunderstood your approach and only one metavariable is generated for each of those `EQ` judgments?

PR #502, opened 2017-11-26
Remove level tags from atomic judgments.

favonia:
@jonsterling You might want to check whether this is still on the common path shared by us.


jonsterling:
this sounds just like what I have in mind. good!

src/redprl/refiner_kit.fun
```diff
@@ -207,6 +207,17 @@ struct
     if Abt.eq (a, b) then NONE
     else SOME @@ makeSubType H (a, b)
 
+  (* functions which blur the difference between EQ and EQ_TYPE *)
+  val viewAsEqType =
```
favonia: Manual "view as" pattern matching.

src/redprl/refiner_types.fun
```diff
@@ -56,10 +56,10 @@ struct
     fun EqType _ jdg =
       let
         val _ = RedPrlLog.trace "Bool.EqType"
-        val H >> AJ.EQ_TYPE ((a, b), l, k) = jdg
+        val H >> ajdg = jdg
+        val ((a, b), _, k) = viewAsEqType ajdg
```
favonia: @jonsterling So this is the new programming style to handle both `EQ` and `EQ_TYPE` at once.

jonsterling: cool! That sounds reasonable.

favonia:
Hmm let me merge the coequalizers...

favonia:
@jonsterling I introduced non-deterministic auto-synth in `auto`. This way we can `claim` some well-typedness.

jonsterling:
> @jonsterling I introduced non-deterministic auto-synth in auto. This way we can claim some well-typedness.

Interesting! Is there an example of where this fires? (If not, you don't need to come up with one, I was just curious.)

favonia:
> Is there an example of where this fires?

Somehow I suddenly need this in `semi-simplicial.prl`. See the addition of `aux2` in two places. If we can avoid it that would be great!

Issue #495, opened 2017-11-25
beta rules?


ecavallo:
Is it currently possible to construct a path from the identity function `lam [a] a` on S1 to its eta-expanded equivalent `lam [a] (S1-rec [_] S1 a base [x] (loop x))`? This would require a beta rule for `S1-rec` applied to `loop x`, which seems not to exist; the loop case isn't a stable reduction.

Here is a proposal to solve this in a swoop. Theorem: If M in A and we have an unstable reduction M |->* M', we can still conclude that M = M' in A if we already know M = M' in A _at those restrictions where the reduction no longer holds_.

For example, suppose we want to show `(= S1 (S1-rec [_] S1 (loop r) base [x] (loop x)) (loop r))`. Since the left-hand side is in `S1` and we have `S1-rec [_] S1 (loop r) base [x] (loop x)` |-> `loop r` when r /= 0,1, the theorem above reduces this to proving the equality at the restrictions r = 0 and r = 1, i.e.,  proving `(= S1 (S1-rec [_] S1 (loop 0) base [x] (loop x)) (loop 0))` and  `(= S1 (S1-rec [_] S1 (loop 1) base [x] (loop x)) (loop 1))`. These two hold by stable computation, so QED.

So, if the machine were modified to do unstable reduction and report the apartness conditions under which that reduction holds, you could have a tactic which generates subgoals where it doesn't. (Of course, here you have to say what type you're reducing in and start with a term in that type.) The unstable beta rules (`S1-rec` on `loop`, `wif` on `fcom`, etc) can all be derived using that. I'd work on this (at some point) if it sounds like a good idea.

favonia:
Related: #417.

ecavallo:
Since we have the intro rules, it would be simpler for beta rules to use Lemma 54 from CHTT II, which implies:

If M in A, M' in A, and M |-> M', then M = M' in A.

But this doesn't make sense when we're reducing open terms, if I understand correctly?

ecavallo:
To clarify: if I have it right, the meaning of M |-> M' when these are open terms is that the reduction holds for all closing term substitutions, whereas the meaning of M = M' in A is that the equation holds for all _dimension substitutions_ followed by a closing term substitution. So there's a mismatch when you try to openfy Lemma 54. This isn't an issue for my first suggestion, since the equality subgoals take care of the dim substitutions where the reduction no longer holds.

jonsterling:
[...]

PR #503, opened 2017-12-01
Unstable beta rules (#495)

ecavallo:
Added loop-beta and glue-beta rules per #495. (And poorly-written test scripts.)


ecavallo:
Going to make these simpler.

PR #504, opened 2017-12-01
Unstable beta rules

ecavallo:
Beta rules for the unstable reductions of `s1-rec` (`loop`) and `pushout-rec` (`glue`). The rules as they are specified here are justified by the lemma I mentioned in #495.


favonia:
I wonder if we can just have one single tactic to do unstable computation for well-typed term? The machine already supports unstable computation.

cangiuli:
Related question: do we currently have a tactic that says if a term is closed, it is equal to its unstable value? This will be handy for complicated calculations involving univalence.

ecavallo:
@favonia Yes, definitely. But the machine needs to be modified in the following way:
> given an open term M, the machine should return a term N and a list E of equations such that M\psi |->* N\psi for any \psi which does not make any of the equations in E true.
To prove M = P in C, if the machine produces (N,E) on input M, then the generated subgoals are M = P in C [\xi] for each \xi in E. (In particular, if the reduction is stable, then there are no subgoals.)

(Types aren't closed under arbitrary unstable computation, if that is what you were suggesting.)

jonsterling:
Nice work.

jonsterling:
> Related question: do we currently have a tactic that says if a term is closed, it is equal to its unstable value? This will be handy for complicated calculations involving univalence.

We don't yet have this, but it sounds sensible.

And @ecavallo's idea of changing the machine to generate this list sounds interesting too; I'd like to understand it a bit better on paper, but it sounds promising.

ecavallo:
[Here](http://www.cs.cmu.edu/~ecavallo/garbage/typed-reduction.pdf) is some paper.

jonsterling:
@ecavallo I am sorry that I lost track of this! Thanks for adding it...

I want to coordinate it's merge with @favonia. And re: your idea for enhancing the machine, I think that also sounds cool! Sometime we should figure out how to carry it out.

jonsterling:
I'm going to close this and momentarily reopen another one

PR #506, opened 2017-12-01
Fix a few tiny "bugs"

ecavallo:
- `seq` was biting off unnecessarily many names
- Pushout.EqElim used `alpha 2` twice
- `insertAstDecl` and `elabDecl` were mysteriously truncating the input signature

### DIFF
--- a/src/redprl/name_seq.sml
+++ b/src/redprl/name_seq.sml
@@ -13,7 +13,7 @@ struct
         if i < n then
           List.nth (us, i)
         else
-          alpha (i + n)
+          alpha (i - n)
     end
 
   fun bite n alpha =
--- a/src/redprl/refiner_types.fun
+++ b/src/redprl/refiner_types.fun
@@ -1685,9 +1685,9 @@ struct
 
         (* glue branch *)
         val (goalTyC, holeTyC) = makeMatch (O.PUSHOUT, 2, holeTyPushout, [])
-        val v = alpha 2
+        val v = alpha 3
         val vtm = VarKit.toDim v
-        val c = alpha 3
+        val c = alpha 4
         val ctm = VarKit.toExp c
         val q0vc = VarKit.renameMany [(v, v0), (c, c0)] q0v0c0
         val q1vc = VarKit.renameMany [(v, v1), (c, c1)] q1v1c1
--- a/src/redprl/signature.sml
+++ b/src/redprl/signature.sml
@@ -419,16 +419,11 @@ struct
       end
 
     fun elabDecl (sign : sign) opid (decl : src_decl, pos) : elab_sign =
-      let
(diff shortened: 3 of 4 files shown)


src/redprl/name_seq.sml
```diff
@@ -13,7 +13,7 @@ struct
         if i < n then
           List.nth (us, i)
         else
-          alpha (i + n)
+          alpha (i - n)
```
jonsterling: Wow. I'm really surprised this didn't cause more things to go wrong.

src/redprl/tactical.fun
```diff
@@ -60,9 +60,8 @@ struct
 
     fun seq (mt1 : multitactic, (us : Sym.t list, mt2 : multitactic)) : multitactic = fn alpha => fn st =>
       let
-        val beta = Spr.prepend us alpha
-        val (beta', modulus) = Spr.probe (Spr.prepend us beta)
-        val st' = mt1 beta' st
+        val (beta, modulus) = Spr.probe (Spr.prepend us alpha)
```
jonsterling: Likewise, I'm just flabbergasted that we didn't notice this before! Really good catch.

src/redprl/signature.sml
```diff
@@ -474,11 +469,7 @@ struct
 
 
     fun insertAstDecl sign opid (decl, pos) =
```
jonsterling: I think that the reason for doing this was in case the declaration already existed; if so, replacing it would invalidate the subsequent declarations. This would never come up in the current interaction model, though...

Is this change in response to observed behavior, or did the code just look weird?

ecavallo: Just weird-looking. Since there is a `handle` for duplicate entries, I assumed that raising an error was the desired behavior. The reason that said error does actually get raised with the current code is because of (what I perceive to be) a bug in sml-telescopes, where `truncateFrom sign opid` doesn't actually remove the `opid` entry from the telescope's dictionary, despite removing the label from its list of labels.

jonsterling:
Oh, you are right about duplicates... Let’s keep your change here, and also fix the bug in the library.

src/redprl/refiner_types.fun
```diff
@@ -1685,9 +1685,9 @@ struct
 
         (* glue branch *)
         val (goalTyC, holeTyC) = makeMatch (O.PUSHOUT, 2, holeTyPushout, [])
-        val v = alpha 2
+        val v = alpha 3
         val vtm = VarKit.toDim v
-        val c = alpha 3
+        val c = alpha 4
```
favonia: Very nice. Thanks!

src/redprl/tactical.fun
favonia: Agreed! It must be my daily prayer to the god of syntax. Cannot think of any other reason.

PR #511, opened 2017-12-05
Add vim mode.

cangiuli:
Somebody should double-check that this works on a computer that isn't mine before merging.


jonsterling:
I'll upgrade my vim and try it out! This is so cool!!!

jonsterling:
I've tested this on my laptop and it works perfectly!

favonia:
Yay! ~Time to uninstall Emacs!~ Oh wait, I still need Proof General.

PR #514, opened 2017-12-19
Add command-line option for linewidth. (Closes #449.)

### LINKED ISSUE #449: Flag to adjust linewidth of redprl printout
This is important for when the term is too big, and you end up with like 1000 lines of silly five-character spurts.

@cangiuli

cangiuli:
(Work in progress.)

### DIFF
--- a/src/frontend.cm
+++ b/src/frontend.cm
@@ -3,6 +3,5 @@ Group is
   ../lib/sml-final-pretty-printer/final-pretty-printer.cm
   redprl.cm (bind:(anchor:libs value:../lib/))
   debug.cm
-  redprl/config.sml
   frontend/frontend.sml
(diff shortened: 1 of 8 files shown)


cangiuli:
(Work in progress.)

cangiuli:
My last commit works in SML/NJ but not in MLton. The frontend code correctly sets `Config.maxWidth` but `pretty.sml` sees the default setting of `Config.maxWidth`. Does someone here understand how `.mlb` files work? @favonia

cangiuli:
OK, so that commit didn't work because I forgot to add `config.sml`, but my problem still remains. :)

cangiuli:
Never mind, it works now, thanks to @jonsterling.

Issue #457, opened 2017-11-18
Remove member equality judgments


jonsterling:
Now that we have equality pretypes, it is technically possible for us to remove member equality judgments and work in the Nuprl style.

Related to #351.

PR #515, opened 2017-12-20
Tactics to rewrite types

### LINKED ISSUE #404: Selectors inside judgments?

### LINKED ISSUE #465: Rules for rewriting types
Rules for rewriting types enjoy a special property that functionality always holds. The usual rewrite rules are next to impossible to rewrite types in the real application. Already discussed in private emails, it seems there is no obvious way to generalize this, and so let's implement this special rule after #457 but before #383. This probably also needs syntactical support in #404.

favonia:
Closes #404. Closes #465.


src/redprl/operator.sml
```diff
@@ -150,6 +150,13 @@ struct
 
   (* TODO: move elsewhere *)
   datatype 'a selector = IN_CONCL | IN_HYP of 'a
+  datatype accessor = AT_TOP | AT_TYPE | AT_LEFT | AT_RIGHT
+
+  val accessorToStrung =
```
jonsterling: `accessorToString` 😜

favonia: Somehow I can survive with tab completion...

src/redprl/atomic_judgment.sig
```diff
@@ -62,11 +62,16 @@ struct
       *)
      | TERM of sort
   end
+
+  (* favonia: I do not like the current usage of "invariant" in many PLs,
+   *          so I coined the word "anti-variant". *)
```
favonia: (Trolling.)

favonia:
Note: the `rewrite-hyp` tactic remains the same, because it will be removed after #457 and it is annoying to maintain two copies of complicated code.

src/redprl/atomic_judgment.sig
jonsterling: lol

src/redprl/refiner_types.fun
```diff
@@ -2040,24 +2057,42 @@ struct
...
         val motivem = substVar (holeM, x) motiveHole
 
-        fun replace jdg = 
-          case jdg of 
-             AJ.TRUE _ => AJ.TRUE motiven
-           | _ => jdg
-
-        val (H', ajdg') = Selector.map sel replace (H, ajdg)
-        val (rewrittenGoal, rewrittenHole) = makeGoal @@ H' >> ajdg'
+        val replace =
```
jonsterling: @favonia Do you think that there is any way that this kind of code can be factored out? If it sounds like it is too difficult or won't help, don't worry about it.

favonia: I tried, but then decided to write out the actual program before refactoring. Any ideas?

src/redprl/atomic_judgment.sig
favonia: Do you think this is a good word? I wan thinking about "anti-symmetric".

src/redprl/refiner_types.fun
favonia: The difficulty is that the semantics is not independent of rewriting. For example, `SYNTH` should be dealt with in general map/project functions, but it is not mentioned here because it is unclear what the criterion of equality should be for a term with an unknown type.

PR #516, opened 2017-12-20
Rename SUB_UNIVERSE to SUB_KIND and add examples.

favonia:


src/redprl/atomic_judgment.sig
```diff
@@ -41,12 +41,12 @@ struct
       *)
      | SUB_TYPE of (abt * abt)
 
-     (* `SUB_UNIVERSE (a, k)`
-      *   `a` is a sub-universe of U^`k`.
+     (* `SUB_KIND (a, k)`
+      *   `a` is a sub-universe of the universe of `k` types at the omega level.
```
favonia: @jonsterling Please check that `SUB_KIND` is the proper name for this semantics.

jonsterling: I think it mostly makes sense. The only thing that is a bit weird is I would expect subkind to have two kinds as its arguments. Would it make more sense to say `HAS_KIND`? That is, the universe `a` can be regarded as having the kind `k`.

favonia: Wouldn't that be misunderstood as the universe itself being of kind `k`?

jonsterling: @favonia Haha, you are right 😉

favonia:
Any other comments? I would like to merge this because of the conflicts.

jonsterling:
go for it

Issue #383, opened 2017-09-24
Example: Omega1(S1) = Z


jonsterling:
I think that we now have enough stuff implemented to prove that the loop space of the circle is the integers. This would be a great example to try, and would provide a point of comparison with the Swedish implementation (https://github.com/mortberg/cubicaltt/blob/master/examples/helix.ctt).

jonsterling:
As a start, we can start proving some basic results about equivalences, such as "adjointification" (which takes an isomorphism to an VV-style equivalence): https://github.com/RedPRL/sml-redprl/commit/fe9ababa096e054df778b9ed0cf6bc135eebf524

I still need to fill in the hard part of this theorem, which looks like it will be a very gnarly nested composition (at least, if the Swedish version is any indication: https://github.com/mortberg/cubicaltt/blob/a331f1d355c5d2fc608a59c1cbbf016ea09d6deb/examples/equiv.ctt)

favonia:
We need #330 to reduce `($ f loop^2)` to `2` by pure magic.

favonia:
This was partially addressed by #420 and we need some machinery to proceed further.

mortberg:
When you have done this please let me know so that I can compare with what I have in cubicaltt. 

A nice corollary that you could do once you have this is that the loop space of S^1 is a set. To prove this just transport the proof that Z is a set along the pi1(S^1) = Z equality. How complicated is the obtained term?

favonia:
It would be great if we do not have to manually prove that
```
(S1-rec [_] (U 0 kan) (@ p 1) int [x] (@ IntSuccPath x))
```
is just
```
int
```

jonsterling:
@favonia I was looking again at the "theorem" you sent me the other day which was currently hard to prove, and I realized that it's not well-typed:

```
Thm IntToS1LoopToInt/succ : [
  (-> [n : nat]
   (= int
    ($ S1LoopToInt ($ IntToS1Loop (succ n)))
    (succ ($ S1LoopToInt ($ IntToS1Loop n)))))
]
```

It's not well-typed, because `nat` is no longer a subtype of `int`, and also `succ` doesn't apply to integers.

Now, I did try replacing with `int` and `IntSucc`, but I didn't make much progress.

Issue #438, opened 2017-11-11
Better interaction between tactics, dependency between goals, meta variables, etc


favonia:
I think we should revamp the LCF and the auto tactic to support maybe breadth-first search and better handling of dependency.

Firstly, a tactic should be allowed to return "depending on the realizers given to other goals", a temporary failure. To achieve this, we probably need to somehow mark all the meta-variables generated for other goals, and replace the one-line pattern-matching after `Abt.out` with something fancier. The idea is similar to stability checking is the machine. This also applies to the `match` and `query` primitive. A more refined approach is that a temporary failure would return the index of the metavariable it needs to know, and then the framework would prioritize solving the depended goal.

Secondly, once we distinguish permanent and temporary failures, we can then apply the tactics in arbitrary orders as long as we revisit the tactics which temporarily failed. This also made it possible to try to expand or discharge goals which need fewer or simpler subgoals according to some metrics. The idea is to detect permanent failure as soon as possible before wasting too much time.

Thirdly, after all of these have been done, I wonder if we can implement breadth-first search/trying also for the `par`, in the spirit of trying to detect permanent failure early.

jonsterling:
This sounds really interesting! I'm excited about your idea.

jonsterling:
@favonia So that I can better understand what you have in mind, can you give a little example? Particularly of the "temporary failure" stuff you mentioned. This sounds very interesting to me.

favonia:
@jonsterling If you apply `Bool.EqTT` to the goal `#1 = #2 in bool` where `#1` and `#2` are realizers of other goals, then it should temporarily fail and indicate that one of the blockers is #1 or #2 (or both) because it might succeed in the future but not right now, depending on these blockers. The distinction between permanent and temporary failures is to enable better proof search which prunes out branches with permanent failures.

For example, maybe in some branch, two goals are needed to be proved but there are many possible ways to prove the first goal. However, we can abort this branch if the second goal always leads to permanent failure no matter how the first goal is proved.

jonsterling:
@favonia Ahh, I see! This sounds like something I've always wanted, but I wasn't sure the best way to formulate it.

favonia:
Which part are you trying to formulate? The type (interface) of tactics or the proof search engine?

jonsterling:
@favonia Any of it 😄 

I think your Thought here sounds very immortal, and we should pursue it.

favonia:
[...]

PR #518, opened 2017-12-21
All for the Omega1(S1).

favonia:
Getting closer to #383.


favonia:
@jonsterling Try the following for a seemingly infinite loop. (You need `S1UnivCover` in the current development.) I wonder what is the best way to fix this? If I inline `Loop`, then the seemingly infinite loop would go away, but in a more complicated situation I would run into another seemingly infinite loop. Therefore I think it is important to figure out the actual cause. Perhaps we need something like #438 to do breadth-first search?

```
Thm Loop : [
  (path [_] S1 base base)
] by [
  abs i => `(loop i)
].

Thm S1LoopToInt : [
  (-> (path [_] S1 base base) int)
] by [
  lam l => `(coe 0~>1 [x] ($ S1UnivCover (@ l x)) 0);
    claim eq : [(= S1 (@ l 1) base)] by [auto];
    refine coe/eq; auto;
    [ fresh x -> rewrite eq at type; [`($ S1UnivCover x)]; auto
    , fresh x -> rewrite eq at left; [`($ S1UnivCover x)]; auto
    ]
].

Thm S1LoopConcat : [
  (->
   (path [_] S1 base base)
   (path [_] S1 base base)
   (path [_] S1 base base))
] by [
  lam p q => abs x => `(hcom 0~>1 S1 (@ p x) [x=0 [_] base] [x=1 [y] (@ q y)])
].

Thm S1LoopInv : [
  (->
   (path [_] S1 base base)
   (path [_] S1 base base))
] by [
  lam p => abs x => `(hcom 0~>1 S1 base [x=0 [y] (@ p y)] [x=1 [_] base])
].

Thm IntToS1Loop : [
  (-> int (path [_] S1 base base))
] by [
  lam i => fresh n ih -> elim i;
    [ abs _ => `base
    , `($ S1LoopConcat Loop ih)
    , `($ S1LoopInv Loop)
    , `($ S1LoopConcat ($ S1LoopInv Loop) ih)
    ]
].

Thm Test0 : [
  (= int ($ S1LoopToInt ($ IntToS1Loop 1)) 1)
] by [
  auto; ?
].
```

jonsterling:
Interesting... It's a bit concerning...

The auto tactic is so complex right now it feels like we have little control (or expectation) of what it will do. I wonder if what you suggested in #438 will help mitigate the issue; it seems a good idea regardless...

favonia:
@jonsterling Sorry I have been offline for the last week. What would be the recommended way to make progress? I wish we could finish this before the tutorial. I suppose adding some debugging functionality in the current LCF would help a lot.

jonsterling:
@favonia I'm now able to start digging into RedPRL again; would you like me to try and debug this?

favonia:
@jonsterling Yes, please. In particular, I think some support from the LCF library would help a lot.

We could merge this PR first, btw. Or I can start the internalization of EQ judgments.

jonsterling:
@favonia Some kind of debugging support in the LCF library would be awesome. What did you have in mind? Maybe I can implement it...

And you can feel free to merge this PR whenever you want.

favonia:
I want the ability to (1) put "RedPrlLog" into action and (2) ability to inspect the derivation tree, or at least know the location of a subgoal in it. How about https://github.com/RedPRL/sml-dependent-lcf/issues/40 and then turn each currently useless logging statement into some appending operation to the meta information? I want to know how exactly each goal is generated.

jonsterling:
@favonia Maybe I can implement a version of the LCF monad that instruments each proof state with a tree, so that you can see where everything came from.

favonia:
The above strange loop is more-or-less because when the type of `hcom` is a neutral term (containing an unfold symbol `Loop` in this case), `HCom.TubeL` will be tried first because `ComputeBefore` is useless. And then all the uncontrolled trouble...

Issue #435, opened 2017-11-11
Fix pretty printer of universe levels


favonia:

cangiuli:
Resolved by c6f7ba8655f4bb.

PR #519, opened 2017-12-22
Fix pretty printer of universe levels (#435)

cangiuli:
@favonia I took a first crack at this; please take a look. I have a few questions.

1. You have something more complicated in `univ_level.sml`, but I'm not clear on why this normalization code is necessary both on the input and the output. Do you want something fancier than what I've written, which already looks much better than the default pretty-printing?
2. Should we get rid of `RedPrlRawLevel.pretty` and instead use the main pretty-printer, once the latter is satisfactory?
3. Is there a `revMap` function lying around somewhere?
4. Why do the arguments to a `Thm` print as `i` and `j` in the argument list but `#i` and `#j` in the `Thm` body? Is this intended?

### DIFF
--- a/src/redprl/list_util.sml
+++ b/src/redprl/list_util.sml
@@ -28,6 +28,14 @@ struct
       go 0
     end
 
+  fun revMap f l =
+    let
+      fun go [] acc = acc
+        | go (x :: xs) acc = go xs (f x :: acc)
+    in
+      go l []
+    end
+
   (* From MLton: https://github.com/MLton/mlton/blob/master/lib/mlton/basic/list.sml *)
   fun splitAt (xs, i) = 
     let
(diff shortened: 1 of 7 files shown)


favonia:
I think your printer is much better than what's currently there. The normalization is basically an eye candy in pretty.sml, if there is any. My main criterion is whether the printed code can be parsed again, and this should answer your question 4 and imply that certain normalizations are optional. For question 2, yes, as commented in the code. What is `revMap` for? There is `list_util.sml` (or something like that) for generic helper functions. I hope I answered all of your questions.

cangiuli:
Thanks! Here are my remaining questions:

1. Do you still want to keep the fancy `pretty` you wrote in `univ_level.sml`? It seems preferable not to have two different pretty-printers for universe levels floating around.
2. `revMap` is because I currently do a `rev` of a `map` and feel bad about it. :)
3. Question 4 from above, about the current inconsistency (not caused by me) between `i` vs `#i`.

favonia:
1. Is in the same as `pretty o out` now? If so, I fully support the deletion.
3. Again, my main criterion is that the parser should be able to parse the output. You can change either the parser or the printer to make this happen. Perhaps I misunderstood your question?

PR #520, opened 2017-12-30
rename "variant" to "variance" for clarity

jonsterling:


jonsterling:
Actually, I would like to do this together with a more sweeping cleanup and refactoring pass, which I'll submit hopefully today.

PR #521, opened 2017-12-30
Cleanup and refactoring

### LINKED ISSUE #517: Refactor the variant stuff
Right now it is in `atomic_judgments.sml`. That does not make any sense.

jonsterling:
I felt that the RedPRL codebase was becoming too sprawling and that individual modules were intimidating. As a first step, I've started reorganizing the modules that have to do with syntax; in some cases, I have moved code into its own modules (like "variance" "selector", etc.).

I've renamed some things too, as well as made better use of SML's datatype replication mechanism.

This resolves #517.


src/redprl.mlb
```diff
@@ -73,7 +73,6 @@ in
 
 
     ann
-      "nonexhaustiveBind ignore"
```
favonia: Which version of MLton are you using? I believe this is a new feature introduced in 2017: (see the commit about their documentation https://github.com/MLton/mlton/commit/0908f6af4db86fd88461b358411a4488560838e2). If I understand it correctly, the `nonexhaustiveMatch` was broken into multiple annotations.

favonia:
I am thinking that maybe `Selector` and `Accessor` can be put into the same files because they kind of have the same purpose, but separating them is also good.

src/redprl.mlb
jonsterling: I guess my build is from 2016; I'm just using the one that you get in Ubuntu when you run `apt install mlton`.

favonia: I see. I was using a very recent version because I want https://github.com/MLton/mlton/pull/190.

jonsterling:
@favonia 

> I am thinking that maybe Selector and Accessor can be put into the same files because they kind of have the same purpose, but separating them is also good.

Yes, I had that thought too... the only reason I didn't was because I didn't know what to name the file :laughing:

src/redprl.mlb
jonsterling: Ah :smile:

favonia:
Wait, I meant you should revert commit https://github.com/RedPRL/sml-redprl/pull/521/commits/dd5bdafc48adbc5f1ddbd6aa97ecdedae18b82ba unless it stops the old MLton from compiling.

jonsterling:
@favonia I disagree; on the version of MLton that everyone except you likely has installed, it will cause a scary-looking warning during compilation. Unless it's absolutely necessary, I don't want our build to depend on versions of the compiler that are not easy to install using the standard packaging.

jonsterling:
If you feel really strongly about it, we can add it back, though.

favonia:
Alright. It's necessary for me to inspect possible bugs found by MLton, but I guess I can add it back on my local copy. FYI: Everyone needs to solve the PIE issue somehow, and Arch and its friends provide the updated `mlton`, while Debian/Ubuntu and their friends just patch the stable version of `mlton`. Hopefully after https://github.com/MLton/mlton/issues/147 this is no longer an issue.

favonia:
Oops. I meant to say that the distros which use the updated MLton already has the new annotation design.

PR #523, opened 2018-01-01
Set up LCF tracing infrastructure

jonsterling:


favonia:
I suppose `[refine fun/intro; ?foo]` will still print out the trace?

jonsterling:
@favonia Yep. Note that the trace goes back as far as the last transition from global-to-local (which occurs, for instance, when you use the `[t1, ..., tn]` multitactic.

favonia:
Maybe we can remove the completely useless `RedPrlLog` now?

jonsterling:
@favonia If you mean the completely useless calls to `RedPrlLog.trace`, these have all been removed (or at least, I think I've removed them all).

favonia:
Sorry, I meant `RedPrlLogUtil`.

jonsterling:
@favonia Ah! Thanks for the reminder; now deleted.

favonia:
Thanks to tracing I think I know what's going on (wrong, really).

Issue #382, opened 2017-09-15
Reduce arguments of equal types.


favonia:
How about reducing the TRUE judgments of `(= ty m n) true` into `(= ty' m' n') true`? This is to bridge the gap between an `EQ` judgment and an internalized equality.

jonsterling:
@favonia Such a rule seems semantically fine, but why not factor it through a roundtrip between the equality judgment and the internalized equality? that way we might not have to duplicate rules...

favonia:
Hmm I meant changing the current behavior.

favonia:
This will eventually be consumed by #393.

favonia:
Needs to be done to make #457 practical.

jonsterling:
👍

PR #524, opened 2018-01-01
Remove member equality judgments.

### LINKED ISSUE #457: Remove member equality judgments

favonia:
This shows how to close #457. Some adjustments to `reduce` are needed.

### DIFF
--- a/example/semi-simplicial.prl
+++ b/example/semi-simplicial.prl
@@ -38,7 +38,7 @@ Thm Pick/succ/0/succ/1 : [
   lam n m =>
     claim aux0 : [($ Pick n m) in (U 0)] by [auto];
     claim aux1 : [($ Pick (succ n) m) in (U 0)] by [auto];
-    unfold Pick; reduce; auto
+    unfold Pick; reduce at left in aux0; reduce at left in aux1; auto
 ].
 
 Tac Replace(#z, #A) = [
@@ -82,7 +82,7 @@ Thm Pick/compose : [
         ]
       ]
     ]
-  ]
+  ]; auto
 ].
 
 Thm Pick/compose/zero/0 : [
@@ -114,7 +114,11 @@ Thm Pick/compose/tt/4/0/tt/5/0 : [
       ] by [auto];
     claim aux1 : [ ($ Pick (succ a) c) type ] by [auto];
     claim aux2 : [ ($ Pick a) in (-> nat (U 0)) ] by [auto];
-      refine eq/intro; unfold Pick/compose Pick; reduce; auto
+    unfold Pick/compose Pick;
+    reduce at type left right in aux0;
+    reduce at left right in aux1;
+    reduce at left in aux2;
+    auto
 ].
 
 Thm Pick/compose/tt/4/0/ff/5/0 : [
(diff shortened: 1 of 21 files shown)


favonia:
The remaining issue is #382... how should we design the "`deep-reduce`"? If the conclusion is `(= ty m n)`, and the user wants to reduce `m`, what should the user do? One proposal:

1. `reduce in concl at left`

The issue is that it is unclear what `reduce in x y concl at left right` means, if we want to allow reducing at multiple spots. Or, we can just allow the following two tactics:

1. `reduce [x | concl]+`
2. `reduce [x | concl] at [left | right | type]+`

Thoughts?

jonsterling:
@favonia It seems like `reduce` should just take a bunch of `selector,accessor` pairs somehow. Do you think that would resolve the problem?

favonia:
The problem is the grammar design...

favonia:
How would you want to write down "reduce both sides of x (which is an EQ_TYPE) and the left side of the conclusion (which is an internalized EQ)"?

jonsterling:
What about `reduce x at left right, concl at left`?

favonia:
Well, I think we already removed commas...

jonsterling:
@favonia You are right...

And I'm sorry about the merge conflicts, I didn't think of that.

favonia:
Any other ideas? (Don't worry. The merge conflict is easy to solve.)

jonsterling:
> Any other ideas?

Well, my go-to solution for this kind of problem is to add more brackets or something, but that's not really ideal :laughing: 

Or we could have a notation like `reduce ((x|concl).(left|right|type))+` or something. In particular, your example would be written as follows:

```
reduce x.left x.right concl.left
```

or we could even have

```
reduce x.[left right] concl.left
```

if you felt that was better.

favonia:
@jonsterling Alright I implemented the proposed rewrite in #528.

favonia:
@jonsterling I am worried about periods being confusing... did we remove all other usages of periods?

favonia:
Hmm. Unfinished goals in `test/success/bool-fhcom-without-open-eval.prl` are very suspicious. They can solved by `auto` but `lam` somehow fails to solve them?

favonia:
Same things in `test/success/num.prl` now. `lam` is leaving goals that can be discharged by `auto`.

favonia:
Same in `test/success/universes.prl`. @jonsterling Is this something expected due to recent changes?

jonsterling:
@favonia This reminds me of a long-standing issue with `lam`, that we have the ability to solve using https://github.com/RedPRL/sml-dependent-lcf/pull/41/files

favonia:
Did I create too many aux goals in `Fun.True` or something?

test/success/V-types.prl
```diff
@@ -73,7 +73,7 @@ Thm IdV/Test2(#l:lvl) : [
    (= ty (coe 0~>1 [x] (IdV x #l ty) a)
          (coe 0~>1 [_] ty a)))
 ] by [
-  lam ty a => auto
+  lam ty a => reduce at left; auto
```
favonia: I wonder why this is needed? Somehow `ComputeBefore` stopped working.

favonia:
@jonsterling Strangely, by fixing some sort bugs, some lingering subgoals after `lam` disappeared. However, I do not feel comfortable with this kind of magical fix. It seems something deeper is still wrong. I might put in extra `auto` for now.

example/semi-simplicial.prl
```diff
@@ -82,7 +82,7 @@ Thm Pick/compose : [
         ]
       ]
     ]
-  ]
+  ]; auto
```
favonia: Ugly hack here.

test/success/V-types.prl
favonia: This was due to a bug in `AtomicJudgment` which is now fixed.

PR #526, opened 2018-01-01
Improve use tactic

jonsterling:
This makes the `use` tactic a bit less brittle. Previously if the type of the hypothesis being used didn't match exactly, it would throw a weird error. Now it will just give you the opportunity to prove that the thing you are "using" makes sense.


jonsterling:
@favonia Thank you for the review! And I agree :smile:

PR #527, opened 2018-01-01
New top-level signature language

jonsterling:
I'm going to blow away the old very gnarly and confusing signature modules, and replace it with a little  CBPV language. Hopefully I will a working version later today.


src/development.cm
```diff
@@ -1,4 +1,4 @@
 Group is
   $/basis.cm
   redprl.cm (bind:(anchor:libs value:../lib/))
-  frontend/frontend.sml
+  frontend/frontend.sml
```
favonia: I have been wondering how the final new line disappeared in your editor...

jonsterling: A mystery...

favonia: Maybe you could force your editors to add the ending newlines (?).

jonsterling: I bet there's a way for me to do that...

jonsterling: Update: I got it! :smile:

jonsterling:
I think there may be no real point to having this in that monad... I may extricate it. Not sure.

jonsterling:
This basically works in the broad strokes; now I need to re-do the way that cutting lemmas works to be basically less fucked up (payback for a design mistake I made some months ago), and things will work again.

jonsterling:
@favonia @cangiuli  Now this seems to work.

To summarize what I've done:

1. Ripped out the very gnarly signature "elaborator" that was in place before. I decided to replace it with a very simple programming language (the 'metalanguage') and evaluator, based on call-by-push-value together with a Pitts-style restriction and abstraction operator for metavariables; the nominal restriction operator is a computation, and the abstraction operator is a value. These are used to encode the 'second-order bindings' that are morally present when a user defines a theorem or a definitional extension wrt. some metavariable arguments.

2. Note that we still only parse the same old top-level commands; I have written a compiler that turns these commands into expressions in the metalanguage. Later on, we should just allow the user to write metaprograms, and provide the old top-level decls as convenience macros.

3. I've undone some horrific design mistakes that I made a few months ago in the rules and tactics for cutting lemmas. It's a bit hard to describe how wrong it went, but the old idea was that tactics would be used to full in the *meta-arguments* to a theorem. I thought that this would set the stage for implicits, but I was wrong in every possible way. Now things are much simpler: a theorem have its meta-arguments fully instantiated before it can be cut; as for implicits, it's already possible to write tactics that use unification to fill in the arguments in each specific case, but later on we could add more general support.

4. The behavior of `Print` and `Extract` is now identical; it always prints the theorem statement. I want to phase out the `Extract` command, and instead have an expression `extract M` in the metalanguage which projects the extract. Then, the user could write `Print (extract MyThm)` if they just wanted the extract... But it's not a huge priority right now.

cangiuli:
(4) sounds fine, but I still like the idea of providing a convenience macro `Extract M` => `Print (extract M)`.

jonsterling:
@favonia @cangiuli  you both make good points ;-)

jonsterling:
@cangiuli I made it so that Extract does the old thing; right now there are not really enough primitives to program this directly, but it will be easy to update when I expand the language more.

jonsterling:
@favonia @cangiuli How do we feel about merging this?

jonsterling:
Thanks for your review, @favonia and @cangiuli!

PR #528, opened 2018-01-01
Add the ability to reduce parts.

### LINKED ISSUE #382: Reduce arguments of equal types.

favonia:
Turns out I refactored lots of code before adding the new functionality... This is for #524. This closes #382. Kind of.

### DIFF
--- a/src/redprl/redprl.grm
+++ b/src/redprl/redprl.grm
@@ -468,6 +468,7 @@ end
  | selector of ast
  | selectors of ast list
  | accessor of ast
+ | accessors of ast list
  | opnames of string list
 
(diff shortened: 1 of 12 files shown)


favonia:
@jonsterling The grammar might be limited but the functionality is needed by #524. I suppose we could change the user interface later...

favonia:
An alternative syntax modified from @jonsterling's in #524:
```
reduce ((x|concl)(.(left|right|type))?)+
```

PR #536, opened 2018-01-03
improvements to Signature lang; improved error messages

jonsterling:


jonsterling:
@cangiuli This PR contains the fixes to those error message bugs you pointed out; I would have done it in a separate branch, but I didn't want to have to fix the code twice.

PR #539, opened 2018-01-05
[ML] function types

jonsterling:
This adds function types to the metalanguage / signature language. They're unused at the moment, but they will likely come in handy.

It also moves the metalanguage-identifier structure to another file.


jonsterling:
I'm going to merge this; the only non-metalaguage-related thing that this really touches is to fix the naming of fresh variables to be less confusing, a bug that was reported by @cangiuli.

Issue #472, opened 2017-11-21
The application tactics are extremely slow


cangiuli:
Costs 3 extra seconds on my machine for a single application. ???

favonia:
Is this still happening?

cangiuli:
If I check `examples/univalence.prl` with `HEAD`, it takes 1.37 seconds. If I prove the final `Univalence` theorem using `let foo = (RetIsContr (++ #l)) [arg, arg, arg...]; use foo` then it takes 2.13 seconds.

cangiuli:
When we comment out the `(v, AJ.EQ ((VarKit.toExp u, aptm), b'))` goal in function elimination, as an experiment, the runtimes go from 1.44 vs 2.76 to 1.41 vs 2.11. (I'm running off battery power right now, which probably explains the difference from the times I reported earlier.)

cangiuli:
FYI here's what we're testing:

```
Thm Univalence(#l:lvl) : [
  (->
   [ty/a : (U #l kan)]
   (IsContr (* [ty/b : (U #l kan)] (Equiv ty/a ty/b))))
] by [
  lam ty/a =>
  let foo = (RetIsContr (++ #l));
  fresh a b -> elim foo;
  [ `(* [ty/b : (U #l kan)] (Equiv ty/a ty/b))
  , fresh c d -> elim a;
    [ `(* [ty/b : (U #l kan)] (path [_] (U #l kan) ty/a ty/b))
    , fresh e f -> elim c;
      [ `($ (SigEquivToPath #l) ty/a)
      , fresh g h -> elim e;
        [ `($ (SigPathToEquiv #l) ty/a),
          fresh i j -> elim g;
          [ `($ (UARetSig #l) ty/a),
            fresh k l -> elim i;
            [ `($ (IsContrPath #l) ty/a),
              use k
            ]
          ]
        ]
      ]
    ]
  ]
].
```

PR #549, opened 2018-01-10
Don't search the context during 'auto'

jonsterling:
Surprisingly few proofs are broken by this, and these can be repaired easily. This makes all our proofs a lot faster to execute.

Related: #472

### DIFF
--- a/src/redprl/refiner.sig
+++ b/src/redprl/refiner.sig
@@ -13,6 +13,8 @@ sig
   val CutLemma : sign -> abt -> rule
 
   val AutoStep : sign -> tactic
+  val NondetStepJdgFromHyp : tactic
+  
   val Elim : sign -> hyp -> tactic
   val Exact : abt -> tactic
   val Rewrite : sign -> hyp Selector.t * Accessor.t -> abt -> tactic
--- a/example/omega1s1.prl
+++ b/example/omega1s1.prl
@@ -52,7 +52,8 @@ Thm IntSuccIsEquiv : [
     claim eq : [(= int ($ IntSucc ($ IntPred i)) i)] by [use IntSuccIntPred [`i]];
     unfold IntSucc IntPred in eq; reduce at left in eq;
     {
-      {use IntPred [`i], abs _ => `i}
+      {use IntPred [`i], abs _ => `i};
+       symmetry; unfold IntSucc; reduce at left; assumption
     , lam {i',p'} =>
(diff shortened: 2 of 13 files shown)


cangiuli:
On my machine, this makes the tests (including two copies of the univalence file, one of which uses `elim` rather than `exact`) run in 4.54 s instead of 5.36 s.

favonia:
The depth-first `auto` makes sense---think about all the unsolvable meta variables. This might also resolve some weird failures in the past as a side effect.

favonia:
If we are going to implement #438, hopefully the order of application of `auto` will be automatically closer to a depth-first search.

jonsterling:
We (@jonsterling and @cangiuli) have trained our Thought on #472 and resolved it. The problem was that the iterated application of elimination rules is very inefficient, because it induces numerous hypotheses of extraordinary size. To resolve the problem, we have created a "mega" elim rule that manages n-ary applications of functions, paths and lines all at once, producing only two new assumptions.

Closes #472 TASTE IT

favonia:
How about naming it as `App` instead of `Elim`? I am thinking about making it clear `nat-rec` etc are not included.

jonsterling:
> How about naming it as App instead of Elim? I am thinking about making it clear nat-rec etc are not included.

I don't know if I really like `App`, but I don't have any better ideas... If you or @cangiuli have another idea, I'd welcome it; otherwise we can change it to `App`.

favonia:
Alright, another thing is that personally I would remove `Elim = Multi.Elim 1` completely because `Path.Elim` can now technically solve function application and vice versa.

jonsterling:
@favonia We could delete all such `Elim` functions I suppose ,and just call `Multi.Elim` in the general `elim` tactic. @cangiuli what do you think?

jonsterling:
It was necessary to interleave stable evaluation with each application in order for this to be usable. (Consider the case where the synthesized type is a defined operator, which unfolds to a function type.)

src/redprl/refiner.sig
```diff
@@ -23,7 +23,8 @@ sig
 
   val Inversion : hyp -> tactic
 
-  structure Multi : 
+  (* synthetic elim rule for nested pi, path and line types *)
+  structure MultiArrow :
```
favonia: I like this name.

cangiuli: Me too!

jonsterling: yay!

jonsterling:
I'm going to merge this now, since it seems like everyone's happy with it.

Issue #354, opened 2017-09-03
Support nested multiline comments


favonia:
Something like `/* */`?

PR #555, opened 2018-01-12
add support for nested, multiline comments (close issue #354)

jonsterling:


favonia:
Yay! Could you update the Emacs mode as well?

jonsterling:
@favonia 

> Yay! Could you update the Emacs mode as well?

Carlo and I looked into it and it seems extremely complicated (wtf?). I don't want to do it right now; in any case, the emacs mode is not very usable and I prefer the vim mode.

PR #562, opened 2018-01-13
Examples

favonia:
I feel it is easier to track this branch with a PR.


cangiuli:
How much of this actually needs to be salvaged? I think most of it is either already in `master` or severely out of date.

favonia:
I do not feel there is anything left, but do not know the details either. So I would like to double check with you. @cangiuli @jonsterling

cangiuli:
I've now looked through carefully, and think we should scrap this PR and branch completely.

PR #564, opened 2018-01-13
give suitable names to free metas when saving a theorem

### LINKED ISSUE #547: Don't print internal metavariable names
The following code:
```
Thm Not : [
  (-> bool bool)
] by [
  lam b =>
  if b then `ff else `tt
].

Thm NotNot : [
  (->
   [b : bool]
   (= bool ($ Not ($ Not b)) b))
] by [
  lam b =>
  // The next two lines can be replaced by auto.
  unfold Not;
  if b then (reduce at left; refine bool/eq/tt) else (?)
].

Print NotNot.
```
prints out:
```
  [] =>
    Thm:
      (-> [b : bool] (= bool ($ Not ($ Not b)) b))
    
    Extract:
      (lam [b] (if b ax (#27 b)))
```

The `#27` here is an internal name that does not make sense to the user, and should not be displayed. @jonsterling and I think it would be better to sanitize the name entirely by replacing it with `???`. If anyone has a good idea for a helpful thing to name it, that might also be reasonable.

jonsterling:
Closes #547

### DIFF
--- a/src/redprl/metalanguage/elaborate.fun
+++ b/src/redprl/metalanguage/elaborate.fun
@@ -235,12 +235,12 @@ struct
      | ESyn.PRINT (pos, v) =>
        (ISyn.PRINT (pos, #1 @@ elabValue env v), Ty.UP Ty.ONE)
 
-     | ESyn.REFINE (ajdg, script) =>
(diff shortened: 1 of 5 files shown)


jonsterling:
developed w/ @cangiuli

PR #565, opened 2018-01-13
add "trace" tactic (close #113)

### LINKED ISSUE #113: add a "label" tactic
This tactic should name a subgoal so that it can be picked visually out from the proof state easily.

jonsterling:
closes #113


src/redprl/tactic_elaborator.fun
```diff
@@ -368,12 +368,18 @@ struct
...
+
+     | O.TAC_TRACE str $ _ => 
+       Lcf.rule o R.LabelGoal str
      | O.TAC_FAIL $ _ => fail "fail"
      | _ => raise RedPrlError.error [Fpp.text "Unrecognized tactic", TermPrinter.ppTerm tm]
 
   and multitactic_ sign env tm =
     case Tm.out tm of 
-       O.MTAC_ALL $ [_ \ tm] => T.all (tactic sign env tm)
+       O.MTAC_ALL $ [_ \ tm] =>
```
jonsterling: this is sad; one day it won't be needed.

src/redprl/refiner.fun
```diff
@@ -338,6 +338,14 @@ struct
     fun ReducePart sign = SequentReducePart sign
   end
 
+  fun LabelGoal (lbl : string) : rule = 
+    fn alpha => fn H >> jdg => 
+      let
+        val (goal, hole) = makeGoal [lbl] @@ H >> jdg
+      in
+        |>: goal #> (H, hole)
+      end
```
favonia: I hope this works for `MATCH` and `MATCH_RECORD` as well. Is there a reason to limit it to sequents?

src/redprl/tactic_elaborator.fun
favonia: I do not get it right a way. Could you possibly explain why it worked before?

jonsterling: This is very ugly and unfortunate; what is happening is that the parser adds coercions between tactics and multitactics in various places (because we try not to expose to the user the difference). The parser tries to collapse these coercions if they are redundant (for instance, a roundtrip); but in this case, the roundtrip came into being after a substitution, so it had not been crushed.

The reason this results in a different behavior vis-a-vis tracing is that the trace is part of the global proof state and not the local state (which is just a judgment). So, coercing a multitactic to a tactic causes it to lose the global trace, and coercing it back to a multitactic effectively resets the trace.

The solution is to collapse this.

src/redprl/refiner.fun
jonsterling: thanks! I've fixed it now.

favonia:
I decided to make some progress in semi-simplicial types first. Feel free to take this PR back (and merge it).

favonia:
Done.

favonia:
@jonsterling I was able to fix this PR by reverting the hack which collapses certain `MTAC_ALL`. Is it correct? By the way, we do not have an example showing the usage yet.

jonsterling:
@favonia Hmm, I don't know... The reason for that hack was to prevent the trace from getting obliterated.

Can you elaborate what you mean by `able to fix this PR`? What were you fixing in particular?

favonia:
@jonsterling I meant c7a569b failed and the testing seems to work by reverting the hack. I am worried that the trace is actually obliterated....

PR #568, opened 2018-01-13
Add path/eq/from-line rule and use it for groupoid laws.

cangiuli:
Work in progress...


example/groupoid.prl
```diff
@@ -0,0 +1,74 @@
...
+  [ fresh x:dim -> refine path/intro; 
+    [ exact (hcom 0~>y ty (@ p x) [x=0 [_] (@ p 0)] [x=1 [_] (@ p 1)]); auto; fail
+    , auto; fail
+    , auto; fail ]
+  , refine path/eq/from-line;
+    [ symmetry; refine line/eq/eta; auto; fail 
+    , auto; fail
+    , auto; fail ]
+  , unfold Trans; reduce at right; auto; fail ]
+].
```
jonsterling: ouch!

For those watching, we're getting murdered by backtracking. Carlo and my Thought is that in the new order, whatever that means, backtracking shouldn't be used for things like figuring out whether to call the rule for Line or for Path.

Issue #559, opened 2018-01-13
Consider adding motives to the terms for `if` and `nat-rec`


jonsterling:
I would like to propose or open for discussion the matter of adding motives to `nat-rec`, etc. Right now, it is not possible to automatically check membership for terms defined using `nat-rec`, nor is it possible to synthesize their types. This results in frustratingly requiring the user to intervene and write down a type in their tactic script, somewhere far away from where the `nat-rec` was defined.

I want to advance the Thought, **"It's OK If Hard Things Are Hard, But Easy Things Should Be Easy"**. A corollary to this Thought is that we should be able to typecheck routine uses of `nat-rec` that arise from proof-refinement.

One disadvantage of doing this is that term sizes will increase.

favonia:
I would like to see how this goes. In particular, if #534 is no longer needed, I would strongly support this.

favonia:
I think `nat-rec` and `int-rec` (with recursion, in general) are more problematic than `if`. I could be wrong.

jonsterling:
@favonia Can you give a it more detail about why `if` is not as problematic? I can believe that recursion can make things harder, but the problem I'm trying to solve is "Synthesize the type of this expression"; in general, I think this is just as impossible with `if` as it is with `nat-rec`, but I may be missing something.

favonia:
You can cheat---if `m0` is in type `ty0` and `m1` is in type `ty1`, then `(if b m0 m1)` is in type `(if b ty0 ty1)`.

jonsterling:
@favonia Yes, but that doesn't get us anywhere! We also can't necessarily synthesize `m0` and `m1`. The situation with `synth` and `if` basically the same as with `nat-rec` for this reason; with elimination forms of positive types, you can never synthesize generally unless you supply a motive. This corresponds to the fact that usually synthesis applies in left focus, whereas the positive eliminators usually are part of inversion.

favonia:
Did we reach some consensuses? This change is trivial to make, and I wonder if there is any objection.

jonsterling:
@favonia I think we should do it, in order to "make easy things easy"; but I don't think I ever got a final opinion from Carlo on it. @cangiuli what do you think?

[I have experimentally carried out this change in a branch, and together with some changes to how type synthesis works, it has enabled us to resolve a number of usability issues.]

cangiuli:
[...]

PR #570, opened 2018-01-14
Immortal solution to names

### LINKED ISSUE #542: Assign names after the introduction of variables.
I think I would prefer pushing names after the tactic. What it means is that
```
fresh n ih -> elim n,
[ zeroCase
, succCase
]
```
is replaced by something like
```
elim n,
[ zeroCase
, n ih => succCase
]
```
The names, `n` and `ih` were assigned just before the tactic. The use of `=>` has the extra bonus that
```
lam x => body
abs i => body
```
are really nothing special except there's no comma (and maybe scoping? not sure).

(EDIT: delete the `lam body` because `lam` actually takes a variant number of variables.)

jonsterling:
This is **_Jon-Carlo's Immortal Taste It Box for Arguments Unnamed Civilly (AUC)_**.

Resolves #542 

This is work in progress / proof of concept.


favonia:
Hi, I am still hoping that each rule can reset the stack and we can omit `pop` in the syntax.

jonsterling:
@favonia 

> Hi, I am still hoping that each rule can reset the stack and we can omit pop in the syntax.

Re: resetting the stack 'after each rule', I've tried to give some thoughts on that in the #542 thread; I'm not totally sure what is meant by omitting `pop` in the syntax. Did you mean that you just prefer to write `tac; [x => tac]` or something, rather than `tac; [pop x => tac]`? That certainly sounds plausible.

favonia:
Yes, I just meant the syntactical omission, but with the current semantics of non-local stacks, I believe it is important to have `pop` to remind users of what is going on.

src/redprl/redprl.lex
```diff
@@ -151,11 +151,10 @@ whitespace = [\ \t];
 <INITIAL>"claim"            => (Tokens.CLAIM (posTuple (size yytext)));
 <INITIAL>"use"              => (Tokens.USE (posTuple (size yytext)));
 <INITIAL>"with"             => (Tokens.WITH (posTuple (size yytext)));
+<INITIAL>"without"          => (Tokens.WITHOUT (posTuple (size yytext)));
```
cangiuli: My Thought is so powerful!

jonsterling:
Here's a cool example:
```
/* You can rename hypotheses using 'without' and 'with'; the scripts
 * have an appropriate binding structure, treating names pronominally.
 */
Thm RenamingHyps : [
  (-> bool int bool)
] by [
  lam x y => 
    without x y; 
    with a b => 
      use a
].
```

favonia:
Hi, I wonder what the status of this PR is? I am interested in resolving #559 but that might create a conflict with this PR.

jonsterling:
@favonia This PR still has some bugs in it, that Carlo and I are going to start fixing now that the LICS rush is over. We want to get it ready to merge very soon.

jonsterling:
@favonia As for the motives, any merge conflicts that result from that will be easier to resolve, so I don't feel too strongly about what order this is done in.

cangiuli:
@MatthewFluet You may be interested to hear about our latest MLton troubles (resolved by 5d2128d)!

8139c10 caused MLton to start taking multiple minutes to compile RedPRL (up from ~30 seconds), spent mostly in garbage collection during the `deepFlatten` pass. When we `-drop-pass deepFlatten` everything is back to normal, and this doesn't seem to affect either the binary's size or runtime. I'm attaching the `-verbose 3` output of compiling 8139c10 with and without `-drop-pass deepFlatten`.

[With deepFlatten: 64.55 + 172.43 (73% GC)](https://github.com/RedPRL/sml-redprl/files/1696661/log-flatten.txt) (`deepFlatten finished in 30.22 + 169.02 (85% GC)`)
[Without deepFlatten: 33.97 + 4.18 (11% GC)](https://github.com/RedPRL/sml-redprl/files/1696662/log-no-flatten.txt)

To be clear, everything's fine now that we drop the pass, but we thought you might be interested!

jonsterling:
Going to merge, but let's leave the branch up for a while.

MatthewFluet:
@cangiuli Thanks!  We've observed this problem in other programs (see the links at the bottom of http://mlton.org/DeepFlatten).

PR #571, opened 2018-01-16
Add the other connection to examples/connection.prl.

cangiuli:
I needed the other kind of connection (although not the correct diagonal) to prove something in `examples/tutorial.prl`, and I figured it should also be in `examples/connection.prl`.


cangiuli:
It looks like Travis had [some difficulties](https://www.traviscistatus.com/incidents/c1s4dlyxd4lf) today.

PR #573, opened 2018-01-17
A shortcut for :RedPRL.

favonia:


favonia:
The Emacs mode is using `c-c c-l` and I believe `l` stands for `load`. Anyways, I updated the READMEs. Do you think it is better now? I also updated the URL in the main README.

jonsterling:
nice idea! 👍

cangiuli:
In case I wasn't clear, I am happy for you to merge this at any time!

Issue #534, opened 2018-01-03
Special unfolding rules for recursive eliminators.


favonia:
It should be trivial to prove computational rules in systems with built-in recursive definitions. In particular, I would like to propose these rules (for `nat-rec` only).

```
P := blah in D
-----------------
P = blah in D
```

```
b in D
y, z >> c in D
x : nat >> p = nat-rec (x, b, y.z.c) in D
--------------------------------------------
p[0/x] = b in D
```

```
b in D
y, z >> c in D
x : nat >> p = nat-rec (x, b, y.z.c) in D
n in nat
--------------------------------------------
p[(succ n)/x] = c[n,p[n/x]/y,z] in D
```
The intended usage is that `p` would be some custom operator applying to some arguments.

jonsterling:
This would be nice. It's very sad that currently you have to unfold the operator and then you lose information (since it can't easily be folded back!).

jonsterling:
@favonia Another possibility might be to have a rule that gives these unfoldings directly an untyped manner, just like the existing `unfold` and `reduce` tactics/rules. Do you think that would work? It might lead to more efficient proofs.

favonia:
@jonsterling How should that be implemented? Remember that you have to crawl into the lambda binding, etc. It might be possible, but seems non-trivial.

jonsterling:
@favonia Maybe I misunderstood; it seemed like what you had proposed was at least as difficult, but perhaps not.

jonsterling:
Oh, by "applied to some arguments" you meant lambda-application, rather than operator arguments. This seems quite difficult in either case.

favonia:
In a typical and minimal case, I suppose the function would have the definition:
```
P := (lam [x] (nat-rec x ...))
```
and we want to relate `($ P (succ n))` and `($ P n)`. It should not be difficult to prove that
```
x : nat >> ($ P x) = (nat-rec x ...)
```
by `unfold` and `reduce`.

jonsterling:
Is there any way that we could provide these as a tactic rather than as a rule? Using a combination of `claim`and `unfold` and `reduce`? (a tactic written in ML) Or, if so, is it still better to have it as a primitive?

favonia:
The first rule can be easily done by claim, unfold and reduce, but it is so simple that I do not mind adding that as a rule. For the second and the third one, I do not really know---the possible way I know of involves dirty tricks to stop `reduce` in the middle which makes me a little bit uncomfortable....

[...]

PR #577, opened 2018-02-05
Add motives.

### LINKED ISSUE #559: Consider adding motives to the terms for `if` and `nat-rec`

favonia:
This will resolve #559. The example `example/semi-simplicial.prl` is not fixed yet, though.


favonia:
Great! Now it fails properly, as it should. :smile:

favonia:
@jonsterling One option is to comment out the failing part until we can do them almost automatically. There are instances of #534.

jonsterling:
@favonia Yes; I would endorse commenting it out. The proof is very nearly automatic once we deal with the `synth` stuff.

jonsterling:
@favonia (Check out my old exploratory branch which implemented both motives and the improved synthesis, https://github.com/RedPRL/sml-redprl/blob/improve-synth-improved/example/semi-simplicial.prl. I make no guarantees about correctness though)

favonia:
@jonsterling Hmm I thought I saw some commuting conversions which I doubt can be solved by synth, but the output's so messy and I could not really tell.

jonsterling:
Yes, it's really hard to tell... My feeling is that these are the kinds of proofs which we should not be doing: if we really have to reason with output that is this gnarly, we should just change the way that RedPRL works; the synth stuff helps in one respect, and also your idea for improved unfolding of recursive definitions would probably also help a lot.

jonsterling:
It may also be that in my experimental branch, I made some other changes to `auto` that made this work better. I just can't remember, because I was kind of delirious when I wrote it.

PR #579, opened 2018-02-05
propagate exit codes in scripts (close #578)

### LINKED ISSUE #578: travis testing succeeds even if some examples are failing
Like https://travis-ci.org/RedPRL/sml-redprl/jobs/337619065. A false positive. @cangiuli @jonsterling

jonsterling:
To resolve #578

### DIFF
--- a/script/test-mlton.sh
+++ b/script/test-mlton.sh
@@ -11,3 +11,4 @@ fi
 echo "Done!"
 
 exec ./script/run-tests.sh | ts -i "[%.ss]"
+exit "${PIPESTATUS[0]}"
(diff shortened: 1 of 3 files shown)


cangiuli:
I thought I tested this! Maybe I was only testing `run-tests.sh`? But your fix works for me.

Sorry for "fixing" the builds!

favonia:
I see! So the issue is the failure of propagating the error code in the pipeline...

jonsterling:
@favonia Does this look OK now?

favonia:
Personally I would declare `set -o pipeline` upfront, but this is also fine. There are several things I wonder why a default bash does not take as the default setup...

PR #582, opened 2018-02-07
More proofs about hlevels.

cangiuli:
Apparently RedPRL is much nicer than it was when @mortberg was working on these proofs last, so we were able to delete some cruft. This PR is a place for Anders to continue proving basic facts about hlevels.

It's sensible to merge this already, but let's leave it up to @mortberg: add as many commits as you'd like, and ask one of us for a review when you're done!


example/hlevels.prl
```diff
@@ -5,7 +5,7 @@ Def IsSet (#A) = [(-> [a b : #A] (IsProp (path [_] #A a b)))].
 
 Thm InhPropIsContr(#l:lvl) : [
  (->
-  [ty : (U #l kan)] // can I use pre here?
+  [ty : (U #l kan)]
```
favonia: To use "pre", one can use "stable" or just omit the kind.

cangiuli:
I had to update this PR to account for the Great Syntax Shift. I think we should go ahead and merge it before it becomes stale again, and then just open a new PR if we want to work further on this file.

PR #588, opened 2018-02-09
Semi-Simplicial Types: First Contact

favonia:


example/semi-simplicial.prl
```diff
@@ -146,3 +132,111 @@ Thm Pick/compose/ff/4/0 : [
...
+  with b'/ind b' => lam c => elim c;
+  [ lam d p0 p1 => elim p1 ]; // c = 0
+  with c'/ind c' => lam d => elim d;
+  [ lam p0 => elim p0 ]; // d = 0
+  with d'/ind d' => lam p0 =>
+  let {head = p0/h, tail = p0/t} = p0; elim p0/h;
+  [ lam p1 => let {head = p1/h, tail = p1/t} = p1; elim p1/h;
+    [ lam p2 => let {head = p2/h, tail = p2/t} = p2; elim p2/h;
+      [ rewrite ($ Pick/compose/tt/4/0/tt/5/0 b' c' d' p0/t p1/t) at left;
+        [ auto
```
favonia: @jonsterling @cangiuli I am not so happy about the unpredictable numbers of auxiliary subgoals due to `synthTerm` before the "interesting" goals. Not sure how to get around this issue. Even the tactic elaborator needs to know the exact numbers of auxiliary subgoals, unless we are fine with applying `auto` everywhere.

jonsterling: Hmmm, it’s a good point. We almost need some kind of way to differentiate subgoals (hierarchically?) so that it wouldn’t matter how many there are. 

Or another option is to require a prefix of the subgoals to be predictable, and put all the unpredictable ones at the end.

favonia:
We are so close now... I need dependent rewriting.

favonia:
Time to celebrate!

jonsterling:
Re: all this awful unpredictable subgoals stuff, I think that we can solve part of the problem by putting the returned subgoals *at the end* of the telescope. This is maybe not so important for structural equality rules that the user will rarely call themselves, but I think it is probably crucial for `DepRewrite`!!

favonia:
@jonsterling Fixed.

favonia:
@jonsterling Under the new order you have to change your review so that I can merge this PR... unless you have other comments.

PR #596, opened 2018-02-12
Add comments to example/semi-simplicial.prl

favonia:
I think this file deserves better documentation.

By the way, what is the simplest criterion of "correctness"?

- [x] recursive presentation


jonsterling:
By the way, I'm not sure if it is completely germane to this pull request, but I have been wondering. After doing this formalization, do you have any Thought about what the most annoying parts of it were? It would be cool to improve RedPRL along those lines.

I was imagining that the worst parts were the issues with unfolding recursive definitions (#534) and the lack of implicit arguments (especially when rewriting...). In Coq rewriting is very easy: you just write "rewrite foo" and it figures out the maximal way to do the rewrite using unification!

example/semi-simplicial.prl
```diff
@@ -194,6 +205,35 @@ Thm Pick/compose/assoc : [
...
+
+// well-typedness of proj requires proj-coh, which uses exact equality.
+
+proj-coh : (-> [p : nat] [X : sst p] {n m o : nat}
+  [c1 : ($ Pick m o)]
+  [c2 : ($ Pick n m)] [f : ($ folder p X n)]
+    (= ($ folder p X o)
+      ($ proj p X c1 ($ proj p X c2 f))
+      ($ proj p X ($ Pick/compose c1 c2) f)))
+*/
```
jonsterling: It would be so sweet if we could just write code like the above in redprl.... 😿

favonia:
Hmm...

1. Unification for writing the expressions, but we need to rework the entire framework.
2. Rewrite in the opposite direction (#609).
3. Some special support for recursive definitions---not sure what exactly I want, though.

PR #601, opened 2018-02-13
put Synth under a signature

jonsterling:


src/redprl/refiner_types.fun
```diff
@@ -53,169 +53,170 @@ struct
...
+      in
+        (List.rev psi, ty)
+      end
   end
 
+
+
+
+
+
```
favonia: Just curious... are there secret messages hidden in these lines?

jonsterling: lol, I can delete them 😉

PR #603, opened 2018-02-13
Performance tweaking (make RedPRL twice as fast on tricky files)

jonsterling:
These changes make RedPRL more than twice as fast at executing the semisimplicial types development and the univalence development.

The main thrust of the change is to change how the `PROGRESS` tactical is implemented. This tactical is supposed to succeed if the tactic makes the proof "move forward" (whatever that means), and it fails if not.

Previously to see if the proof had "moved forward" we did two things that were fairly bad:

1. To compare the new proof state to the old sequent, we would walk the new proof state to see if it contains a sequent that is equal to the old one. Now, we just check if the new proof state contains only one sequent, namely one that is equal to the one we started with.

2. Previously two sequents were considered equal if they were alpha equivalent, regarding hypotheses as introducing bindings. Now, two sequents are considered equal if they are literally the same.

I did not realize before that the very fancy style of comparison that we did before was actually unnecessary; with the simple-minded version that I now have, everything still works fine.

You can see the performance difference here: https://gist.github.com/jonsterling/a27c74be1d478211484f7b3530f5f5a4

### DIFF
--- a/src/redprl/judgment.sml
+++ b/src/redprl/judgment.sml
@@ -17,9 +17,8 @@ struct
     open S
     infix >>
   in
-    val rec sort =
-      fn H >> catjdg =>
-           (Hyps.foldr (fn (_, jdg, r) => AJ.synthesis jdg :: r) [] H,
-            AJ.synthesis catjdg)
+    fun sort (H >> atjdg) =
+      (Hyps.foldr (fn (_, jdg, r) => AJ.synthesis jdg :: r) [] H,
+       AJ.synthesis atjdg)
   end
 end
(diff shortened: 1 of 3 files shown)


favonia:
By the way, how did you realize this might be a spot to work on?

jonsterling:
@favonia Well, I wanted to make things less horrifically slow, so I started auditing the tactics for things that looked expensive... In particular, because operations like renaming and substitution in the ABT library are fairly expensive, I looked for anything that induced a lot of substitutions and renamings.

favonia:
Are you using the built-in profiling tool to audit the tactics? I would like to learn and help more next time.

jonsterling:
@favonia I have found profiling personally to be fairly useless for this kind of thing (but this may be just because I don't know how to use its full potential, I don't know); what I do is much more simple-minded. Instead, I look at the code and think really hard about it; then make a hypothesis and test it against our existing examples. It's not super scientific...

PR #606, opened 2018-02-15
Fix branchOnDim on meta variables. Close #605.

### LINKED ISSUE #605: Handling of meta variables in branchOnDim
Perhaps it is just... wrong? Now I do not think we can/should draw any conclusion from a meta variable.

favonia:
Line types in action.


favonia:
I think this PR will be correct independent of what we want to do about meta variables. The core issue is whether, for example, `(loop #i)` should be treated as a value. This PR turns it into a neutral term.

jonsterling:
I think it's fine, thanks!

Maybe we need another category of terms beyond "neutral" to express that a term's status is really contingent... But for now this is fine.

PR #607, opened 2018-02-15
Implement V.EqProj and kill autoSynthesizableNeu.

favonia:


src/redprl/refiner.fun
```diff
@@ -339,21 +340,6 @@ struct
         go
       end
 
-    fun autoSynthesizableNeu sign m =
```
jonsterling: excellent

favonia: Thanks, but you also have to formally approve it if you like it. ;-)

src/redprl/list_util.sml
```diff
@@ -60,6 +60,11 @@ struct
   end
 end
 
+infixr 5 ?::
+val op ?:: =
+  fn (NONE, l) => l
+   | (SOME x, l) => x :: l
```
favonia: By the way I cannot stop laughing at this operator...

PR #611, opened 2018-02-16
Fix the error message for failing non-det search.

favonia:


favonia:
I think this is called by `assumption`. Perhaps we can say `No useful assumption found in the context`?

favonia:
Actually your message works well in either case.

Issue #589, opened 2018-02-09
Vim mode: show RedPRL is running


favonia:

cangiuli:
Yes, I completely agree! Do you have any UI suggestions?

favonia:
Maybe the reminder can be displayed in a separate window as the error message when there is no message at all? Usually it can go to the status bar but my bar is kind of full already.

favonia:
If you have a clever way of notifying the user (such as... changing the buffer name to "RedPRL is running"???) that would work with common plug-ins, that might be a good option.

jonsterling:
This would be awesome!

Issue #575, opened 2018-02-05
Correctly highlight all the usage of the reduce tactic


favonia:

favonia:
@cangiuli How possible it is to highlight the fancy grammar for `unfold` and `reduce`?

cangiuli:
I can take a crack at it. The best way I can think of doing this is to have invisible regions for `unfold` and `reduce`, and to have special highlighting within those regions.

Can you explain what the possibilities are for an `elimTargetAnySort` selector?

favonia:
Not sure what `elimTargetAnySort` mean, but a selector should be a hypothesis or `concl`, and an accessor should be `left`, `right` or `type`.
```
unfold Blah Wow Yeah at right in concl
unfold Blah Wow Yeah in x at right left
unfold Blah Wow Yeah at type
unfold Blah Wow Yeah in x concl y
unfold Blah Wow Yeah
reduce at type
reduce at right in concl
reduce in x at left
reduce in concl x y z
reduce
```

cangiuli:
I'm referring to the following code in `redprl.grm`:
```
selector
  : CONCL (Ast.$$ (O.SEL_CONCL, []))
  | elimTargetAnySort (Ast.$$ (O.SEL_HYP, [Ast.\ ([], elimTargetAnySort)]))
```

and `unfold IntPred in eq1` in `example/omega1s1.prl`.

favonia:
`eq1` has the same grammar as `x` in `elim x`. `elimTargetAnySort` is just some hack to make it peacefully coexist with sort-checking. Does that answer your question?

cangiuli:
I think so. The difficulty with syntax highlighting this is figuring out where the `reduce` and `unfold` grammar ends. We could say that it ends whenever `;` or `)` or `]` is encountered. Does that sound reasonable?

favonia:
I have to think about it. It could also be `,` as far as I can see.

favonia:
Actually, now thinking about it, it could be `;`, `)`, `]`, `else`, `|`, `}`, `,` or maybe something weird. So, maybe just `\<else\>` or something out of [a-zA-Z0-9_]?

cangiuli:
What do you mean "something out of `[a-zA-Z0-9_]`"?

favonia:
`[^a-zA-Z0-9_]`

cangiuli:
[...]

PR #615, opened 2018-02-16
Improve vim mode.

cangiuli:
I'm going to improve the vim mode in a number of ways, including better syntax highlighting. Now is a good time to make any feature requests (beyond #575 and #589).


favonia:
1. Sometimes when the error points to a line that needs scrolling, Vim fails to redraw the screen properly... how should I help debug?
2. Is it possible to make each message foldable?
3. Could we have some command as fancy as "check theorems and definitions to the current one as if `Quit.` was inserted after the current block"?
4. After implementing #589, some way to manually kill it.

favonia:
5. I am not sure, but maybe it is cool to highlight the numerals...

cangiuli:
Thanks! Regarding (1), @jonsterling has also reported this, although I'm not sure I've run into it myself. I can try adding a `redraw` and see if it helps.

favonia:
An alternative to #589: shows that RedPRL has done the checking.

cangiuli:
@favonia Is this the behavior you wanted with respect to foldable messages? (Check a RedPRL buffer and then type `zM` in the messages buffer.)

favonia:
@cangiuli I was thinking about folding _error_ messages when some are them are very long, but what you did is definitely great!

cangiuli:
@favonia Try out `:call CheckBufferToCursor()`, which checks the buffer only up to the line that the cursor is on. You asked for a similar function but to the end of the "current block," but I'm not sure how to detect that. Do you have any suggestions? Is the current behavior useful?

cangiuli:
@favonia I think I can also fold error messages, but I actually don't know how to produce very long error messages in RedPRL! Can you give me a hint about how? :)

favonia:
I remember it's something of wrong sort of something. You can probably get one by screwing up one of the rewrites in the semi-simplicial types file.

PR #616, opened 2018-02-17
Starting our documentation!

cangiuli:
:tada:


cangiuli:
@jonsterling @favonia To build this, install the Sphinx documentation generator (on Ubuntu, the `python-sphinx` package) and try `make html` in the `doc` directory. The default style is not sufficiently revolutionary, but we can worry about that later. We can also generate LaTeX (again, currently rather ugly).

Any suggestions as to how to structure the documentation?

favonia:
Perhaps we can do https://readthedocs.org/...

favonia:
> Any suggestions as to how to structure the documentation?

I suggest starting writing now and then we would know how to structure it.

cangiuli:
> PS: Perhaps we can set up https://readthedocs.org/...

Fine with me, let's ask @jonsterling. RTD uses Sphinx, so I think we should proceed with what I've got whether or not we choose to host on RTD.

jonsterling:
I suggest that at least for now we get readthedocs fired up! That way we can have something that works immediately... If we decide to switch to something else then we can do so later.

@favonia @cangiuli

jonsterling:
I'll get it set up

jonsterling:
Yes... @cangiuli Can we merge this?

jonsterling:
Here's the stub: http://redprl.readthedocs.io/en/latest/

favonia:
> Here's the stub: http://redprl.readthedocs.io/en/latest/

Is it important to put files under `docs` instead of `doc`? I thought that is configurable.

jonsterling:
@favonia It will use either `doc` or `docs` automatically.

@cangiuli says I can merge this so I'll do that now

PR #623, opened 2018-02-22
Change license to assign copyright to the RedPRL Development Team

jonsterling:
I wonder if this makes the most sense? I think it's permissible to do this, but just wanted to check that everyone was OK with it.


jonsterling:
Thank you for the feedback!

favonia:
I wonder what could be an issue for using the RedPRL team as the copyright owner.... we probably need a lawyer here. In any case, a very formal approach is to ask everyone to sign some copyright assignment like this https://cla.github.com/. I'm using the review mechanism as a lightweight, less formal alternative. @jonsterling Maybe you can request reviews from past significant contributors.

jonsterling:
@favonia Whenever you're happy with this PR feel free to merge!

favonia:
@jonsterling My only worry is the past significant contributors---who else should we ask, if any? Otherwise I'm done.

jonsterling:
@favonia Well, since nobody else was assigned copyright, I guess it's not technically a problem... But thinking back the main other significant contributor was @wilcoxjay, who provided the first implementation of the typing rules for Kan operations.

favonia:
Thanks. Well, this is not an urgent issue and we have plenty of time.

wilcoxjay:
I'm happy with this.

favonia:
Thank you all for your responses.

PR #625, opened 2018-02-22
Continue with documentation

### LINKED ISSUE #622: Make Pygments Immortal
Ideally, a parser for RedPRL. http://pygments.org/docs/lexerdevelopment/

jonsterling:
(EDIT by @favonia: this closes #622)


favonia:
@jonsterling Infrastructural suggestions:

1. Enable readthedocs.org for branches.
2. Use http://doc.redprl.org or something like that. See http://docs.readthedocs.io/en/latest/canonical.html.

jonsterling:
@favonia That'd be nice; I don't think it is possible to have it automatically build all new branch, but I have now set it up so that it will build the 'docs' branch.

doc/src/language.rst
```diff
@@ -176,8 +176,38 @@ programming constructs inspired by cubical sets:
...
++-----------------------------+--------------------------------+
+| path/line abstraction       | ``(abs [i j...] e)``           |
++-----------------------------+--------------------------------+
+| path/line application       | ``(@ e r1 r2...)``             |
++-----------------------------+--------------------------------+
+| dependent record type       | ``(record [lbl... : A]... C)`` |
++-----------------------------+--------------------------------+
+| tuple (record element)      | ``(tuple [lbl e]...)``         |
++-----------------------------+--------------------------------+
+| record projection           | ``(!lbl e)``                   |
```
favonia: This table will be many people's favorite! I wonder if we should put a space after `!`. Maybe a space before `...` as well.

favonia:
@jonsterling By the way, I feel the table would look better if the separators between fields are dimmer. Right now they stand out too much on my screen. I have not regained the brainpower to to CSS hacking yet.

favonia:
By the way, regarding my complaint about the bottom margin of `<p>`: it seems like a general problem of the RTD theme, but it actually looks okay in other documents. I suspect that syntax highlighting is an important factor---without it, the `<pre>` blocks look so "dry" that they do not deserve much space. Still, I think a tighter bottom margin might look better.

PS: for a moment I thought travis is broken again, but then I realized it's really catching a "bug". Yay!

cangiuli:
@favonia I just pushed a Pygments lexer for RedPRL source. There's much room for improvement, but if it's to your liking, add the magic words to this PR to close #622.

favonia:
Yes, we can close #622 now.

jonsterling:
I think that the addition of the highlighter is monumental enough that we should consider merging this today (?).

favonia:
I agree! (BTW, somehow I still feel the space above the box is still too big---I wish at least the space above and below a code box can be the same.)

jonsterling:
@favonia Yes, the space doesn't seem quite right at the moment...

Issue #627, opened 2018-02-24
Prove the computation rule for J


mortberg:
Prove this and add to: https://github.com/RedPRL/sml-redprl/blob/master/example/J.prl

One should do it both for the direct definition of J in the file above and for J defined using contractibility of singletons.

mortberg:
Yes, the computation rule will only hold up to a path. It seems surprisingly complicated to prove this.

favonia:
What should be the statement, actually? There's no reason to believe the endpoints would match. Maybe we should use connection instead and allow at least coe along constant families.

mortberg:
Here's the statement and proof in cubicaltt: https://github.com/mortberg/cubicaltt/blob/master/examples/prelude.ctt#L82

I don't believe the proof goes through as in cubicaltt as the connection in RedPRL is defined using a complicated hcom and there is no reason for it to go away if you plug in refl for p. But as you say it might still be easier to get the proof to go through using the connection as you are then only coercing in a constant family. This is what I meant by it being "surprisingly complicated" to prove this in this system.

favonia:
I wasn't making it clear---the constancy is necessary to make things type check. BTW I think the connection hcom is okay.

favonia:
Actually this turns out to be straightforward in the sense that you can just follow the typing requirement to complete the composition. Is there anything special about the contractibility of singletons? My current understanding is that it is exactly the same as using connections, and if `J` was done using connections then there is nothing to do in `J/comp`.

mortberg:
Can you do the other proof as well please? Preferably using your complicated connection from https://github.com/RedPRL/sml-redprl/blob/master/example/connection.prl#L1

I think that the problem is that J defined this way is  

``coe 0->1 (<i> fam (p @ i) (connectionAnd A a x p @ i)) d``

so that when you plug in `<_> a` for p you end up with a similar problem as with the other J: you cannot prove the equality using a `coe 0->j` as the family in the term is not constantly `fam a (<_> a)` as required for the computation rule for J...

favonia:
> you cannot prove the equality using a coe 0->j as the family in the term is not constantly fam a (<_> a)

You are right. I did not think it through carefully.

mortberg:
I also thought that this proof would just go through the same way as in cubicaltt, but the fact that the family has to be constant makes it a lot more subtle. This is why one should always formalize everything :)

favonia:
@mortberg Should we close this now?

[...]

PR #629, opened 2018-02-25
Computation of J.

favonia:
#627. I worried too much---this is straightforward (in the sense that almost every step is guided by the typing requirement and no creativity is involved, not that the whole expression is as succinct as in CCHM).

### DIFF
--- a/example/J.prl
+++ b/example/J.prl
@@ -43,12 +43,11 @@ Thm J/comp(#l:lvl) : [
     ($ (J #l) ty a fam d a (abs [_] a))
     d))
 ] by [
-  lam ty a fam d =>
-    `(abs [k]
-      (com 0~>1
-       [i] ($ fam (J/comp/cube i (dim 1) k ty a)
-            (abs [j] (J/comp/cube i j k ty a)))
-       d
-       [k=0 [i] (J/coe i ty a fam d (abs [_] a))]
-       [k=1 [_] d]));
+  lam ty a fam d => abs k =>
+    `(com 0~>1
+      [i] ($ fam (J/comp/cube i (dim 1) k ty a)
+           (abs [j] (J/comp/cube i j k ty a)))
+      d
+      [k=0 [i] (J/coe i ty a fam d (abs [_] a))]
+      [k=1 [_] d])
 ].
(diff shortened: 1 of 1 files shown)


example/J.prl
```diff
@@ -14,13 +22,33 @@ Thm J(#l:lvl) : [
...
+   [ty : (U #l kan)]
+   [a : ty]
+   [fam : (-> [x : ty] (path [_] ty a x) (U #l kan))]
+   [d : ($ fam a (abs [_] a))]
+   (path [_] ($ fam a (abs [_] a))
+    ($ (J #l) ty a fam d a (abs [_] a))
+    d))
+] by [
+  lam ty a fam d =>
+    `(abs [k]
```
jonsterling: Does it work to use `abs k =>` here?

favonia: Sure.

favonia:
@mortberg Thanks for the review. Again, personally I do not feel any conceptual complexity because most steps are forced, but maybe you were saying the length of program is longer than the proof in CCHM? I agree it is longer. About the alternative J---the proof should work because it has nothing to do with the computation of connections. The faces of connections, not the computation of them, are the things that matter.

mortberg:
About J defined using connections see my comment on the issue. I would be happy to be proved wrong! :)

By more complicated I mean that the length is a lot longer yes. My experience when formalizing in cubicaltt is that this kind of complications due to lack of regularity gets very annoying and tedious. This is really the simplest thing one should prove about J, so when doing much more complicated things using it one might have much bigger troubles because of things not disappearing by computation. I think you will see what I mean if you formalize pi4(s3). At some point I had to spend a few hundred lines just to get rid of things that should have disappeared by regularity... My conclusion was that one should never use J in cubicaltt unless one really has to.

favonia:
Unfortunately, as you pointed out in the issue #627, it still needs extra fixing.

I totally agree with the comment about J. I believe this J is to satisfy the followers of intensional type theory and/or the book HoTT, not that it itself is particularly important to our development of other theorems. Seriously, just do the composition! The "true" J in my opinion lies in the different type that is inductively defined by refl, which will be there once Part IV is implemented in RedPRL. For that "true" J, things will compute definitionally.

PR #631, opened 2018-02-27
More documentation.

cangiuli:
Yet another PR for documentation.


favonia:
Should we use the style adopted in the actual RedPRL output, that is, horizontal lines?

jonsterling:
@favonia I don't think it's analogous; the horizontal lines in RedPRL's output distinguish a context from a conclusion within a sequent; but in a presentation of rules, we need a compact way to show a sequent on a single line, so `>>` or something analogous seems to be what we need there.

Sorry if I misunderstood your suggestion!

favonia:
Oh, indeed. My suggestion did not make sense. That said, I am worried about `|`---are we using the pipe in the tactic language?

jonsterling:
@favonia The pipe here is to distinguish the subgoals, as a notational convention. It's kind of like how indentation is used in the Nuprl rule presentation.

I think it's irrelevant whether we are using it in the tactic language, because there is no chance of ambiguity.

jonsterling:
Unless you are just worried about using the same lexer for both! That's a valid concern, but I'm confident that we can resolve the problem easily if it becomes a problem. Personally, I may not even want to document the `ORELSE` tactical, seeing as we don't use it in any developments (outside the internals).

cangiuli:
If it becomes a problem we can easily fork the lexer. However, I am currently matching only against `|` when it is the first character of a line.

favonia:
By the way, `#k type at #l` is no longer accurate. You are either a member of some universe, or just `#k type`. (Most (all?) rules handle both uniformally.)

cangiuli:
I think that's unnecessarily confusing. Maybe we should have a remark at the top saying that there are also "omega" versions of the rules.

cangiuli:
I synthesized the Overview page and redprl.org in a first cut at making the front page of the documentation into a suitable replacement for redprl.org. Please edit as you see fit.

I think it would help to apply some graphic design: it looks incredibly ho-hum right now. Ideas?

jonsterling:
@cangiuli  Re: the ho-hum-ness, I'm not sure it's a huge problem? Certainly not against making it better, but I am hoping that we can *turn down* the propaganda a bit in relation to the old page...

favonia:
I want "We can prove it!" even without the reference to communist propaganda.

doc/src/conf.py
```diff
@@ -102,7 +102,7 @@
 html_static_path = ['static']
 
 def setup(app):
-    app.add_stylesheet('fonts/PragmataPro.css')
+    app.add_stylesheet('http://cs.cmu.edu/~jmsterli/fonts/PragmataPro.css')
```
favonia: Hmm... why are we using absolute paths? (Or just that you accidentally committed something?)

jonsterling: @favonia I'm hosting the fixed webfonts on my AFS rather than on this repo; the reason is that, I realized that to really comply with Fabrizio's license, I need to place the webfonts behind an htaccess barrier.

favonia: Perhaps we can get something like http://www.cs.cmu.edu/~redprl/.

Issue #574, opened 2018-01-30
Implement th grammars and judgments in Part IV.A.3


favonia:

favonia:
Should the labels/constructors be capitalized like CustumOp, or do we prefer other marking? @ecavallo @jonsterling A bigger issue is, how should we declare a new inductive type? I suppose we want to extend signatures? Any architectural comments?

favonia:
@ecavallo I wonder if you already have some ideas how these should be implemented? Jon and I can probably hack up something, but your inputs will be very useful.

jonsterling:
I did rewrite the signature stuff, but that was a long time ago

ecavallo:
Here are some scattered thoughts about the implementation:
1. An inductive type declaration should probably be its own kind of signature declaration. The annotations on constructor terms can then just contain the name of the inductive, instantiations of its parameters, and the constructor label; when you need the constructor data you can look it up in the signature.
2. The fact that a declaration is well-formed can require reasoning, so maybe a declaration should be paired with a proof of that? Or maybe that's a separate theorem. There's a question here of how much you want to expose the well-formed constructor list judgment to the user. (Probably less is better than more.)
3. There should be two kinds of constructors: ones with and ones without a specified boundary. (If a constructor has a boundary, the set of constraints needs to be valid.) In the paper only 0-constructors are allowed to have no boundary, but there's no real reason for this. (Incidentally, make sure you have v2 of the preprint, I forgot to account for 0-constructors in v1.) For constructors with a boundary, coe does endpoint correction, for ones without it doesn't.
4. I'm not sure how you want to handle the transition from formal schematic terms to ordinary terms... I imagine it's expensive to redo the interpretation whenever you step to a boundary and so on. I guess they could be stored in interpreted form in the signature.

Some of these things are different from the paper, but not in ways that have significant impact on proofs. I don't have any opinions about the surface syntax. I'm trying to focus on getting something on untyped computation out right now, so I don't want get too involved at the moment, but I'll try to answer any questions of course.

favonia:
Perhaps it makes sense to name constructors as `TheInductiveType.constr` or something like that. Then we know where to look up for `constr`. For example there would be `S1.base` and `S1.loop`.

favonia:
[...]

PR #636, opened 2018-03-04
Inductive types, without fancy indexes or fancy kinds.

favonia:
Address #574.

- [x] Arity calculation
- [x] Elaboration
  - [x] Arity filling during elaboration
- [x] Parsing
- [ ] Machine:
  - [ ] intro
    - [x] boundary helper functions
  - [ ] coe
  - [ ] hcom
- [ ] Rules and the auto tactic
- [ ] Cache arity and equations?


favonia:
@jonsterling @cangiuli @ecavallo There are lots of design issues about the recursor. Here is one. Suppose there are three recursive variables `a`, `b`, `c`, and now we are defining the eliminator `f`. Should the context be `a`, `f a`, `b`, `f b`, `c`, `f c` or `a`, `b`, `c`, `f a`, `f b`, `f c`?

src/redprl/syntax/operator.sig
```diff
@@ -36,6 +36,16 @@ struct
...
    | COEQUALIZER | CECOD | CEDOM | COEQUALIZER_REC
+
+   (* inductive types *)
+   | IND_RECTYPE_SELF | IND_RECTYPE_FUN
+   | IND_RECTERM_INTRO of opid
+   | IND_RECTERM_FCOM | IND_RECTERM_LAM | IND_RECTERM_APP
+   | IND_CONSTRUCTOR of {label : opid, ndim : int, nvar : int, nrecvar : int}
+   | IND_ELIM_MK_CASE of {label : opid, ndim : int, nvar : int, nrecvar : int}
+   | IND_INTRO of {indtype : opid, label : opid}
+   | IND_ELIM of {iondtype : opid}
```
ecavallo: Did you mean `indtype` here?

ecavallo:
This is great, thanks Favonia!

I don't have a preference for eliminator context order...

src/redprl/syntax/operator.sig
favonia: Yes. I was forcing SML/NJ to recompile and then...

favonia:
@ecavallo By the way, is there a reason that `func` cannot be a meta-function? I didn't put it into the syntax because I believe it can be done that way.

jonsterling:
pretty exciting! (BTW I don't really have a preference about order of assumptions... maybe we will find that one version is easier than the other later)

favonia:
@jonsterling Could you remind me again why there is no reasonable view for `CUST`? If it is just because of unpredictable sorts of the arguments, the constructors are fine and can have views. Otherwise, we might have to handle them as `CUST`.

jonsterling:
@favonia I can't remember, but I think the reason was just sheer laziness on my part.

[BTW, here's a comment that is not immediately relevant, but I wanted to say it: I'm of two minds, by the way, on the issue of these views... It seemed important because it let us impose some abstraction on the very raw world of ABTs, but on the other hand, the view "isomorphisms" (I'm not sure if it is actually an isomorphism!) are probably pretty expensive in some places.]

favonia:
@jonsterling What's the Immortal way to extend the signature now? I don't want to touch `Src` yet.

ecavallo:
> @ecavallo By the way, is there a reason that func cannot be a meta-function? I didn't put it into the syntax because I believe it can be done that way.

No, no reason.

favonia:
@jonsterling @ecavallo @cangiuli The implementation here is so simple-minded that it cannot even have pushouts which has parameters. It is not too difficult to support them, but the issue of `CUST` will be more pressing---how should we put all the arguments to the type and the constructors together?

favonia:
Another direction is to implement the full proof theory in the paper. The difference is whether an inductive types with the same constructors are definitionally the same type. In the current PR, no, but in the full theory, yes.

ecavallo:
By the way, I think it is a good idea to have two separate forms of constructor: constructor with boundary and constructor without boundary. (In the paper, there is instead a distinction between 0-constructors and >0-constructors, but that's not really the issue.)

It seems fine to me to not allow proving equalities of inductive types by comparing constructors.

favonia:
@ecavallo Okay. Could you remind me again the special treatment of such a constructor?

By the way, if no constructor mentions any dimension, then the type is also discrete. Is there anyway to integrate this with the special treatment above, if any?

The practical difficulty for me right now is to allow parameters (such as `(Pushout a b c f g)`) without comparing constructors. Any thoughts?

ecavallo:
For a constructor without boundary, coe on an intro term just steps to intro of coes, no boundary-fixing fcom needed. I think this is the only difference. I don't know if there is a good way to unify this with the conditions for discreteness.

To clarify my comment about comparing constructors, I meant rather that we should do whatever is more practical. But could you explain how this relates to parameters?

jonsterling:
@favonia This is not super important yet, but I find the proposed notation to be very confusing... Particularly the fact that it ends in this `(U #l #k [lbl : ty] ...)` thing. I see what you mean by it, but I think it will just mystify everyone.

Can we do something like a more traditional datatype declaration (but which has the same structure, of course)? Perhaps:

```
Data S1 : [ (U 0 kan) ] where [
  [base : S1]
  [loop : (-> dim S1)]
] by [
  ?
]
```

favonia:
Alright, but it still has to be
```
Data S1 : [ (U 0 kan) ] where [
  [base : S1]
  [loop : (-> dim S1) [[x] x=0 base] [[x] x=1 base]]
] by [
  ?
].
```
or
```
Data S1 : [ (U 0 kan) ] where [
  [base : S1]
  [loop : (-> [x : dim] S1 [x=0 base] [x=1 base])]
] by [
  ?
].
```
or something like that.

favonia:
Now, the one with parameters...
```
Data Pushout(#i:lvl) : [
  (->
   [tya tyb tyc : (U #i)]
   [f : (-> tyc tya)]
   [g : (-> tyc tyb)]
   (U #i kan))
] with {
   [left : (-> tya ($ Pushout tya tyb tyc f g))] ???
   [right : (-> tyb ($ Pushout tya tyb tyc f g))] ???
   [glue : (-> [x : dim] [c : tyc] (($ Pushout tya tyb tyc f g) [x=0 ($ f c)] [x=1 ($ g c)]))]) ???
} by [
  ?
].
```
How to resolve `tya` in the type of `left`?

jonsterling:
@favonia If we write something that looks like it has binding structure, then in our notation we have to respect that. In any sensible implementation of the above, `tya` is not in scope in that location. So, maybe we need to think of a better notation... Now i see why you had written it before with all those things after `(U ...`; we can't do exactly that notation, but your idea about the binding structure was correct.

If you don't think of anything, I'll try to propose something within the next day or so.

favonia:
@jonsterling Remember there's another urgent issue about extending the metalanguage. I need to do the following three things:

1. Start checking an inductive type, say S1. It will be marked as an incomplete inductive type.
2. Start checking and adding each constructor, for example base and loop. While checking loop, the information about base should be available.
3. Mark the type as complete. The semantics is that there's no new constructor and the recursor is available.

The current metalanguage simply does not support these, but instead of me spending three days internalizing the language, maybe you can speed up my learning process.

jonsterling:
@favonia OK, let me come up with a lightweight proposal for this... On the one hand I believe it is possible to do this without greatly extending the ML, but by doing it carefully, I may be able to save you a ton of work.

jonsterling:
@favonia Just FYI, I'm a little strapped for time though, because of preparing for the MURI meeting, etc. But I'll make time.

favonia:
@ecavallo Is it okay if I extend the proof theory by always doing beta reduction before formal type checking?

ecavallo:
> @ecavallo Is it okay if I extend the proof theory by always doing beta reduction before formal type checking? This is to avoid the situation where type synthesizing is difficult.

I'd have to think about it, I'm not sure.

favonia:
@ecavallo FYI: A simple case is that `($ (lam [_] base) garbage)` will be a valid recursive term.

favonia:
@ecavallo Actually, is it possible to redo the grammar so that for any recursive term `m` there is no need to synthesize any type? We want to ban `($ (lam [_] base) garbage)`, and I suppose there is a grammatical way of banning this?

favonia:
@ecavallo How about this: a recursive term must be of one of the following forms
```
\x y ... z. intro(...) a0 a1 a2
\x y ... z. fcom(...)
```
where lambdas must be in the prefix?

ecavallo:
That seems fine.

favonia:
@ecavallo It seems we still need to consider `(\x y ... z. p a0 a1 a2 ... an) [(\x y ... z. q b0 b1 b2 ... bn) / p]`. It would be great if we can always do beta-reduction.

favonia:
@ecavallo Is it possible to prove that, a formal term can be replace by another if their interpretations are (stably) (computationaly?) equal?

ecavallo:
> @ecavallo Is it possible to prove that, a formal term can be replace by another if their interpretations are (stably) (computationaly?) equal?

I doubt that this would be true with the current argument proof theory, which doesn't have function extensionality for example. Would it help to have annotations on binders in the argument language?

favonia:
@ecavallo Sorry for my typo... I meant whether it is okay to _make_ that property true. Yes it will help a lot to bave annotations on binders.

ecavallo:
I suppose it wouldn't be a problem to add it, since it will be true in the interpretation.

favonia:
@ecavallo Then my radical proposal is to base the semantics of two spec types/terms being equal on their (all possible) interpretations being equal. Then the rules in papers are just some true lemmas. If you are confident that the paper could have been written with this new semantics, I would go ahead and rethink how to type check these spec terms.

By the way, I am now considering calling them "specification types" and "specification terms". For some reasons I still don't feel completely comfortable with "argument" when what we are dealing with are for both arguments to constructors and boundaries. Recursive types/terms are unfortunately taken.

ecavallo:
I like specification types/terms.

What type do you want their interpretations to be equal at? The inductive type is not defined until we already know the specification is well-formed. I suppose one could quantify over all PERs which are algebras for the prefix...

favonia:
@ecavallo That's a good question---and relatedly I want your inputs on universe levels if you already thought about them.

ecavallo:
I do think quantifying over algebras would work (the algebras are basically type candidates), but it would take some work to rewrite the proofs.

For universe levels, an inductive should be in the least upper bound of the levels of the types which occur in its spec.

favonia:
@ecavallo I think about this and feel requiring annotations is too demanding from the user's point of view. I will go with the new semantics.

favonia:
I wonder why in https://travis-ci.org/RedPRL/sml-redprl/builds/351291264 MLton and SML/NJ gave different results. Do we have wildly different cm and mlb files?

src/redprl/refiner_composition_kit.fun
```diff
@@ -149,7 +149,8 @@ struct
       end
 
     local
-      (* TODO: why do these have tick marks after them? - JMS *)
+      (* These functions have ticks because their correctness depends on the code in
+       * genInterTubeGoals. -favonia *)
```
favonia: FYI @jonsterling

src/redprl/refiner_types.fun
```diff
@@ -1919,6 +1919,91 @@ struct
...
+    fun appReduce tm : abt =
+      case Syn.out tm of
+         Syn.IND_SPEC_INTRO params => tm
+       | Syn.IND_SPEC_FCOM params => tm
+       | Syn.IND_SPEC_LAM (x, ax) =>
+           Syn.into (Syn.IND_SPEC_LAM (x, appReduce ax))
+       | Syn.IND_SPEC_APP (a, b) =>
+           case Syn.out (appReduce a) of
+              Syn.IND_SPEC_LAM (x, ax) => appReduce (Abt.substVar (b, x) ax)
+            | a => Syn.into (Syn.IND_SPEC_APP (Syn.into a, b))
```
favonia: @ecavallo So here is the function to do beta reduction blindly.

src/redprl/refiner_composition_kit.fun
jonsterling: thank you!

jonsterling:
@favonia

1. It's possible that MLton is right, but that CM is overly permissive... I have noticed that sometimes it accepts things that it shouldn't for reasons I don't quite understand.

2. Or it may be that there is a different order of imports in a .mlb file. MLB is order-sensitive, whereas CM does a topological sort of dependencies.

favonia:
@jonsterling @ecavallo One thing I just realized is that, with the new semantics, all the new atomic judgments can be discharged automatically by a type-checker of the simply-typed lambda calculus---we technically do not need any new judgment forms! Just like we do not need `Synth` in the end. The only issue is, as a design issue, do we care about showing an overall goals first? We can just invoke the little type-checker and only show the subgoals.

favonia:
Please comment on the concrete syntax
```
Data S1 : (U 0)
with
  [ base
  , loop [x : dim] [x=0 base] [x=1 base]
  ]
by [ auto ].

Data Pushout (#l: lvl)
  [a b c : (U #l kan)]
  [f : (-> c a)] [g : (-> c b)] : (U (#l ++) kan)
with
  [ left a
  , right b
  , glue (x : c) [y : dim] [y=0 (left ($ f x))] [y=1 (right ($ g x))]
  ]
by [ auto ].
```

jonsterling:
this looks pretty cool!

favonia:
To simplify the parser:
```
Data PropTrunc (#l:lvl) [a : (U #l kan)]
  { incl a
  , sq [x y : self] [z : dim] [z=0 x] [z=1 y]
  }
by [ auto ].
```

ecavallo:
I realized I have a lingering question about the new computational version of schema-typing. Does it create any problem with the dependent interpretation of spec terms needed for the elimination typing judgment?

favonia:
It shouldn't create any new problems because any specific type is a value. Perhaps I misunderstood your concern?

ecavallo:
What I mean is: for the original definition of spec term equality, it is a theorem that (1) all interpretations of equal spec terms are equal, but also (2) all dependent interpretations of the spec terms are equal. If we change the definition of the spec term judgment to equate anything that satisfies (1), what do we do for (2)?

favonia:
@jonsterling I don't know the new syntax. What should we do about the following?
```
data S1 : (U 0 kan)
  { base
  , loop [x : dim] [x=0 base] [x=1 base]
  }
by { auto }.

data Pushout (#l:lvl)
  [a b c : (U #l kan)]
  [f : (-> c a)] [g : (-> c b)] : (U #l kan)
  { left a
  , right b
  , glue [x : c] [y : dim] [y=0 (left ($ f x))] [y=1 (right ($ g x))]
  }
by { auto }.

Data PropTrunc (#l:lvl) [a : (U #l kan)] : (U #l kan)
  { incl a
  , sq [x y : self] [z : dim] [z=0 x] [z=1 y]
  }
by { auto }.
```

jonsterling:
@favonia IMO this notation looks great and seems to cohere with the style I've been using in my other PR.

src/redprl/refiner_inductive_spec.fun
```diff
@@ -0,0 +1,360 @@
...
+    fun EqSpecType H ((ty0, ty1), level) =
+      case (Syn.out ty0, Syn.out ty1) of
+         (Syn.IND_SPECTYPE_SELF, Syn.IND_SPECTYPE_SELF) => []
+       | (Syn.IND_SPECTYPE_FUN (a0, x, b0x), Syn.IND_SPECTYPE_FUN (a1, y, b1y)) =>
+           let
+             val z = Sym.new ()
+             val b0z = VarKit.rename (z, x) b0x
+             val b1z = VarKit.rename (z, y) b1y
+             (* favonia: more research needed for other kinds *)
+             val goalA = makeEq trace H ((a0, a1), Syn.intoU (level, K.COE))
```
favonia: @ecavallo This is to say that, for `A -> X` to be a spec type, `A` needs to be a coe type.

src/redprl/refiner_inductive_spec.fun
```diff
@@ -0,0 +1,360 @@
...
+             val true = List.all (fn (r0, r1) => inSet r0 andalso inSet r1) eqs
+             val _ = Assert.tautologicalEquations "checkConstr' tautology checking" eqs
+           in
+             EqSpecInterBoundary H (constrs, specctx) boundaries
+           end
+       | Syn.IND_CONSTR_FUN (a,x,bx) =>
+           let
+             val w = Sym.new () (* is it possible to save this? *)
+             val goal = makeMem trace H (a, Syn.intoU (level, K.KAN))
+             val rest = checkConstr' (H @> (w, AJ.TRUE a), dimset) (constrs, specctx) (VarKit.rename (w, x) bx) level
```
favonia: @ecavallo This is to say that, in the Gamma of the constructor data, everything still needs to be Kan. I believe it is still the case even with the 5-level type theory?

favonia:
@ecavallo @jonsterling Hey, can we agree that, the arguments to meta variables must be alpha equivalent for types/intro/elim to be the same? I do not want to check whether two instances are the same.

jonsterling:
@favonia I don't have much context here so you will need to rely on @ecavallo's opinion, but let me say generally that in CTT, the strongest notion of definitional equivalence (as opposed to judgmental equivalence) possible is alpha equivalence, so it makes sense :) (LOL!)

favonia:
@jonsterling For example I can define a data type
```
data Blah (#e:exp) ...
```
How do I know whether `(Blah (lam [_] zero))` is the same as `(Blah base)`? Well, they simply aren't!

favonia:
Now `Assert` and `Restriction` become new structures under `syntax`.

favonia:
@jonsterling The resolver should know the arity during the elaboration of the data type declaration... So I should change the type of `env` in `resolver.fun`, right?

jonsterling:
@favonia Yes, I think you may have to do so.

src/redprl/metalanguage/type.sml
```diff
@@ -2,4 +2,26 @@ structure MlType :> ML_TYPE =
 struct
   datatype vty = datatype MlTypeData.vty
   datatype cty = datatype MlTypeData.cty
+
+  val rec eqVty =
+    fn (ONE, ONE) => true
+     | (ONE, _) => false
```
favonia: Because of the new precomputed arity info, `vty` is not an `eqtype` anymore, but I think you would like the new pattern matching style which still exploits coverage checker but looks nicer.

test/success/S1-as-inductive.prl
```diff
@@ -0,0 +1,7 @@
+data S1' : (U 0 kan)
+  { base'
+  , loop' [x : dim] [x=0 self.base'] [x=1 self.base']
+  }
```
favonia: Behold, the future `S1`! (There is still some strange bug.)

favonia:
The new dotted syntax. Note that `(.$ f x)` and `(.lam [x] blah)` are for application and lambda in the specification language.

```
data S1 : (U 0 kan)
  { base
  , loop [x : dim] [x=0 .base] [x=1 .base]
  }
by { auto }.

data Pushout (#l:lvl)
  [a b c : (U #l kan)]
  [f : (-> c a)] [g : (-> c b)] : (U #l kan)
  { left a
  , right b
  , glue [x : c] [y : dim] [y=0 (.left ($ f x))] [y=1 (.right ($ g x))]
  }
by { auto }.

data PropTrunc (#l:lvl) [a : (U #l kan)] : (U #l kan)
  { incl a
  , sq [x y : self] [z : dim] [z=0 x] [z=1 y]
  }
by { auto }.
```

src/redprl/syntax/variable_kit.sml
```diff
@@ -23,4 +23,6 @@ struct
 
   val substMany = RedPrlAbt.substVarenv o ctxFromList
   fun subst s = substMany [s]
+
+  fun alphaRenameTubes w = List.map (fn (eq, (u, tube)) => (eq, rename (w, u) tube))
```
favonia: @jonsterling Not sure if this function belongs to the variable kit...

jonsterling: It's ok, we can find a better place for it later.

favonia:
I suggest merging this now because it has some improvements over the infrastructure. I can continue working on it later.

PR #644, opened 2018-03-12
Use HTTPS as much as possible.

favonia:


jonsterling:
A more serious problem that I noticed is apparently the fonts are not loading even as-is (master). I had to reset my browser cache to be able to observe the problem... Apparently it is getting blocked by an issue with CORS.

But here is the bizarre thing: they fail to load from `~redprl` but when I change it to `~jmsterli`, it works fine. The two have the exact same `.htaccess` file. Could there be something differently configured on AFS or something that is causing this? In the meanwhile we need to switch to `~jmsterli` until we figure out why the `~redprl` one doesn't work.

jonsterling:
And things that affect the display of the documentation (like loading fonts, etc.) need to be done on the `docs` branch, not a separate branch; this is because the `docs` branch gets built and served by readthedocs, whereas we can't tell in advance whether other branches break the docs.

favonia:
Closed in favor of another PR for the `docs` branch.

PR #649, opened 2018-03-19
Revamp metalanguage, get rid of "source signature" crap

jonsterling:
This is a necessary precondition to completing #636.


src/redprl/metalanguage/elaborate.fun
```diff
@@ -148,59 +127,184 @@ struct
...
+  fun elabDataDecl (name, psi, foo) : elab_cmd =
+    let
+      (* TODO *)
+      val result : elab_val = fn env => (ISyn.DATA_INFO {foo = ()}, Ty.DATA_INFO)
+      val resultAbs =
+        (* To insert a check of a sequent against a script, use:
+            elabBind (fn env => refineSequent (name, sequent, elabAst env (script, RedPrlSort.TAC)), MlId.fresh "_", .....)
+
+            To check multiple sequents, just do nest this multiple times.
+            - jms
```
jonsterling: 🎯

src/redprl/metalanguage/semantics.sig
```diff
@@ -3,13 +3,14 @@ signature ML_SEMANTICS =
...
-  type jdg = AtomicJudgment.jdg
+  type jdg = Sequent.jdg
   type term = Tm.abt
   type metas = (Tm.metavariable * Tm.valence) list
 
   datatype value =
      THUNK of env * syn_cmd
-   | THM of jdg * term
+   | THM of jdg * Tm.abs
+   | DATA_INFO of {foo : unit} (* For Favonia: This should be the info associated with the *instances* of the data declaration; so it would not include its parameters. Those would be kept in an ABS. *)
```
jonsterling: 🎯

src/redprl/metalanguage/syntax.sml
```diff
@@ -0,0 +1,80 @@
...
+  type vty = MlType.vty
+
+  type metas = (metavariable * valence) list
+  type arguments = (metavariable * valence) list
+
+  datatype cmd =
+     DEF of {arguments : arguments, definiens : term}
+   | THM of {name : string, arguments : arguments, goal : jdg, script : RedPrlAst.ast}
+   | TAC of {arguments : arguments, script : RedPrlAst.ast}
+   | DATA_DECL of {name : string, arguments : arguments, foo : unit}
```
jonsterling: 🎯

src/redprl/metalanguage/syntax.sml
```diff
@@ -0,0 +1,80 @@
...
+  type metas = (metavariable * valence) list
+
+  datatype value =
+     THUNK of cmd
+   | VAR of id
+   | NIL
+   | ABS of value * value
+   | METAS of metas
+   | TERM of term
+   | DATA_INFO of {foo : unit}
```
jonsterling: 🎯

src/redprl/mini_signature.sig
```diff
@@ -13,4 +13,7 @@ sig
   val unfoldOpid : sign -> opid -> abt RedPrlAbt.bview list -> abt
 
   val isTheorem : sign -> opid -> bool
+
+  (* For Favonia - jms *)
+  val dataDeclInfo : sign -> opid -> unit
```
jonsterling: 🎯

favonia:
Thanks :heart: I think it will be an `abt` as the first cut (because I want to recycle the code for binding). We'll see.

favonia:
BTW What happened to those two functors?

jonsterling:
@favonia They used to be functorized over an implementation of the ML syntax, but now there is only one implementation of the syntax, so I got rid of the functor.

favonia:
@jonsterling Is it possible to check multiple sequents against only one script?

jonsterling:
@favonia Do you mean that you want to run the same tactic to solve multiple goals? Or did you want to have a multitactic that is run against the goals as a whole?

The former is easy, the latter is a little harder, but if it's absolutely needed, I can probably support it somehow.

favonia:
The latter. In the example there's only one `auto`. A work around is to have a dummy judgement and then some fancy rule to expand it, but I'm reluctant to write the pretty printer now...

jonsterling:
I see... I guess I need to change the ML's `REFINE` command to support a proof state as input... Let me figure out how to do this now.

PR #653, opened 2018-03-20
Fix font on docs [temporary fix]

jonsterling:
For some reason, the `redprl` account on `cs.cmu.edu` is blocking CORS requests, even though I have set `.htaccess` properly; however, it works fine with my personal account. Until we get that figured out, I'm reverting the site to point to `jmsterli`. Hopefully we can get the CORS issue sorted out after we have a bit more time.


favonia:
I don't have time now, but someone should dump the HTTP requests and responses to see whether the headers are there.

favonia:
I got these errors in the console
```
Loading failed for the <script> with source “https://www.google-analytics.com/analytics.js”.
atomic-judgment.html:1

Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://media.readthedocs.org/fonts/fontawesome-webfont.woff2?v=4.6.3. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing).

Use of getPreventDefault() is deprecated.  Use defaultPrevented instead.
jquery-2.0.3.min.js:3:21579

Source map error: request failed with status 404
Resource URL: https://media.readthedocs.org/css/sphinx_rtd_theme.css
Source Map URL: theme.css.map
```

jonsterling:
@favonia That is just a problem with a font that is loaded by the "readthedocs" template, and I don't think it's for us to fix on our end. I think it's just for some wingdings or something...

favonia:
Well, in any case, the web font does not seem to work for me...?

jonsterling:
@favonia Wow, weird... it works for me?? To clarify, you are not seeing the PragmataPro font when you load http://www.redprl.org/en/docs/ (this is the version built from this branch)?

Note that http://www.redprl.org/ should currently be broken, and also if you try to run sphinx locally, the fonts will not work either.

favonia:
I think the main reason is that `https://www.cs.cmu.edu/~jmsterli/fonts/PragmataPro.css` gives me 403.

favonia:
Oh, I see. I spoof all the referrer and potentially tracking headers. Glad that you are the main one who do the demo.

favonia:
I found the bug---somehow you are using `http` for `~redprl`.

favonia:
Sorry that's not the reason. Somehow the `Origin` is missing.

favonia:
I think maybe the AFS permission is the reason...

PR #666, opened 2018-03-31
Document the rest of the refinement rules

timjb:


timjb:
@jonsterling @favonia This is ready to be reviewed

doc/src/refine.rst
```diff
@@ -636,9 +767,8 @@ V types
       (hcom, hcom) if #k == hcom
       (coe, com) if #k == coe
       (pre, pre) if #k == pre
-    #isEquiv f =
-  | H >> #e0 = #e1 in (Equiv #a0 #b0)
-  | H >> #a0 = #a1 in (U #l #ka)
+  | H, #r=0 >> #e0 = #e1 in (Equiv #a0 #b0)
```
jonsterling: Good catch 😄

doc/src/refine.rst
```diff
@@ -653,6 +783,13 @@ where ``Equiv`` is defined by
 
 :index:`V/eq/uain`
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+::
+
+  H >> (vin #r #m0 #n0) = (vin #r #m1 #n1) in (V #r #a #b #e)
+  | H, #r=0 >> #m0 = #m1 in #a
+  | H >> #n0 = #n1 in #b
+  | H, #r=0 >> ($ (! f #e) #m0) = #n0 in #b
```
jonsterling: This should be `! proj1` instead of `! f`

doc/src/refine.rst
```diff
@@ -615,12 +672,86 @@ Composite types
...
+  | H, k:dim, #r(n-1)=#s(n-1), #rn=#sn >> (#T0(n-1) k) = (#T1n k) coe type
+  | H, #r0=#s0 >> #Cap = (#T0 #i) type
+  | ...
+  | H, #rn=#sn >> #Cap = (#Tn #i) type
 
 :index:`fcom/intro`
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+::
+
+  H >> (fcom #i~>#j #Cap [#r0=#s0 [k] (#T0 k)] ... [#rn=#sn [k] (#Tn k)])
```
jonsterling: I didn't even realize we had implemented this rule! It's a monster 😆

doc/src/refine.rst
timjb: fixed

timjb:
@jonsterling One other idea: I suggest replacing `#t(n-1)` (which is not a valid name in RedPRL) by `#t/n-1`. I've added one commit doing this for 'hcom/eq/tube'. If you agree, I will do this for all other composition rules.

jonsterling:
@timjb I like your idea for names!

timjb:
@jonsterling I've applied the new naming scheme in all rules. Let's merge now?

PR #667, opened 2018-04-01
Inductive types, stage 2: rules.

favonia:


favonia:
(Oh my, the current arity handling is so mortal...)

favonia:
There is a shift/reduce conflict...

favonia:
I am not happy with this syntax, but it is very easy to parse.

jonsterling:
> I am not happy with this syntax, but it is very easy to parse.

I agree, but I'm fine with it for now 😄It's exciting that this stuff works now!

jonsterling:
I have an exam I really need to study for, so I need to recuse myself from examining pull requests for the next few days. Can @cangiuli or @ecavallo review this?

cangiuli:
What does `&` do in the syntax?

favonia:
I forgot to remove it after experimenting alternative syntax. Now it's removed.

jonsterling:
I wonder, should we (not in this PR) just remove the built-in version of the circle since we can define it now?

favonia:
@jonsterling Yes, eventually. Not just circle, but also natural numbers, integers, booleans, pushouts and coequalizers. Maybe empty types as well.

favonia:
Note that the current `coe` of the defined `S1'` is different from the `coe` of the built-in one.

ecavallo:
Thanks Favonia!
How does `elim` work for inductive types? I tried

```
theorem Test :
  (-> (. S1' type) S1)
by {
  lam a => elim a;
  ?
}.
```

in `test/success/inductive.prl`, but I get `Sort Error: expected dim, but got exp`.

favonia:
I no longer know how it works now, except that it is wrong. Expecting another PR.

favonia:
Fixed in #674.

PR #671, opened 2018-04-18
Make (U pre) and (U hcom) members of (U kan).

favonia:


src/redprl/refiner_types.fun
```diff
@@ -2411,6 +2411,45 @@ struct
     (* TODO Add the Elim, EqCap and Eta rules. *)
   end
 
+  (* FormalComposition's evil twin *)
```
jonsterling: LOL E stands for Evil!

timjb:
The refinement rule `universe/eqtype` in the documentation also needs to be updated. (You can also fix a mistake I made while documenting this rule: It currently states that we must have `#l <= #l'` instead of the correct `#l < #l'`.)

PR #673, opened 2018-04-18
Optimize coe of fcom further.

favonia:
Thanks to Anders' suggestion. @mortberg


mortberg:
Is this implementing the optimization I suggested yesterday where one decompose the system first and only use Q when necessary?

favonia:
@mortberg Yes.

mortberg:
Cool! I'll take a look later. Do you have any examples using this code? I'm currently trying to compute winding numbers using yacctt, but I have bug somewhere... The computation of the winding number of `loop ; loop` you should call this code.

This optimization reminds me of how we do comp for Glue in CCHM. We first define `delta = forall i. phi` and then use that throughout the algorithm.

favonia:
Tested with `($ winding ($ loop^ 5)) = 5`, the improvement on average is about 1-2%.

```
$ sudo perf stat -r 10 bin/redprl example/omega1s1.prl

 Performance counter stats for 'bin/redprl example/omega1s1.prl' (10 runs):

       2755.159027      task-clock (msec)         #    1.000 CPUs utilized            ( +-  0.31% )
                15      context-switches          #    0.006 K/sec                    ( +- 35.79% )
                 0      cpu-migrations            #    0.000 K/sec                  
           871,839      page-faults               #    0.316 M/sec                    ( +-  0.00% )
     9,253,212,032      cycles                    #    3.359 GHz                      ( +-  0.22% )
    12,343,462,093      instructions              #    1.33  insn per cycle           ( +-  0.01% )
     2,769,464,710      branches                  # 1005.192 M/sec                    ( +-  0.01% )
        27,526,588      branch-misses             #    0.99% of all branches          ( +-  0.29% )

       2.755610110 seconds time elapsed                                          ( +-  0.31% )

$ sudo perf stat -r 10 bin/redprl-master example/omega1s1.prl

 Performance counter stats for 'bin/redprl-master example/omega1s1.prl' (10 runs):

       2749.866388      task-clock (msec)         #    1.000 CPUs utilized            ( +-  0.24% )
                14      context-switches          #    0.005 K/sec                    ( +- 28.99% )
                 0      cpu-migrations            #    0.000 K/sec                    ( +-100.00% )
           871,840      page-faults               #    0.317 M/sec                    ( +-  0.00% )
     9,258,102,361      cycles                    #    3.367 GHz                      ( +-  0.07% )
    12,344,206,354      instructions              #    1.33  insn per cycle           ( +-  0.01% )
     2,769,506,106      branches                  # 1007.142 M/sec                    ( +-  0.01% )
        27,371,344      branch-misses             #    0.99% of all branches          ( +-  0.17% )

       2.750357499 seconds time elapsed                                          ( +-  0.24% )

$ sudo perf stat -r 10 bin/redprl-coe example/omega1s1.prl

 Performance counter stats for 'bin/redprl-coe example/omega1s1.prl' (10 runs):

       2716.201768      task-clock (msec)         #    1.000 CPUs utilized            ( +-  0.17% )
                22      context-switches          #    0.008 K/sec                    ( +- 29.36% )
                 0      cpu-migrations            #    0.000 K/sec                  
           871,211      page-faults               #    0.321 M/sec                    ( +-  0.00% )
     9,115,342,370      cycles                    #    3.356 GHz                      ( +-  0.12% )
    12,289,316,191      instructions              #    1.35  insn per cycle           ( +-  0.01% )
     2,758,518,595      branches                  # 1015.579 M/sec                    ( +-  0.01% )
        25,184,106      branch-misses             #    0.91% of all branches          ( +-  0.25% )

       2.716728973 seconds time elapsed                                          ( +-  0.17% )

$ sudo perf stat -r 10 bin/redprl-coe example/omega1s1.prl

 Performance counter stats for 'bin/redprl-coe example/omega1s1.prl' (10 runs):

       2724.598222      task-clock (msec)         #    1.000 CPUs utilized            ( +-  0.33% )
                25      context-switches          #    0.009 K/sec                    ( +- 32.03% )
                 0      cpu-migrations            #    0.000 K/sec                  
           871,212      page-faults               #    0.320 M/sec                    ( +-  0.00% )
     9,091,677,441      cycles                    #    3.337 GHz                      ( +-  0.13% )
    12,289,246,855      instructions              #    1.35  insn per cycle           ( +-  0.01% )
     2,758,510,525      branches                  # 1012.447 M/sec                    ( +-  0.01% )
        25,180,454      branch-misses             #    0.91% of all branches          ( +-  0.17% )

       2.725129573 seconds time elapsed                                          ( +-  0.33% )
```

mortberg:
Cool! I guess the systems are not that big in that example so the optimization doesn't really make much of a difference, but it's nice that it works.

favonia:
@mortberg By the way, the `forall` thing is achieved by `keepApartTubes`. It is used everywhere as well, but I feel maybe it does not have too much to do with this optimization. I think this optimization is more about simplifying
```
com ... [eq1 -> com ... [eq1 -> n]]
```
to
```
com ... [eq1 -> n]
```

mortberg:
Do I read your benchmarks correctly if I understand that it takes RedPRL 2.75s to comute winding of 5 loops?

The corresponding comutation in cubicaltt (master branch):

```
> loopZ5
EVAL: inr (suc (suc (suc (suc (suc zero)))))
Time: 0m0.089s
```

There is currently a bug in yacctt, but hopefully I can compute this there soon as well.

favonia:
To be precise, `5` and `-5`. Time to throw away the current ABT library, I suppose. @jonsterling

favonia:
@mortberg Wait, is that some build-in function to measure the time? The benchmark above is done against the entire file.

mortberg:
Yes, it only measures the time of that specific computation. Typechecking the whole file together with 3 proofs of univalence and the "grad lemma" takes about 0.3s. It would probably be instant if I would strip out all the stuff that it not used in the definition. The most accurate comparison should be to just measure the time it takes for RedPRL to do that specific computation and nothing else.

favonia:
@mortberg Hmm we do not support that yet. @jonsterling Maybe we should consider dumping the general ABT library sooner.

By the way, someone needs to approve this PR before I can merge it.

jonsterling:
@favonia Can you clarify what you mean in terms of the ABT library? I'm not sure I see the connection. (BTW, I'm fine with dropping that at any time if you feel it's worth it!)

favonia:
@jonsterling According to @mortberg yacctt is ~10x faster and I can only guess the main culprit is the general ABT library.

Perhaps what we need is an ABT library generator or something.

favonia:
@mortberg Other than specialized data types rather than the general ABT library, did you use other magic in cubicaltt/yacctt? Perhaps names are another place we should work harder on in RedPRL.

jonsterling:
@favonia Ah, I see! Yes, I can believe that. But please note that there are other things that are slowing us down besides the ABTs themselves:

1. Proof refinement is very slow
2. Our machine proceeds using substitution everywhere. The cubicaltt and yacctt evaluators use environments.

I think we should consider switching away from the stupidly slow ABT representation, but it is probably the case that the only way to make things much faster is to unleash what I am currently developing, which is I guess in some sense a rationalized version of yacctt.

mortberg:
Note that the computation I measured was on **cubicaltt** (master branch), and not yacctt. There is still some bug in yacctt making the winding numbers not work.

We are treating names extremely naively in cubicaltt. We compute the support every time we need a fresh dimension variable (which is all the time). This is not so bad for something simple like the winding numbers where the terms are small, but for the size of terms involved in pi4(s3) this does not work. Instead I have a separate branch where I use a global counter and `unsafePerformIO` to generate fresh names, this forces me to switch off some optimizations in GHC making simple things like the winding numbers slower (about 0.3s for the thing we measured above), but for things needed in pi4(s3) it is incredibly much faster (micro seconds compared to minutes). 

 In yacctt we use a global counter implemented using a monad and no `unsafePerformIO`. This should be much faster as I can keep all GHC optimizations on. I'll send you the timings for the winding numbers once I get them to work.

favonia:
@mortberg Thanks for the information. It is nice to have multiple implementations so that we can learn from each other.

PR #674, opened 2018-04-18
Inductive: bug fixes.

favonia:


favonia:
There are still bugs in the machine...

example/omega1s1-inductive.prl
```diff
@@ -0,0 +1,161 @@
+data S1' : (U 0 kan)
+  { base'
+  , loop' [x : dim] [x=0 (self base')] [x=1 (self base')]
+  }
+by {
+  auto
+}.
```
favonia: @jonsterling We are very close to dump the old `S1`...

PR #677, opened 2018-04-19
Fix the variable ordering bug in Restriction. Close #668.

### LINKED ISSUE #668: Bug in the logic of Restriction.restrict?
I believe `Restriction` might output an invalid dimension substitution `[y/x]` for the sequent
```
x:dim, a:(V x ...), y:dim >> J
```
This was okay when dimension variables must be in the prefix. An ugly solution is to consult the order of variables in the context. A less ugly solution is to introduce `z:dim` in the prefix and then do `[z,z/x,y]`. Any Thoughts or at least thoughts? @jonsterling @cangiuli

favonia:
The solution which consults the ordering in the context.


jonsterling:
I don't really understand this solution, but I believe you 😄 

@cangiuli any thoughts?

PR #679, opened 2018-04-19
Fix #664.

### LINKED ISSUE #664: Bug in HCom.Eq
While documenting our rules, @timjb noticed that the [`HCom.Eq` rule](https://github.com/RedPRL/sml-redprl/blob/d08b0937242685ed80ae3fd73e5484c3aa8dea1c/src/redprl/refiner_composition_kit.fun#L152) does not seem to be checking that the type arguments of the two compositions are equal. This seems wrong to me, but I might be confused...

@favonia Do we need to change the subgoal?

favonia:
Thanks to @timjb.

### DIFF
--- a/src/redprl/refiner_composition_kit.fun
+++ b/src/redprl/refiner_composition_kit.fun
@@ -168,7 +168,7 @@ struct
         val _ = Assert.tautologicalEquations "HCom.Eq tautology checking" eqs0
 
         (* type *)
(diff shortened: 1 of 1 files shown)


jonsterling:
@favonia  Does any documentation need to be updated?

favonia:
No, it was already fixed in the documentation.

PR #681, opened 2018-04-25
Representation independence demo

cangiuli:
Here's the cleaned up version of an example I gave in a talk this week, about using univalence to transport functions and laws between two implementations of a "data structure." (I didn't want to mess around with datatypes yet, but it should be possible to write much more convincing examples.)


favonia:
My only complain is that now it seems we have multiple files with the same materials. Maybe we should create a `demo` subfolder for these demos and date them?

example/invariance.prl
```diff
@@ -0,0 +1,184 @@
...
+  (->
+    [ty : (U 0 kan)]
+    [fun : (-> bool ty)]
+    (path [_] (-> bool ty) ($ SwapFun ty ($ SwapFun ty fun)) fun))
+by {
+  lam ty =>
+  `(coe 1~>0
+      [x] (-> [elt : (@ ($ FunEqPair ty) x)]
+              (path [_] (@ ($ FunEqPair ty) x)
+                    ($ (SwapCoe ty x) ($ (SwapCoe ty x) elt))
```
jonsterling: I don't like this indentation. `($...` should have its indentation not determined by the right end of `path`, but by the left end of that lexeme.

cangiuli:
@favonia The real problem is that many different files have the definition of weak connections, the definition of equivalences, ...

Ideally we would have some method of importing definitions so they wouldn't need to be replicated. I'm not against having a `demo` folder but I'd really just like to not have to copy-paste as much!

PR #682, opened 2018-04-25
Rename refinement rules for V types.

cangiuli:
It turns out nobody had ever tried to invoke a refinement rule for V types before today, because they were impossible to invoke. Let's see if you can spot the issue:

in the lexer:
```
<INITIAL>{lower}{identChr}* => (Tokens.VARNAME (posTupleWith (size yytext) yytext));
```
in the parser:
```
atomicRawTac
  : REFINE VARNAME (Ast.$$ (O.RULE_PRIM VARNAME, []))
```
in the refinement rule lookup:
```
     | "V/eqtype" => Lcf.rule V.EqType
```

### DIFF
--- a/doc/src/refine.rst
+++ b/doc/src/refine.rst
@@ -756,7 +756,7 @@ Composite types
 V types
 -------
 
-:index:`V/eqtype`
+:index:`v/eqtype`
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 ::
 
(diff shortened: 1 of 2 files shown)


jonsterling:
lmao!! I actually knew about this at one point but I forgot to open a ticket. Taste it! 😎

favonia:
I really want capitalized V...

PR #687, opened 2018-06-06
(WIP) Update documentation.

cangiuli:
- An update to the readthedocs Sphinx theme broke some of our CSS.

### DIFF
--- a/doc/src/conf.py
+++ b/doc/src/conf.py
@@ -73,7 +73,7 @@ highlight_language = 'redprl'
 exclude_patterns = ['build', 'Thumbs.db', '.DS_Store']
 
 # The name of the Pygments (syntax highlighting) style to use.
-pygments_style = 'sphinx'
+pygments_style = 'colorful'
 
 # If true, `todo` and `todoList` produce output, else they produce nothing.
 todo_include_todos = True
--- a/doc/src/static/red.css
+++ b/doc/src/static/red.css
@@ -79,13 +79,13 @@ code {
 }
 
 div[class^='highlight'] pre {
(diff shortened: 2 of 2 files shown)


favonia:
The horizontal lines look weird to me. Dashes are grouped in three. LOL

doc/src/conf.py
```diff
@@ -73,7 +73,7 @@
 exclude_patterns = ['build', 'Thumbs.db', '.DS_Store']
 
 # The name of the Pygments (syntax highlighting) style to use.
-pygments_style = 'sphinx'
+pygments_style = 'colorful'
```
jonsterling: Is this change needed? I feel like many of the colors in this theme are kind of strange.

cangiuli: As far as I can tell, the highlighting style we used to use just isn't there anymore, the default that replaced it was horrible, and this was the best one I could find. If anyone can figure out how to get the old style back, please let me know!

cangiuli:
@favonia Yeah, I don't know how to turn off the ligatures! So I made sure I used 9 dashes, since it looked even worse at 8 (two em-dashes followed by an en-dash, lmao).

doc/src/conf.py
jonsterling: ok!

jonsterling:
@cangiuli I think the solution is, we shouldn't turn of the ligatures because ligatures are superior to no ligatures: but we should probably be using the box-drawing symbols instead of dashes for this kind of thing.

jonsterling:
but it's not a big deal

PR #689, opened 2018-06-14
More bugfixes for coe of fcom types.

favonia:
@cangiuli I also updated some notes to avoid the same confusion, stealing a little bit from the Part III style.

@jonsterling Due to the lack of semantic simplification, these optimizations have to be done manually, complicating the RedPRL codebase. Perhaps we found another issue about the RedPRL I approach...


favonia:
@jonsterling @cangiuli It seems challenging to make some regression testing to detect these kinds of typos... any ideas?

PR #691, opened 2018-07-02
add link to lfmtp paper

jonsterling:


cangiuli:
I also replaced a link to the POPL 2017 paper with a link to the CSL 2018 paper. Certainly the second should be linked; if you think we should keep both links, feel free to revert the deletion.

PR #693, opened 2018-10-09
Installation script for vim mode. (see RedPRL/redtt#370)

cangiuli:


jonsterling:
@cangiuli I just force-pushed, lol

PR #697, opened 2019-10-25
Fix keyword.

arthuraa:


jonsterling (approved):
thanks!

jonsterling:
God bless faithful MLton, the only part of the build that still works 🥴

